PR #19, opened 2016-03-14
Exit tests with the return value from RunAllTests()

sbraz:
The test programme should exit with a non-zero status if the tests fail.


sbraz:
@jonoomph Any chance for you to look at this?

jonoomph:
Thanks for the fix, and sorry this took a while =) Much appreciated.

PR #21, opened 2016-03-29
Enable system jsoncpp

### LINKED ISSUE #14: Enable build with system jsoncpp
The Debian packaging policy discourages using embedded dependencies for building binary packages. Right now, libopenshot compiles unconditionally with an embedded copy of `jsoncpp`, which is a breach of this policy.

It would be desirable to allow for (optional ?) compilation and linkage with a system installed `jsoncpp`, such as the one available with the `libjsoncpp-dev` package in Debian / Ubuntu.

komackaj:
This PR resolves #14 by enabling to build libopenshot with system installed libjsoncpp using CMake option `USE_SYSTEM_JSONCPP` (default OFF). If not set, embedded sources of libjsoncpp will be used as before.

 As remarked [here](https://github.com/komackaj/libopenshot/commit/36af4edc23f2adf37e26e54edbb993ff0068dde4#commitcomment-16855054) there is an issue with `jsoncppConfig.cmake` distributed in `libjsoncpp-dev` on Debian, therefore we use custom CMake find module leveraging pkg-config if available and ignore CMake config files for jsoncpp.

It also fixes compilation ambiguity error when using libjsoncpp 0.6.0 and 1.6.5

### DIFF
--- a/cmake/Modules/FindJsonCpp.cmake
+++ b/cmake/Modules/FindJsonCpp.cmake
@@ -1,0 +1,68 @@
+# Find jsoncpp
+#
+# Find the jsoncpp includes and library
+#
+# if you nee to add a custom library search path, do it via via CMAKE_PREFIX_PATH
+#
+# This module defines
+#  JSONCPP_INCLUDE_DIRS, where to find header, etc.
+#  JSONCPP_LIBRARIES, the libraries needed to use jsoncpp.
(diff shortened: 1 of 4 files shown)


jonoomph:
Thanks for the change to FindJsonCpp! Sorry I was a bit delayed getting this merged. Much Appreciated! =)

PR #28, opened 2016-04-30
improve comments, use the trope established by GetSmallestFrame, fix …

RussNelson:
…indent


RussNelson:
I see no easy programmatic way to find the code that is overwriting a data structure. I doubt that it is anywhere in Python itself, so I'm starting to read all of libopenshot to see if I can find the bad code. In the process, I'll find improvements.

PR #29, opened 2016-05-05
Abs looks good

RussNelson:
AudioBufferSource looks clean. Just improving a couple of comments. My only concern is that AudioSourceChannelInfo have a big enough buffer into which to copy the data, but that will wait until I look at that source.

### DIFF
--- a/src/Cache.cpp
+++ b/src/Cache.cpp
@@ -59,7 +59,7 @@ void Cache::Add(long int frame_number, tr1::shared_ptr<Frame> frame)
 	// Create a scoped lock, to protect the cache from multiple threads
 	const GenericScopedLock<CriticalSection> lock(*cacheCriticalSection);
 
-	// Remove frame if it already exists
+	// Freshen frame if it already exists
 	if (frames.count(frame_number))
 		// Move frame to front of queue
 		MoveToFront(frame_number);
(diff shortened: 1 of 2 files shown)


src/Cache.cpp
```diff
@@ -115,19 +115,19 @@ tr1::shared_ptr<Frame> Cache::GetSmallestFrame()
...
-	 deque<long int>::iterator itr;
-	 long int smallest_frame = -1;
-	 for(itr = frame_numbers.begin(); itr != frame_numbers.end(); ++itr)
-	 {
-		 if (*itr < smallest_frame || smallest_frame == -1)
-			 smallest_frame = *itr;
-	 }
+	// Loop through frame numbers
+	deque<long int>::iterator itr = frame_numbers.begin();
+	long int smallest_frame = *itr;
```
komackaj: Cannot frame_numbers be empty when calling this? Then itr should not be dereferenced. Maybe check and return NULL if frame_numbers.empty() ?

RussNelson: If that's a problem here, it's a problem in other places in Cache. Oh, and apologizes for including the Cache changes here. I had the wrong branch checked-out.

komackaj: I can't see any problem anywhere else in Cache.cpp - frame_numbers.begin() is always used in for cycle with termination condition `itr != frame_numbers.end()` that will terminate the loop on first check with empty frame_numbers, so it won't dereference in the cycle's body.

RussNelson: Now I'm confused. I was pretty sure I had copied that model from somewhere else in Cache.cpp, but now I can't find it. :-(

PR #46, opened 2017-01-07
Ffmpegreader fix 1

nlsnho:
Summary
1. Improve stability of FFmpegReader::ReadStream() function that decodes compressed packets
2. Modified behaviour in FFmpegReader::UpdateVideoInfo() function when handling invalid frame rates

Details
1...
a) In the main while loop in ReadStream(), check if the requested_frame is available in the final_cache before decoding any packets. The decoding loop often gets ahead of itself so the requested_frame may already be available. This prevents the working_cache from overflowing and discarding frames that may still be needed for the output.
b) Increased the stuck frame threshold from 40 iterations to 1000. 40 was too low for some of my test material. Normally, frames should not get stuck anyway so although 1000 seems like a big number, it's a recovery event so perhaps better to be too big than too small.
c) Removed end_of_stream condition from 'working frame is final' check since it can cause a premature flush of the working_cache, resulting in blank frames in final_cache.

2...
a) Changed the default video_timebase to 90kHz instead of 24Hz since 90kHz is the most common PTS timebase
b) In the case where the stream's framerate is invalid, disable video for the stream instead of trying to work with an incorrect frame rate which seems to cause the seeking logic to get stuck in loops

@jonoomph

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -90,28 +90,12 @@ int AudioLocation::is_near(AudioLocation location, int samples_per_frame, int am
 		// This is too far away to be considered
 		return false;
 
-	int sample_diff = abs(location.sample_start - sample_start);
-	if (location.frame == frame && sample_diff >= 0 && sample_diff <= amount)
+	// Note that samples_per_frame can vary slightly frame to frame when the
+	// audio sampling rate is not an integer multiple of the video fps.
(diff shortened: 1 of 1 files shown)


jonoomph:
Hiya, sorry for the delayed feedback. Feel free to rebase this off the latest version and I'll take another look. Going to close this for the moment until it's rebased.

PR #54, opened 2017-03-30
Fixes aspect ratio not applied on export (openshot-qt issue #456)

croco3008:
Currently  aspect ratio of videoprofile is passed to FFmpegWriter::SetVideoOptions but not used anyhow in output setup. So, all videoprofiles requering aspect ratios are nonusable. Please review and apply following patch.

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -904,6 +904,10 @@ AVStream* FFmpegWriter::add_video_stream()
 	c->width = info.width;
 	c->height = info.height;
 
+	// Set aspect ratio of stream
+	c->sample_aspect_ratio.num=info.pixel_ratio.num;
+	c->sample_aspect_ratio.den=info.pixel_ratio.den;
+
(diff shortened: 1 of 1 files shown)


jkateva:
This did not by itself fix DVD PAL widescreen problem (always aspect ratio 4:3) for me, had to set setting scaling option to stretch also.

croco3008:
Yes, I know. The second patch will go to the openshot-qt branch. I'm working on it. But this fix is prerequest and is allready usefull. You can export any correcty scaled movie to DVD format now.

jonoomph:
Sorry for the huge delay in getting to this pull request. =) Here is the test that fails: [webm @ 0x1057d60] Aspect ratio mismatch between muxer (0/1) and encoder layer (1/1). If you have time to fix this, I'll be happy to take another look at the pull request.

jonoomph:
Removing this for failed unit tests... please reopen once the issues are resolved. thanks.

Issue #55, opened 2017-04-29
TextReader just displaying black image


jeffski:
I've compiled and installed libopenshot and everything is working as expected including ImageReader and effects (which use ImageMagick of course). I can't get the text reader to work and when I use the example code I just get a black image.

My code is: 

```
#include "OpenShot.h"

using namespace openshot;

int main(int argc, char **argv)
{
	// Create a reader to generate an openshot::Frame containing text
	TextReader r(720, // width
	             480, // height
	             5, // x_offset
	             5, // y_offset
	             GRAVITY_CENTER, // gravity
	             "Check out this Text!", // text
	             "Courier", // font
	             15.0, // size
	             "#fff000", // text_color
	             "none" // background_color
	             );
	r.Open(); // Open the reader
	// Get frame number 1 from the video (in fact, any frame # you request will return the same frame)
	tr1::shared_ptr<Frame> f = r.GetFrame(1);

	// Now that we have an openshot::Frame object, lets have some fun!
	f->Display(); // Display the frame on the screen
	// Close the reader
	r.Close();

	return 0;
}
```

I'm using Ubuntu 14.04 and installed everything following the installation guide. I've built and installed from the latest master.

Any ideas?

jeffski:
in src/TextReader.cpp:

```
//TODO: Reimplement this with QImage
//image_frame->AddImage(copy_image);
```

It's broken? Needs to be updated to use QImage?

jeffski:
Could be `image_frame->AddMagickImage(copy_image);` for now. I can submit a pull request in the next few days.

jeffski:
Thanks for merging @jonoomph

PR #56, opened 2017-04-30
Allowing TextReader to create text frames using ImageMagick

jeffski:
This pull request fixes issue: https://github.com/OpenShot/libopenshot/issues/55 - which is resulting in black frames when trying to create text.


jeffski:
Recreated PR using dedicated branch

PR #77, opened 2018-01-30
Preserve file structure when copying over headers

RichiCoder1:
Changes the install command for headers to install the include directory, rather than the header files, thereby persevering the header structure. Currently using most of the headers in the default include install will fail as the relative paths are wrong.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -334,8 +334,9 @@ INSTALL(  TARGETS openshot
   LIBRARY DESTINATION ${LIB_INSTALL_DIR}
   COMPONENT library )
 	
-INSTALL(FILES ${headers} 
-	DESTINATION ${CMAKE_INSTALL_PREFIX}/include/libopenshot )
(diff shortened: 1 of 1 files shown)


jeffski:
Would this fix the issue where all effects headers are placed in the same folder as other headers? I always find myself having to copy the effects manually when I do a fresh install. This would be useful.

RichiCoder1:
It would indeed, though if I were to redo this patch, it would need more work as just correctly placing the headers isn't enough do to some weird include paths.

jeffski:
This is from my install script that I run for a fresh install:
```
cp -r /opt/libopenshot/include/effects /usr/local/include/libopenshot/effects
cp -r /opt/libopenshot/include /usr/local/include/include
```

I take it you are talking about the `include/include` issue too?

I don't see any harm in solving both problems separately, it is broken anyway. Maybe re-open this and the other issue can be filed as a bug and fixed at a later time?

PR #82, opened 2018-04-06
Prepend current time (hires) to messages logged via ZmqLogger

N3WWN:


N3WWN:
Not sure why this failed - it's a simple change and works perfectly on my dev system which is current git clone 🤷‍♂️

jonoomph:
Thanks, I think this is fine to merge (but we need to watch our builds, and make sure they all support this)

PR #84, opened 2018-04-20
Revert "Prepend current time (hires) to messages logged via ZmqLogger"

jonoomph:
Reverts OpenShot/libopenshot#82, this is not cross-platform, and fails on older Mac builds


N3WWN:
I'm going to submit a new PR which addresses the missing clock_gettime() function on some versions of Mac OS X

PR #88, opened 2018-04-30
[WIP] Use resvg library for more full-featured SVG rendering

N3WWN:
Resolves improper rendering of SVG files per the issue I opened here :  https://github.com/OpenShot/openshot-qt/issues/1081

Also resolves reissue reported here : https://github.com/OpenShot/openshot-qt/issues/838

There will need to a change to `openshot_qt/installer/build-server.py` and `openshot_qt/freeze.py` so that the `libresvg.so` library is packaged with the AppImage builds. I'm not sure what else will need to be changed for other pre-packaged builds as I'm only ever using AppImage builds, even for my own local testing.

Basically, if the image file has an `.svg` extension, the rendering is performed by an external library (`libresvg.so`) and the rendered image is saved to the `homePath()/.openshot_qt/cache` directory. 

resvg supports a "static SVG Full 1.1 subset", which is much more support than what is offered in QtSvg, which "officially supports only a tiny portion of the SVG Tiny 1.2 subset".

resvg is also cross-platform (Linux, Windows and Mac OS X), although I've only used it within my Linux dev env.

`cmake/Modules/FindRESVG.cmake` will try to find the resvg library and, if it is missing, libopenshot will be compiled without resvg support and will continue to function exactly as it does now. This means that devs and power users can clone the libopenshot repo (or otherwise download the source) and compile it themselves in the presence of the `libresvg.so` library to gain this feature, but it does not need to be compiled into the publicly released builds right away.


DylanC:
@N3WWN - This is a really nice PR! :)

N3WWN:
> This is a really nice PR! :)

Thanks @DylanC !

ferdnyc:
Huh. I might've been slightly less chatty on those changes, if I'd realized that _all_ of our comments would be displayed in the PR when it was submitted. :hushed:  (Probably not, but I would've at least considered it!)

ferdnyc:
Hmm, this PR still appears to be sucking in the ZmqLogger hires-time change as a side-effect, @N3WWN .  Think it's worth pushing in another commit to revert that here, so it doesn't cause issues if it's merged before your other PR?

jonoomph:
This looks awesome!!!

jonoomph:
I'm a bit scared to merge this in before our next release (which is any day now), since it will likely break on Mac/Windows... just a guess.

ferdnyc:
It'll certainly require extensive changes to the build environments on all
three platforms, with the incorporation of the rust language compiler and
the cargo package management system. (And you're right, at least on Windows
it'll probably have to look for libresvg.dll instead of the .so.)

Is it possible to merge it onto a new branch, from an existing PR? That
might be a good first path.

N3WWN:
> Hmm, this PR still appears to be sucking in the ZmqLogger hires-time change as a side-effect, @N3WWN . Think it's worth pushing in another commit to revert that here, so it doesn't cause issues if it's merged before your other PR?

Ah, I see it now @ferdnyc !  I just reverted that change.  I must have branched from my ZmgLogger_hires_timestamp branch instead of checking out the master branch first.

N3WWN:
> So, we're not merely adding resvg as a dependency, we're adding (by my count) 41 new dependencies for OpenShot.

@ferdnyc This is my first experience with any rust whatsoever... and I'm going to really show off my ignorance here, but is there a possibility of creating a "rust-static library" that does not require any of the rust packages after the library is built?

ferdnyc:
> if resvg isn't present on the build system, the old (aka current) method is used and OpenShot behaves exactly as it does right now without this PR. This means it should be fine to merge as-is.

That's a good point, I'd forgotten about that. So there really shouldn't be any downside to merging it even in the master tree.

> This is my first experience with any rust whatsoever... and I'm going to really show off my ignorance here, but is there a possibility of creating a "rust-static library" that does not require any of the rust packages after the library is built?

Same here, and I honestly don't know the answer to that. It seems _possible_ to build `staticlib` crates with `rustc` directly (instead of using `cargo`), but I'm not sure if that means the also include all dependencies or if each one has to be separately built the same way. I'm also not sure how that would work with the linked Qt libs and their runtime-loaded plugins. (The build docs mention that resvg makes use of the `plugins/imageformats/libqjpeg.so` JPEG image format plugin that isn't a shared object dependency of the generated `libresvg.so`, so I guess it'd have to be manually linked in if that would even work.)

src/QtImageReader.cpp
```diff
@@ -51,9 +60,70 @@ void QtImageReader::Open()
 	// Open reader if not already open
 	if (!is_open)
 	{
+
+#if USE_RESVG == 1
+		image = std::shared_ptr<QImage>(new QImage());
```
RichiCoder1: You can also do `std::make_shared<QImage>();`. Thats the recommend way.

src/QtImageReader.cpp
```diff
@@ -51,9 +60,70 @@ void QtImageReader::Open()
...
+		bool success;
+
+		QFileInfo image_file_info(QString::fromStdString(path));
+
+		// Only use resvg for files with a '.svg' extension
+		if (image_file_info.suffix() == "svg") {
+			// Init resvg structs
+			resvg_render_tree *rtree = NULL;
```
RichiCoder1: nullptr

src/QtImageReader.cpp
```diff
@@ -51,9 +60,70 @@ void QtImageReader::Open()
...
+			QImage img = QImage(resvgSize.width, resvgSize.height, QImage::Format_ARGB32_Premultiplied);
+			img.fill(Qt::transparent);
+			QPainter p;
+			p.begin(&img);
+			p.setRenderHint(QPainter::Antialiasing);
+			resvg_qt_render_to_canvas(rtree, &opt, resvgSize, &p);
+			p.end();
+
+			// Copy the QImage so it is usable below
+			image = std::shared_ptr<QImage>(new QImage(img));
```
RichiCoder1: std::make_shared<QImage>(img);

RichiCoder1:
No to poo poo on this (excellent) work, but is there a reason you didn't just use QSvgRenderer since you're already taking a dependency on Qt anyways?

src/QtImageReader.cpp
N3WWN: I used this method for consistency since this is the way images are created elsewhere in the project source.

src/QtImageReader.cpp
N3WWN: Yes, initialized to a known value.  Is this causing a warning or error when you try to compile the code?

src/QtImageReader.cpp
N3WWN: Is the suggestion that `(img)` is more efficient than `(new QImage(img))`?

N3WWN:
Yes, QSvgRenderer only supports a small (some would say “Tiny” 😉 )  portion of the SVG spec.

Per http://doc.qt.io/qt-5/svgrendering.html:

Qt supports the static features<http://www.w3.org/Graphics/SVG/feature/1.2/#SVG-static> of SVG 1.2 Tiny<http://www.w3.org/TR/SVGMobile12>.

On the other hand (per https://github.com/RazrFalcon/resvg):

resvg can be used to render SVG files based on a static<http://www.w3.org/TR/SVG11/feature#SVG-static> SVG Full 1.1<https://www.w3.org/TR/SVG/Overview.html> subset, excluding fonts support<https://www.w3.org/TR/SVG11/feature#Font>. In simple terms: no animations, scripting and embedded fonts.

So, while it’s not an entirely complete SVG solution, it is the best that I could find that had multi-platform support.

src/QtImageReader.cpp
RichiCoder1: Ever so slightly, yes. Saves a copy. Outside a tight loop, doesn't mean too much, but still a best practice

src/QtImageReader.cpp
RichiCoder1: Nah, just another best practice thing.

src/QtImageReader.cpp
RichiCoder1: I noticed that right after I made these comments. Tis up to you, see other comment for reasoning

src/QtImageReader.cpp
N3WWN: Gotcha!  I've updated the source, tested it and committed an update.

src/QtImageReader.cpp
N3WWN: Oh, I see!  The comment was a little too laconic and I didn't realize what @RichiCoder1  was saying.  I've updated the source, tested it and committed an update.

src/QtImageReader.cpp
N3WWN: Okay.  I'm going to leave it as is for consistency.  Thanks for pointing this out, though!

src/QtImageReader.cpp
```diff
@@ -84,21 +84,16 @@ void QtImageReader::Open()
...
 
 			// use byte array as data source for resvg
-			char *error = nullptr;
-			rtree = resvg_parse_rtree_from_data(ba.constData(), &opt, &error);
+			int error = 0;
+			error = resvg_parse_tree_from_data(ba.constData(), ba.size(), &opt, &rtree);
 			if (!rtree) {
-				ZmqLogger::Instance()->AppendDebugMethod("QtImageReader::Open", "resvg_parse_rtree_from_data error", -1, error, -1, "", -1, "", -1, "", -1, "", -1);
-				resvg_error_msg_destroy(error);
+				ZmqLogger::Instance()->AppendDebugMethod("QtImageReader::Open", "resvg_parse_tree_from_data error", error, "", -1, "", -1, "", -1, "", -1, "", -1);
```
ferdnyc: You should probably change this to check `if (error)` instead of `if (!rtree)`, since it looks like the semantics of the `resvg_parse_[r]tree_from_data()` function(s) are different. It's conceivable that `rtree` might be non-null even when there is an error.

Also, if `error` is an integer, logging it as-is won't be very meaningful. Is there a utility function to translate the return values from `resvg_parse_tree_from_data` into meaningful error messages? (Or are they just ERRNOs or something standard like that? Then you can use `strerror(error)` to convert to a useful error message string.)

ferdnyc: Ah, I see how the semantics of `AppendDebugMethod()` work now, and why `error` would be logged "raw". (Though, if it is possible to turn the error value into a string with `strerror()` or similar, I would think that would still be useful as an additional component of the log message.)

peanutbutterandcrackers:
I was wondering if switching to resvg means we can now [import svg files as image sequences](https://github.com/OpenShot/openshot-qt/issues/675) now?

I am really in favour of getting this PR merged and release within the `2.4.2.x` bug-fix series as this is, in essence, a bug-fix (IMHO).

ferdnyc:
> I am really in favour of getting this PR merged and release within the 2.4.2.x bug-fix series as this is, in essence, a bug-fix (IMHO).

I object. ("I _strenuously_ object!") 

ReSVG is an entirely new library dependency, hell an entirely new _language_ dependency (Rust, which has never been a dependency of OpenShot component before), and a completely new code path for SVG support. No matter how much we want it (and I'm all in favor of it being available starting with any upcoming release), it's grossly incorrect to classify a complete library replacement as a "bug fix". It fixes no bugs in the existing SVG support. Terminology has to have agreed-upon meanings or it's pointless.

Plus, ReSVG is inevitably going to have to remain an _optional_ dependency for the foreseeable future, since Rust support is too immature to make it a hard requirement. (It would prevent OpenShot being packaged for many Linux  distros.) That alone means that, while it is a very good thing to have, it can't be seen as a bug fix because it won't fix anything for users without the ReSVG component. (The bugs they _are_ seeing are in QtSVG, and ultimately will end up being `wontfix` from our end unless Qt greatly improves QtSVG. However, for those who have the option of using OpenShot with ReSVG, those bugs will no longer be an issue.)

Anyway, _by itself_ ReSVG won't make SVG files importable as image sequences, for the simple reason that the image sequence import code is limited to the following extensions:
```python3
extensions = ["png", "jpg", "jpeg", "gif", "tif"]
```
This is despite the fact that OpenShot considers the following extensions to signify image data:
```python3
if path.endswith((".jpg", ".jpeg", ".png", ".bmp", ".svg", ".thm", ".gif", ".bmp", ".pgm", ".tif", ".tiff")):
```

...Looking at the code, there's no obvious reason why SVG files couldn't be importable as image sequences even with the current QtSVG code, except that they're not. There's no indication whether they were simply omitted from the list, whether they were intentionally excluded because of QtSVG's rendering limitations (in which case, ReSVG might overcome that), or if there was some other reason for the exclusion that has nothing to do with QtSVG (which would presumably still apply with ReSVG).

ferdnyc:
...Also, and entirely tangentially: Looking at those two lines of code above, there's **definitely** no reason why ".tif" files can be image sequences, but ".tiff" files _can't_. I mean, not any _valid_ reason. And yet, that's exactly how things are right now.

peanutbutterandcrackers:
@ferdnyc - Hmm. True. I had not taken that into account. (Sorry. I am too much an OpenShot fanboy at times. lol) All righty then. I shall proceed to remove this from my 2.4.2.x list, then.

2.4.3, perhaps, then?

Looking up PyQT5's repo so that I can submit the svg rendering issue report upstream. But can't seem to find it.... Hmm.....

N3WWN:
Closing this PR in favor of PR #143 to get the PR against `libopenshot:develop` instead of `libopenshot:master`.

ferdnyc:
> Looking up PyQT5's repo so that I can submit the svg rendering issue report upstream. But can't seem to find it.... Hmm.....

Not PyQt. Has _nothing_ to do with them. All they're doing is calling into Qt. It's Qt's issue.

However, it's also one Qt knows about, and doesn't currently intend to fix. QtSVG is documented as being intended only for simple things like display of icons and UI controls, not full-featured generalized SVG processing. As @N3WWN quoted in the OP,  QtSvg "officially supports only a tiny portion of the SVG Tiny 1.2 subset".

Effectively, we're using it wrong by trying to render the titles with it, it's not designed for that. That's a job more suited to something like librsvg. (Or resvg!)

peanutbutterandcrackers:
@ferdnyc - Hmm... I see. I think I saw somewhere that resvg will be there only for those who have resvg but OpenShot will continue to use QT's thingy for it (please correct me if I am wrong) So, that means this change won't be universal? o.O Sorry I'm quite stupid. That is my mutant power. :laughing:

N3WWN:
@peanutbutterandcrackers - It's a compile time thing... if cmake can find libresvg.so and resvg.h, then libopenshot will be compiled supporting resvg as the external SVG renderer.  If cmake can't find them (or the preprocessor is set to ignore resvg), then the current internal renderer (QtSvg) is used.

ferdnyc:
I've been thinking about this some more. ...And now that I've realized this is a closed PR, I shall pick my thinkings up and leave! To PR #143, where I shall instead post them.

PR #90, opened 2018-05-10
Do not clobber gainFactor when determining volume adjustments and add…

N3WWN:
… a TODO note about current_max_volume always being 0

Resolves:
https://github.com/OpenShot/openshot-qt/issues/1533
https://github.com/OpenShot/openshot-qt/issues/1576


jonoomph:
Thanks! Another nice patch!

PR #102, opened 2018-06-05
Update wipe-tests example to latest API

ferdnyc:
`src/effects/Openshot Wipe Tests.py` was no longer usable, since the code was way out of date with respect to the current libopenshot APIs. Since examples are still useful to have, this updates the code to execute properly.

(Also: Output the frame numbers on a single line, instead of one per line, for readability.)


ferdnyc:
Never mind.

ferdnyc:
Now that things are being looked at, this one's still valid. The example is useless in its current state in the repo.

ferdnyc:
@jonoomph 

If you have a sec to look at this... the only file it changes is `src/effects/Openshot Wipe Tests.py`, which doesn't affect the library build at all. (In fact it isn't even part of the build.) But it's handy to have in the tree, and no longer works as-is since the code was years out of date.

jonoomph:
Nice! Looks good to me.

PR #127, opened 2018-06-22
Rebase to OpenShot/develop

ferdnyc:


ferdnyc:
That... was not at all what I wanted to do.

PR #128, opened 2018-06-22
Build fixes for compiling under ffmpeg 3.5 / 4.0

ferdnyc:
This is a version of the patch rpmfusion applies when packaging libopenshot-0.1.9 release, to successfully build against ffmpeg 3.5 (and now 4.0) on Fedora. It's been resynced to the source changes since 0.1.9 was released, but is otherwise identical to [their patch](https://pkgs.rpmfusion.org/cgit/free/libopenshot.git/plain/ffmpeg35_buildfix.patch).


ferdnyc:
Oh well. Still failing, but at least it's based off the OpenShot:develop branch now.

peanutbutterandcrackers:
Build man to the rescue here, too!

[libopenshot_build_271_build-and-tests.log](https://github.com/OpenShot/libopenshot/files/2168623/libopenshot_build_271_build-and-tests.log)

ferdnyc:
Excellent. Thanks, as expected it's failing in the 4.0-compatible code, which _isn't_ compatible with whatever version of FFMpeg is currently in use on the build server. With FFMpeg's moving-target API, that's far from surprising.

In theory the code could be conditionalized, but it's not worth it. The build servers are likely the only place libopenshot will ever attempt to link with a version as old as the one it's using. When the build system gets updated to a more current FFMpeg, this PR is here to make the libopenshot source compatible.

...I hadn't noticed that CMake doesn't actually _report_ the version of FFMpeg in use, though, when generating the build config. I'll see if I there's any way to fix that without too much pain. It would be nice to know what we're building with, when debugging build failures.

peanutbutterandcrackers:
Hmm... perhaps the build system should be made to echo ffmpeg --version while it is being updated/upgraded the next time? :)

I'll probably have to remember to remind Mr. Thomas regarding this too, then.

ferdnyc:
> Hmm... perhaps the build system should be made to echo ffmpeg --version while it is being updated/upgraded the next time? :)

That's... no. The command-line tool doesn't even need to be installed, to use the ffmpeg libraries. CMake should be using pkg-config to pick up ffmpeg, which will automatically report library versions as it discovers them. See [this comment](https://github.com/OpenShot/libopenshot/pull/141#issuecomment-402960956) for an example of that in action.

peanutbutterandcrackers:
Neat!

ferdnyc:
\**nod*\* It does depend on ffmpeg's libraries being properly installed, or at least having their pkgconfig files with them, because it's effectively just the output of `pkg-config --modversion libavcodec` (or whichever of the libraries). But it's safer to depend on `libavcodec.pc` being installed with `libavcodec.so`, than to assume the `ffmpeg` encoder will be available. The libraries are packaged separate from the binary by most (all?) distros. And the CMake build for libopenshot is already using `pkg-config` to pick up the configuration for JsonCpp.

ferdnyc:
So, a lot of my general musings recently have been focused on the question, "How do we make the libopenshot build process less painful for anyone who wants to download the source and compile it for themselves?" I started down that road in comments over at @N3WWN's ReSVG PR (#143), but it also ties in to this PR, and to my own CMake updates in #141. And I have a proposal:

Since we're beholden to the versions of tools and dependencies on the build server, which tend to be rather backrev, we can't merge changes into the tree that provide compatibility with the latest-and-greatest (read: "current") versions of the various inputs to the build process, if those changes in turn _break_ compatibility with the versions in use for OpenShot's own builds.

But that doesn't mean we can't distribute them.

I'm proposing we add a `patches/` directory to the source tree, and distribute patches in there that can be applied to the distributed source if building under various conditions. It's kind of an old-school approach, and not seen much anymore with the advent of build tools that manage versioning themselves, but since our build tools are also out of date and completely version-ignorant it's the recourse we have available. 

The patches would have to be hand-applied by the user, which is not ideal, but we can at least add some information to the build instructions. (Which reminds me, first we'd have to write build instructions that aren't in a 4-year-old Google Docs file. It's borderline offensive that the project README is _still_ pointing to that hopelessly out of date document, and acting as if it's at all useful.)

So, `patches/ffmpeg35-40.patch` would contain the changes in this PR, to make `src/FFmpegReader.cpp` and `src/FFmpegWriter.cpp` build against FFMpeg > 3.4. `patches/CMake3.patch` would update the various `CMakeLists.txt` with the PR #141 fixes, which are incompatible with the build server's CMake 2.8.12.2. And so on.

The goal would be to get these patches applied to the _actual_ source tree at some point, when the necessary parts of the build environment catch up to this century. But until then, it would at least give other users of the library some way to successfully build libopenshot when using current tools.

ferdnyc:
Never mind.

PR #143, opened 2018-07-11
[WIP] Use resvg library for more full-featured SVG rendering

N3WWN:
Replaces #88 so that the PR is against `libopenshot:develop` instead of `libopenshot:master`.  

See #88 for the discussion up until this point.


ferdnyc:
I've been thinking about this some more. And, fair warning, this got long.

I think the best approach, in terms of the OpenShot distribution, would be to incorporate building resvg as a library into the libopenshot build process, the same way that jsoncpp can be built locally today (in `thirdparty/jsoncpp`, unless `USE_SYSTEM_JSONCPP` is set). 

We'd have to look into the licensing and/or open a dialogue with RazrFalcon to see if we can do that _directly_, like with jsoncpp<sup>**1**</sup>. Or, there might be the possibility of configuring `thirdparty/resvg` as a [git submodule](https://blog.github.com/2016-02-01-working-with-submodules/), but that has its own [associated difficulties](https://blog.github.com/2016-02-01-working-with-submodules/#advice-on-using-submodules-or-not).

Regardless of how it gets there, I think the options should be there for `thirdparty/resvg` to contain a clone of the resvg Github repo, which will then be built into a `libresvg.so` by CMake during the overall libopenshot build process. This would, of course, be optionally disabled by both `USE_SYSTEM_RESVG=1` (which would skip the build and instead look for an already built `libresvg.so` in standard or supplied paths), and by `USE_RESVG=0` which would build libopenshot without resvg.

I just don't think libopenshot can absorb another library dependency, not in the "you're on your own, build it yourself and just point me at it when it's ready" form that nearly **ALL** of its dependencies currently take. It's not so much that libopenshot is so impossibly huge or has so many complex dependencies (that suggestion is laughable, compared to things like VLC). It's that many of its dependencies are outdated, requiring the builder to either patch it to compile with current versions of the dependency library, or to build the old version from scratch, because it's generally impossible to download (working) pre-built packages of 4-year-outdated libraries. The package management systems keep current, and expect software being installed to either do the same, or to work things out for itself. 

libopenshot does neither. 

That's why I feel that it's already teetering on the edge of being unbuildable, due to all of the dependencies involved. And with resvg, we have another problem: There doesn't appear to **be** an install process for libraries built with cargo. According to its docs / help, it can only `cargo install` crates with `bin` outputs, binary executables. `resvg` has none, and `cargo install libresvg.so` isn't a valid thing. So, there's _no_ established standard process by which `FIND_PACKAGE(RESVG)` would be made to succeed under normal circumstances. 

Which means that when libopenshot is making use of `libresvg.so`, **we're** responsible for installing it alongside `libopenshot.so` (ditto `libresvg.dll` alongside `libopenshot.dll`). Otherwise, it's going to be linked to wherever the user placed their `libresvg.so`, or to our build directory if we build it. For packages like the AppImages or the Windows installer, that's workable as our build scripts can suck it in like everything else. And doing this won't matter to the Linux distros, which will want `libresvg.so` packaged discretely according to their own standards. (Those **do** solve the installation problem, anyway, and would place `libresvg.*` in a standard location. So, this isn't for them.) But for users building libopenshot on their own, expecting them to "figure out" installing `libresvg.{so,dll}` and the headers is no help at all, and I feel like libopenshot has been neglecting those users for so long that it's already almost completely unusable to them.

resvg still adds dependencies to libopenshot, in the form of the Rust language and the `cargo` package manager. But those are tools, installable things that have well-defined processes for getting them integrated with the system. `libresvg.{so,dll}` does not, basically by design, and that's why I think libopenshot needs to take the reins on that (in the form of handling things within its own `cmake`/`make` process) if we're going to hitch our wagon to resvg.

Now that I've finished the translations-loading rewrite, I'll be happy to take a crack at hooking resvg's build and "install" process into the libopenshot CMake infrastructure.

#### Notes

1. resvg-qt is MIT-licensed, which is _very_ permissive and mercifully brief. Confusingly, though, resvg is covered under the MPL-2.0 (Mozilla Public License) which is positively sprawling. Still it contains the same basic concepts as any FLOSS license:
   > Each Contributor hereby grants You a world-wide, royalty-free, non-exclusive license: (a) under intellectual property rights (other than patent or trademark) Licensable by such Contributor to use, reproduce, make available, modify, display, perform, distribute, and otherwise exploit its Contributions, either on an unmodified basis, with Modifications, or as part of a Larger Work;

ferdnyc:
So, I built myself a copy of `libopenshot.so` with ReSVG integrated. (Built by hand, I just used CMake to link it in from `thirdparty/resvg` in the source tree, not very different from how @N3WWN already had things set up to do. _Mostly_. Read on...)

Along the way I answered a very old question we'd asked: Is it possible to build a static lib version of `libresvg`?. The answer appears to be yes! It requires modifying the `Cargo.toml` definition in the resvg source before building it with cargo, but it's possible to get cargo to generate a `libresvg.a` that can then be statically linked to `libopenshot.so`. So, that should be another option we can provide in the CMake configuration. (And one we should consider as our default for builds with resvg enabled. Though I feel strongly that we should still support building with a dynamically-linked `libresvg.so` as well. I'm not sure yet which I feel makes more sense for the packaged builds, so if anyone has any thoughts...?)

The statically-linked resvg grows the final `libopenshot.so` from 2.0 MB  to 3.2 MB, which makes sense as the stripped `libresvg.so` is 1.3 MB. So, no difference in total disk space consumed, as you'd expect.

I did have two questions for @N3WWN , if you have a moment. One is about the build, the other is about resvg integration.

On the build front, I notice that `FindRESVG.cmake` is looking for `resvg/resvg.h` as its "key" header file, and expecting it to be in `$RESVGDIR/include/`. The library, OTOH, is expected to be in `$RESVGDIR/lib/`. My question is, how did that path structure get set up? Did you just create the directories somewhere and copy the `libresvg.so` and header files into the appropriate locations? I ask because, as I said in my previous comment, I don't see _any_ sort of "install" type process ANYWHERE for resvg and/or its C api, and in my testing the files don't end up in those locations relative to each other. going into `$RESVG_SRC/capi` and running `cargo build --release --features="qt-backend"` places the `libresvg.so` (or `libresvg.a`, if you've edited for static building) output in `../target/release/` (that is, `$RESVG_SRC/target/release`), while it doesn't do _anything_ with the header files so the're still only located at `$RESVG_SRC/capi/include/resvg.h` and ``$RESVG_SRC/capi/include/ResvgQt.h` (no `resvg/` subfolder at all). 

So, I'm really trying to figure out if I'm _missing_ some install process, or if you just had to manually create the structure, (in which case CMake's build process for `resvg` can do the same, when the integrated build is selected).

I should also note that some time post-0.3.0 release, in fact only a month ago, the resvg C api was changed to be a header-only setup, with no C++ wrapper and _all_ necessary functionality implemented in Rust. (Which, in general, I think is an excellent enhancement.) So if you haven't built a fairly recent resvg, there may have been changes in the process. It's possible the headers _were_ previously installed, as a consequence of building the C++ wrapper, and so eliminating that had the side-effect of no longer "installing" the headers. In which case, we can maybe encourage RazrFalcon to add something similar back in. Because I find it extremely odd that the header files are never "delivered" into the `target/release/` build output directory, since even though they don't get modified from their source form,  writing any code that uses `libresvg.so` is quite impossible without them. It seems like the C API build would want to consolidate all of its deliverables in one place, but it does not.

...Anyway, the integration question: When ReSVG is enabled in `libopenshot.so`, is it used to render the Title Editor "preview" display as well? Or only what's shown in the OpenShot (video) preview pane?

I ask because, I loaded a test project that contains an svg Title Editor Clip to compare the SVG support. I noticed some subtle differences in the video preview render with QtSVG doing the rendering vs. ReSVG doing the rendering, which is to be expected and actually a _good_ thing, as it made it very easy for me to observe that ReSVG was indeed enabled and in use when I loaded OpenShot with that version of my `libopenshot.so`.

But what I found _more_ strange was that I **didn't** see the same differences in the Title Editor display. Those two views looked identical whether I was using QtSVG or ReSVG, only the video preview view changed. I found that odd, as I'd have expected them to both be different. Or maybe it's a PNG caching thing, and I wasn't looking at an SVG render at all in the Title Editor window? Anyway, screenshots for your perusal and comparison...

#### Main window video preview 

svg-resvg2
svg-qt2

#### Title Editor preview

svg-resvg1
svg-qt1

ferdnyc:
Heh. Looking at the two Title Editor windows stacked like that, my eyes kept tricking me into thinking the titles were off-center from each other. I had to bring them both back into GIMP and align them, then snap guides to convince myself that no, they really are identical.

image

ferdnyc:
Also, for the record, this is how the source `title.svg` for that title renders when loaded into GIMP and placed against a black background. So, as we suspected, ReSVG is rendering more correctly, and QtSVG's rendering has been broken **#ThisWholeTime**. :+1: 

image

peanutbutterandcrackers:
@ferdnyc - Since @N3WWN is not here yet, I think I'll give you [how `resvg` is currently installed following his guidance](https://gist.github.com/peanutbutterandcrackers/73d3af8040b82b1d3e64930c6562d4c6) like I did. I noticed you asked to see if you missed anything... Hope that helps.

ferdnyc:
@peanutbutterandcrackers Yes, immensely, thank you. That's exactly what I was looking for. 

(Unfortunately it also confirms everything I'd guessed/assumed above, that the installation is totally manual. Well, that only makes me all the more convince that the right thing to do is have our CMake build process for libopenshot handle the building and installation of libresvg, too... in whatever form it's used.)

peanutbutterandcrackers:
@ferdnyc - Glad I could help. I wonder if we should report an request upstream to make an automated installation system. Upstream first? :)

ferdnyc:
> I wonder if we should report an request upstream to make an automated installation system.

I think we should hold off for right now, and pick our battles. If @N3WWN comes back with the answers I'm expecting to my second question (that the Titles Editor window is still using QtSVG to render its preview, because it's rendering directly from Python, and ReSVG can only be called from `libopenshot` via the C API), then the next thing I'm going to propose is a set of SWIG Python bindings for `libresvg` allowing it to be used in place of QtSVG in our Python code as well. 

Because, my feeling is that if we're going to offer optional resvg rendering, when it's enabled it needs to be used _everywhere_. The **most** value is clearly in improving the rendered output, no question, and that's what this PR gives us already. But if we show title previews that don't match what they'll see in the rendered output, because they're being generated with two completely different SVG renderers, then it starts to feel like we're messing with the users.

...And if we **do** go down that (SWIG → Python resvg) road, I think that'll potentially be a far more important component to move upstream, or even approach RazrFalcon to be involved in and/or lead the development. I'd rather sort this aspect out first, before we start worrying about an installer for three files. (Well, five, if you include those not-yet-extant Python bindings.)

peanutbutterandcrackers:
@ferdnyc - Hmm... I see. I'll leave it all in you and @N3WWN's (and hopefully RzrFalcon's) capable hands then. :)

ferdnyc:
Anybody heard from @N3WWN  recently? Is this a planned absence? It's been nearly a full week now, seems unusual.

And as always, I'll close with my "favorite" refrain: **_WHERE'S @jonoomph ??????_**

jonoomph:
@N3WWN @ferdnyc Just circling back around to this PR, and man, I really want to improve SVG rendering in OpenShot sooner than later! So I'm still really jazzed up about this one. Currently, Qt's SVG rendering uses the SVG 1.2 Tiny standard (https://www.w3.org/TR/SVGMobile12/), which is why we lack support of many common SVG features.

I'm going to start looking into this as well.

jonoomph:
FYI: I have reached out to @RazrFAlcon, and will attempt contact. :wink:

RazrFalcon:
Hi! 

I've very interested in this idea and will help to push it as far as I can.

1. As for `USE_SYSTEM_RESVG`, resvg isn't available in any distro yet, afaik, so it's not that important for now. I mean, I would take years for, let's say, Debian to include it. And it's written in rust, which makes things a bit more complicated.
1. >resvg-qt is MIT-licensed, which is very permissive and mercifully brief. Confusingly, though, resvg is covered under the MPL-2.0

   As for now, there is no point in licensing `resvg` as MIT, because it depends on `cairo` (LGPL/MPL) and Qt (LGPL/GPL). IANAL, but MPL is basically a more permissive LGPL,  so you can link it statically without a problem.
1. >Is it possible to build a static lib version of libresvg?

   Yes. Since v0.5
1. >The statically-linked resvg grows the final libopenshot.so from 2.0 MB to 3.2 MB, which makes sense as the stripped libresvg.so is 1.3 MB.

   Half of the size is `jemalloc`. After Rust 1.32, the system one will be used by default. 
1. >So, I'm really trying to figure out if I'm missing some install process,

   There are none. After building a C-API you will get a library in `target/release`. That's it. 
1. >I think I'll give you how resvg is currently installed following his guidance like I did.

   Step 3 is unnecessary. And you should use releases/tags and not the master branch.

Feel free to ask anything.

PS: I've joined Zulip, but because of time zones the real-time chat will be problematic. So github is a preferred place for me.

src/QtImageReader.cpp
```diff
@@ -51,9 +60,65 @@ void QtImageReader::Open()
...
+
+		// Only use resvg for files with a '.svg' extension
+		if (image_file_info.suffix() == "svg") {
+			// Init resvg structs
+			resvg_render_tree *rtree = nullptr;
+			resvg_options opt;
+			resvg_init_options(&opt);
+
+			// Set resvg options (ones where the defaults aren't what we want)
+			opt.dpi = 144;
```
RazrFalcon: Why?

RazrFalcon:
Code using `ResvgQt.h` (not tested):

```cpp
#include <ResvgQt.h>

ResvgRenderer renderer(data);
if (!renderer.isValid()) {
    qDebug() << renderer.errorString();
    return;
}

QImage img(renderer.defaultSize(), QImage::Format_ARGB32_Premultiplied);
img.fill(Qt::transparent);

QPainter p(&img);
renderer.render(&p);
p.end();
```

jonoomph:
@RazrFalcon Thanks for the feedback, much appreciated! I'm starting to think that maybe I should build some Debian packaging around resvg, and include the package in our the OpenShot official PPA (for Ubuntu and Ubuntu-based distros). This should be simple enough, and then we could utilize this for our Travis and GitLab Linux builders. It doesn't have to be "Debian" perfect either... just good enough. :wink:.

... unless resvg is already available in a well known PPA that we can use.

With respects to Mac and Windows, I think it might make sense to manually install resvg on those builders, and document the steps in a README (for my future self).

That seems like a pretty reasonable way to make resvg available to all the OSes and builders/runners that are needed for OpenShot's build system.

RazrFalcon:
No PPA available, afaik. And I'm not sure that Ubuntu has the latest Rust. You need at least 1.29.

jonoomph:
@RazrFalcon I have a working local version of libopenshot using your updated syntax, and using the <ResvgQt.h> include! Now I need to work on the Debian packaging for our PPA, and see how difficult that is going to be. :yum: Thanks!

jonoomph:
I'm going to push these changes to a new branch, since I will be making a bunch of changes to it.

src/QtImageReader.cpp
N3WWN: I did not want to force the use of resvg as it was getting some push back from other developers, so I made it a compile-time option.  You can let the build system determine if resvg is available and use it if it is found, or you can override it and disable resvg outright.

RazrFalcon: No, I'm about the DPI. Why 144?

N3WWN: Oh, sorry 😁 

I don't recall exactly.  I do remember specifically wanting to set it, but I don't recall why at the moment.  (Should have included a comment)

Likely it was due to having too low of a DPI by default (lower than the DPI used elsewhere in OpenShot or libopenshot), at the time, or perhaps for antialiasing to get a smoother, less pixel-perfect rendering.

PR #145, opened 2018-07-16
Fix bug with FFmpeg > 3.2 flushing frames

jeffski:
Final frames stored in buffer need to be flushed/drained so that they
can be written to the video file. Credit to Peter for the update.


DylanC:
@jeffski - Thanks, we should really get this merged in soon! (I don't have the permissions unfortunately)

jonoomph:
LGTM! Merged. Thanks again and sorry for the delay! =)

PR #157, opened 2018-08-30
cmake: use GNUInstallDirs and fix documentation installation

alaviss:
This allows customizing various installation folders.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,5 +87,4 @@ add_subdirectory(tests)
 # Find Doxygen (used for documentation)
 include(cmake/Modules/UseDoxygen.cmake)
 
-file(GLOB_RECURSE doc_files ${CMAKE_CURRENT_BINARY_DIR}/doc/html/*.*)
-INSTALL(FILES ${doc_files} DESTINATION ${CMAKE_INSTALL_DOCDIR}/libopenshot)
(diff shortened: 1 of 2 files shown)


jonoomph:
@alaviss Is this a duplicate of #? If not, can you please edit this PR and target the following branch instead of develop: https://github.com/OpenShot/libopenshot/tree/improve-cmake. Thanks!

alaviss:
No, #159 depends on this PR. I simply merged this branch over there with a merge commit to declare dependency, but I don't think github supports that.

alaviss:
@jonoomph 
> Is there some great benefit we get from switching to GNUInstallDirs? Or is the functionality the same as our current method? Thx

Functionally it's still the same. The different is that instead of hardcoding paths like `share/doc`, it's possible for packager to modify them. This is needed for systems that doesn't follow the typical *nix file system heirarchy.

The defaults of those variable should be the same as what they replace. You can consult CMake's [`GNUInstallDirs`](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) reference for more information.

jonoomph:
@alaviss Sorry this PR was not better handled. We have some new PRs (i.e. https://github.com/OpenShot/libopenshot/pull/253) that address these same issues... or most of them I believe. I'm going to go ahead and close this one. But feel free to submit a new one if you still have issues after our latest merges, and if you don't see a quick response from me, you can also contact me at jonathan@openshot.org anytime.

PR #158, opened 2018-08-30
cmake/Modules: simplify search logic

alaviss:
This also allows greater freedom in supporting different filesystem hierarchies.

### DIFF
--- a/cmake/Modules/FindOpenShotAudio.cmake
+++ b/cmake/Modules/FindOpenShotAudio.cmake
@@ -9,9 +9,9 @@ message("$ENV{LIBOPENSHOT_AUDIO_DIR}")
 
 # Find the base directory of juce includes
 find_path(LIBOPENSHOT_AUDIO_BASE_DIR JuceHeader.h
-			PATHS $ENV{LIBOPENSHOT_AUDIO_DIR}/include/libopenshot-audio/
-			/usr/include/libopenshot-audio/
-			/usr/local/include/libopenshot-audio/ )
+			HINTS ENV LIBOPENSHOT_AUDIO_DIR
+			PATH_SUFFIXES include/libopenshot-audio
+			libopenshot-audio )
 
 # Get a list of all header file paths
 FILE(GLOB_RECURSE JUCE_HEADER_FILES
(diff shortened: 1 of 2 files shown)


jonoomph:
@alaviss This PR also makes me nervous. :grimacing:. Perhaps what we need to do make a new "cmake" branch in the official repo, and merge your 2 PRs into it so we can see what happens on all the build servers. Let me work on this real quick.

jonoomph:
@alaviss Can you please edit this PR and target the following branch instead of develop: https://github.com/OpenShot/libopenshot/tree/improve-cmake.  Thanks!

ferdnyc:
I'm very into this change, as it's a step towards letting CMake work the way it's supposed to instead of hand-coding lots of things the hard way around, something that's proved fragile many times in the past. In fact, there's still a lot more that can be done. (Using `pkgconf` to discover the installed location of things like `libavcodec/libavformat/libavutil`, `libzmq`, `UnitTest++`, and etc., for example.)

The main barrier to more effectively using CMake was previously that BuildKite was using some ancient version — I think it was `2.8.something`. Travis CI uses CMake `3.9.2`, which is actually vaguely current (cmake is up to 3.12.x) and a whole world of difference. 

(My only remaining worry is what version the GitLab build servers are using. If it's something < 3.0, that should be corrected however it has to be, and `libopenshot` should stop hamstringing itself by continuing to support ancient CMake 2.x versions.)

For example, cmake (3.0+) [can automatically handle the `lib`-vs-`lib64` path adjustment](https://cmake.org/cmake/help/v3.8/command/find_library.html?highlight=lib64), based on whether the OS uses suffixed paths. And will, according to the docs, [if the extended 3.0+ `project()` macro is used to set up the environment](https://cmake.org/cmake/help/v3.8/command/project.html#command:project). 

(When used to the full extent of its features, the `project()` definition should look something like this, with the parsing of the version lines in `include/Version.h` in the middle):
```cmake
cmake_minimum_required(VERSION 3.2)
cmake_policy(SET CMP0048 NEW)
####
#### XXX: Do include/Version.h parsing
####
project(openshot VERSION ${MAJOR_VERSION}.${MINOR_VERSION}.${BUILD_VERSION}
        LANGUAGES CXX)
```
That'll automatically set `PROJECT_VERSION` and `PROJECT_VERSION_{MAJOR,MINOR,PATCH}`, so they no longer need to be manually set.

And because `LANGUAGES CXX` is included, it'll enable [`FIND_LIBRARY_USE_LIB64_PATHS`](https://cmake.org/cmake/help/v3.8/prop_gbl/FIND_LIBRARY_USE_LIB64_PATHS.html#prop_gbl:FIND_LIBRARY_USE_LIB64_PATHS) on Fedora and friends, removing the need to add `lib64` paths to the `find_library` calls.

jonoomph:
@alaviss Sorry this PR was not better handled as well. We have some new PRs (i.e. https://github.com/OpenShot/libopenshot/pull/253) that address these same issues. I'm going to go ahead and close this one. But feel free to submit a new one if you still have issues after our latest merges, and if you don't see a quick response from me, you can also contact me at jonathan@openshot.org anytime.

PR #159, opened 2018-08-31
cmake: fix documentation installation

alaviss:
The glob approach used prior to this commit only works if the documentation has been generated **before** cmake was run.

Depends on #157

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,5 +87,4 @@ add_subdirectory(tests)
 # Find Doxygen (used for documentation)
 include(cmake/Modules/UseDoxygen.cmake)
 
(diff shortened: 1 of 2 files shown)


jonoomph:
@alaviss Thanks for the PR. We currently use CMAKE_INSTALL_PREFIX in a bunch places during our build process, which supports Linux, Mac, and Windows builds. I'm not very familiar with GNUInstallDirs, but I'm very nervous about this breaking all my build automation. 

Is there some great benefit we get from switching to GNUInstallDirs? Or is the functionality the same as our current method? Thx

jonoomph:
@alaviss Can you please edit this PR and target the following branch instead of develop: https://github.com/OpenShot/libopenshot/tree/improve-cmake. Thanks!

PR #177, opened 2018-11-17
De-limit amount of threads.

dandalton1:
I propose we remove the set limit from the function that calculates the amount of threads used.

Before my commit, the function took the maximum of the processor threads and 6 for OpenShot, and maximum of the processor threads and 12 for FFmpeg.

The code states that this was based on "performance tests," yet, if there are results from those, where are the results?

### DIFF
--- a/include/OpenMPUtilities.h
+++ b/include/OpenMPUtilities.h
@@ -32,10 +32,9 @@
 #include <stdlib.h>
 #include <string.h>
 
-// Calculate the # of OpenMP and FFmpeg Threads to allow. We are limiting both
-// of these based on our own performance tests (more is not always better).
-#define OPEN_MP_NUM_PROCESSORS (min(omp_get_num_procs(), 6))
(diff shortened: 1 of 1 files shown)


jonoomph:
FFmpeg has issues on systems with more than 16 cores based on my testing and feedback from higher end systems on AWS we tested. Sorry, I don't have all the specific tests... because most are just email feedback from users and myself verifying the results on AWS. I never dug into why > 16 threads on FFmpeg messes up stuff either... it just did. Also, as @ferdnyc mentioned, this helps keeps the memory footprint smaller as well. 

If there is a really powerful reason to change this, I'm open-minded to it, so long as we can verify systems with more than 16 threads. :wink:

jonoomph:
Closing this unless a valid reason can be raised for this change. I'm not against it, we just need to verify on higher CPU systems the behavior, and make sure this is actually needed.

PR #178, opened 2018-11-19
Remove Eclipse files and add to gitignore

### LINKED ISSUE #97: Remove .cproject and .project files
These files are environment and IDE specific and should be setup on a user by user basis - for example the .cproject file contains: `/home/jonathan/Blackmagic DeckLink SDK 9.5/Linux/include`. Please consider removing them as users will most likely want to have their own environment and IDE preferences.

ferdnyc:
They were last touched on 2015-02-07 (`.cproject`) and 2014-07-11 (`.project`), and are probably too out of date to be usable, even if they are theoretically useful when current.

Fixes #97


jonoomph:
:+1:

PR #180, opened 2018-12-15
Add a text background colored box option to the text reader

jeffski:
I'm keen to make some enhancements to the TextReader including text stroke. First I have added the ability to add a colored box around text. See the example image of how this looks:

title-sample

I didn't want to add any more params to the constructor as it already has 10 params so I added a setter method.

Looking for feedback if I am taking the right approach before doing the stroke and to see whether these enhancements are desired by maintainers/users.

### DIFF
--- a/include/TextReader.h
+++ b/include/TextReader.h
@@ -90,6 +90,7 @@ namespace openshot
 		double size;
 		string text_color;
 		string background_color;
+		string text_background_color;
 		std::shared_ptr<Magick::Image> image;
 		list<Magick::Drawable> lines;
 		bool is_open;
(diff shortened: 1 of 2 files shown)


peanutbutterandcrackers:
Wow. This looks neat! I must confess, I do not understand what this PR is going to be exactly. Does this mean we can opt in for a box around text from titles? Can the box (let's call it highlight) be keyframed to give the impression of text being highlighted using a marker? or something? I'm just curious.

But this looks neat!

Sorry. Empty vessels make much sound.

jeffski:
@peanutbutterandcrackers Yes, this PR should let users create text that looks like it has been highlighted, you can optionally add the text property by calling `SetTextBackgroundColor("#000000")` on a TextReader and set the colour to whatever you want.

I don't think you will be able to animate it using keyframes as really it is just a property of the text like bold, font or font size. All I am really doing is setting `DrawableTextUnderColor` as documented here: https://www.imagemagick.org/Magick++/Drawable.html. You can still animate the text reader to animate the text being revealed using a wipe effect (text and highlite together).

I am actually not sure if this can be used by the main Openshot-qt project as my understanding is that it creates titles using SVG files, but as an additional feature of the library for anyone else using it for their own purposes I think it is useful.

As well as an interesting looking title or highlited text I think it would be good for adding subtitles.

peanutbutterandcrackers:
@jeffski - Hmm... I see. Pretty neat work! :+1:

jonoomph:
Yikes, sorry this took so long @jeffski !!!  It got stuck at the bottom of my list of PRs for too long.

PR #182, opened 2019-01-09
Fixing Scale Mode (None) in previews

jonoomph:
Often in a preview, **Scale: None** renders differently than on final render, due to optimizations in previews which reduce the image size. Now disabling scaling works correctly in both preview and final render.


jonoomph:
This should fix https://github.com/OpenShot/openshot-qt/issues/2480

PR #185, opened 2019-01-18
Integration of resvg SVG library (optional during build)

jonoomph:
* Integration of resvg library (https://github.com/RazrFalcon/resvg/) for much improved SVG rendering features (compared to the Tiny 1.2 Specification in Qt). Fallback to Qt if resvg is not found during build.
* Major refactor of max_width and max_height for preview optimization
* Fixed many bugs related to preview resizing, with regards to cached frames
* Based on a previous PR: https://github.com/OpenShot/libopenshot/pull/143
* Special thanks to @N3WWN, @ferdnyc, and @RazrFalcon! 

It will take a while for resvg to be packaged in the various Linux distros. It is a rust library, and uses cargo for it's build process. I was unable to include the package in our PPA due to the dependency on cargo (and GitHub is internally used, which is forbidden on LaunchPad's build servers). The Debian packaging will require much more effort than I had hoped, since each cargo dependency would need to be replaced with Debian packages, etc...

### DIFF
--- a/src/QtImageReader.cpp
+++ b/src/QtImageReader.cpp
@@ -36,7 +36,7 @@
 #if USE_RESVG == 1
 	// If defined and found in CMake, utilize the libresvg for parsing
 	// SVG files and rasterizing them to QImages.
-	#include "resvg/ResvgQt.h"
+	#include "ResvgQt.h"
 #endif
 
 using namespace openshot;
--- a/src/QtImageReader.cpp
+++ b/src/QtImageReader.cpp
@@ -69,8 +69,9 @@ void QtImageReader::Open()
 #if USE_RESVG == 1
 		// If defined and found in CMake, utilize the libresvg for parsing
 		// SVG files and rasterizing them to QImages.
-		// Only use resvg for files ending in '.svg'
-		if (path.find(".svg") != std::string::npos) {
+		// Only use resvg for files ending in '.svg' or '.svgz'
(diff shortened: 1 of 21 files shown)


jonoomph:
Linux and Mac build servers have resvg manually compiled and available at the moment, so our official AppImage and DMG files will support resvg out of the box. Windows support is still being worked on (but I'm hopeful it will be available soon as well). Linux distros will take much longer before resvg is widely available in repos.

src/QtImageReader.cpp
```diff
@@ -51,17 +63,45 @@ void QtImageReader::Open()
...
-		// Attempt to open file
+		bool success = true;
 		image = std::shared_ptr<QImage>(new QImage());
-		bool success = image->load(QString::fromStdString(path));
+
+#if USE_RESVG == 1
+		// If defined and found in CMake, utilize the libresvg for parsing
+		// SVG files and rasterizing them to QImages.
+		// Only use resvg for files ending in '.svg'
+		if (path.find(".svg") != std::string::npos) {
```
RazrFalcon: You can use `resvg` for `.svgz` too.

jonoomph:
As long as you build and install libresvg.so|dylib into the following places, libopenshot will automatically detect it and link against it:

/usr/local/lib/libresvg.so   (found in target/release/*.so)
/usr/local/include/resvg/*.h   (found in capi/include/*.h files)

PR #186, opened 2019-01-19
Integrate Constant Rate Factor (CRF) for FFmpegWriter

jonoomph:
The range of the quantizer scale is 0-51: where 0 is lossless, 23 is default, and 51 is worst possible. A lower value is a higher quality and a subjectively sane range is 18-28. Consider 18 to be visually lossless or nearly so: it should look the same or nearly the same as the input but it isn't technically lossless.

When exporting to VP9, a "0 crf" will produce a lossless file.

### DIFF
--- a/cmake/Modules/FindRESVG.cmake
+++ b/cmake/Modules/FindRESVG.cmake
@@ -3,7 +3,6 @@
 # RESVG_FOUND - System has RESVG
 # RESVG_INCLUDE_DIRS - The RESVG include directories
 # RESVG_LIBRARIES - The libraries needed to use RESVG
-# RESVG_DEFINITIONS - Compiler switches required for using RESVG
 find_path ( RESVG_INCLUDE_DIR ResvgQt.h
             PATHS ${RESVGDIR}/include/resvg
                   $ENV{RESVGDIR}/include/resvg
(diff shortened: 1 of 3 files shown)


src/FFmpegWriter.cpp
```diff
@@ -146,7 +146,9 @@ void FFmpegWriter::SetVideoOptions(bool has_video, string codec, Fraction fps, i
 		info.pixel_ratio.num = pixel_ratio.num;
 		info.pixel_ratio.den = pixel_ratio.den;
 	}
-	if (bit_rate >= 1000)
+	if (bit_rate >= 1000)			// bit_rate is the bitrate in b/s
```
jonoomph: Do we still need this logic, since both set info.video_bit_rate?

src/FFmpegWriter.cpp
```diff
@@ -284,6 +287,39 @@ void FFmpegWriter::SetOption(StreamType stream, string name, string value)
 			// Buffer size
 			convert >> c->rc_buffer_size;
 
+		else if (name == "crf") {
+			// encode quality and special settings like lossless
+			// This might be better in an extra methods as more options
+			// and way to set quality are possible
+			#if  LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(55, 39, 101)
+					switch (c->codec_id) {
```
jonoomph: Can we reduce the indention on the switch statement? Seems really far over to the right. Might be another tab vs space indention issue.

PR #187, opened 2019-01-22
Some polishing for the cmake ReSVG discovery

ferdnyc:
`find_package_handle_standard_args` is supposed to set the `_FOUND` variable automatically (as the comment right above it says), as well as handling things like `REQUIRED`, `QUIETLY`, etc. It should always be run at the end of a module, for this reason.
    
This change removes the conditionals around the call, lets it handle what it's meant to handle, and defines a custom failure message for discovery that replaces the manual one in `src/CMakeList.txt`.
    
In addition, the `REQUIRED` is removed from `tests/CMakeLists.txt`, since it's _supposed_ to mark the module as required (which ReSVG is not), and was only working due to the aforementioned improper
conditional wrapping of the module's cleanup.

With these changes, on my system libopenshot still builds and `make test` still runs successfully both with and without ReSVG being discovered. 

Additionally, if only one of the two parts is missing, cmake will automatically indicate that in the failure message:
```console
$ cd /tmp/resvg/target/release/
$ ln -s ../../capi/include .
$ ls
build  examples  incremental  libresvg.d        libresvg.rlib  libusvg.rlib
deps   include   libresvg.a   libresvg_qt.rlib  libresvg.so    native
$ rm libresvg.so libresvg.a # (To my great surprise, libresvg.a would be discovered and used!)
$ cd /tmp/libopenshot/
$ rm CMakeCache.txt
$ RESVGDIR=/tmp/resvg/target/release /usr/bin/cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr
[...]
-- Could NOT find RESVG, using Qt SVG parsing instead (missing: RESVG_LIBRARY) 
```
The message correctly indicates that only `RESVG_LIBRARY` was not found, because `RESVG_INCLUDE_DIR` was still successfully discovered as `/tmp/resvg/target/release/include/`.

Also:
* `$ENV{RESVGDIR}` is added to the search list for `find_path( RESVG_LIBRARY...`, so that copying/linking the `include` dir into `target/release/` and running `RESVGDIR=/path/to/resvg/target/release/ cmake ...` will successfully discover the library
* The comment in `FindRESVG.cmake` claiming it sets `RESVG_DEFINITIONS` is removed, because it doesn't

### DIFF
--- a/cmake/Modules/FindRESVG.cmake
+++ b/cmake/Modules/FindRESVG.cmake
@@ -3,7 +3,6 @@
 # RESVG_FOUND - System has RESVG
 # RESVG_INCLUDE_DIRS - The RESVG include directories
 # RESVG_LIBRARIES - The libraries needed to use RESVG
-# RESVG_DEFINITIONS - Compiler switches required for using RESVG
 find_path ( RESVG_INCLUDE_DIR ResvgQt.h
             PATHS ${RESVGDIR}/include/resvg
                   $ENV{RESVGDIR}/include/resvg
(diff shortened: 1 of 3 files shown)


jonoomph:
Worked great on my system! Merged and hopefully all the build servers will like it also. :smile:

PR #196, opened 2019-02-27
MP3 Special Handling & Missing Frame Refactor

jonoomph:
MP3 files are unique, in that they typically contain a single video frame, and then the rest of the data is audio frames. So, I've added a condition to better support that, so OpenShot doesn't spend forever looking for missing video frames. Also, I've refactored the missing frame logic to better handle audio data and video data which is missing (I've separated the logic), and all my broken test files work great now!

### DIFF
--- a/include/Timeline.h
+++ b/include/Timeline.h
@@ -152,7 +152,6 @@ namespace openshot {
 		map<Clip*, Clip*> open_clips; ///<List of 'opened' clips on this timeline
 		list<EffectBase*> effects; ///<List of clips on this timeline
 		CacheBase *final_cache; ///<Final cache of timeline frames
-		map<Clip*, int64_t> clip_last_sample; ///<List of last audio sample referrenced by each clip object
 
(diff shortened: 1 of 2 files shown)


ferdnyc:
@jonoomph It seems this commit is causing crashes on some MP3 files with cover art. See OpenShot/openshot-qt#3043, and in particular the analysis by @PhysSong in https://github.com/OpenShot/openshot-qt/issues/3043#issuecomment-546186184 .

ferdnyc:
(The example file provided by the reporter at OpenShot/openshot-qt#3043, which causes crashes for everyone who's tried it so far, is here:

https://github.com/OpenShot/openshot-qt/files/3731505/JAWS.1975.-.Main.Title.Theme.From.Jaws.Full.HD.zip

Deleting the cover art image from the file metadata makes it importable, but as-is it'll crash OpenShot when it tries to preview or export the file.

PR #210, opened 2019-04-05
Cmake 3.1+ project updates (and 3.12+ SWIG build fixes)

### LINKED ISSUE #140: Build failure with cmake-3.12
This is a heads up issue which will become imminent once cmake-3.12 series has been released. Right now there's cmake-3.12-rc2 and that's what I used in my attempt to build libopenshot which failed like this:
```
Generating build files for openshot (0.2.0)
-- Found FFMPEG: /usr/lib64/libavformat.so;/usr/lib64/libavcodec.so;/usr/lib64/libavutil.so;/usr/lib64/libavdevice.so;/usr/lib64/libswscale.so;/usr/lib64/libavresample.so  

-- Found LIBOPENSHOT_AUDIO: /usr/lib64/libopenshot-audio.so  
'LIBOPENSHOT_AUDIO_INCLUDE_DIRS:/usr/include/libopenshot-audio'
-- Found OpenMP_C: -fopenmp (found version "4.5") 
-- Found OpenMP_CXX: -fopenmp (found version "4.5") 
-- Found OpenMP: TRUE (found version "4.5")  
-- Found ZMQ: /usr/lib64/libzmq.so  
Using embedded JsonCpp
-- Found SWIG: /usr/bin/swig (found suitable version "3.0.12", minimum required is "2.0") 
-- Found PythonLibs: /usr/lib64/libpython3.6m.so (found suitable version "3.6.5", minimum required is "3") 
-- Found PythonInterp: /usr/bin/python3.6 (found suitable version "3.6.5", minimum required is "3") 
CMake Deprecation Warning at /usr/share/cmake/Modules/UseSWIG.cmake:455 (message):
  SWIG_ADD_MODULE is deprecated.  Use SWIG_ADD_LIBRARY instead.
Call Stack (most recent call first):
  src/bindings/python/CMakeLists.txt:46 (SWIG_ADD_MODULE)


CMake Error at /usr/share/cmake/Modules/UseSWIG.cmake:475 (add_library):
  add_library cannot create target "openshot" because another target with the
  same name already exists.  The existing target is a shared library created
  in source directory
  "/var/tmp/portage/media-libs/libopenshot-0.2.0_pre/work/libopenshot-0.2.0/src".
  See documentation for policy CMP0002 for more details.
Call Stack (most recent call first):
  /usr/share/cmake/Modules/UseSWIG.cmake:456 (swig_add_library)
  src/bindings/python/CMakeLists.txt:46 (SWIG_ADD_MODULE)


CMake Deprecation Warning at /usr/share/cmake/Modules/UseSWIG.cmake:677 (message):
  SWIG_LINK_LIBRARIES is deprecated.  Use TARGET_LINK_LIBRARIES instead.
Call Stack (most recent call first):
  src/bindings/python/CMakeLists.txt:49 (SWIG_LINK_LIBRARIES)


CMake Error at /usr/share/cmake/Modules/UseSWIG.cmake:681 (message):
  Cannot find Swig library "openshot".
Call Stack (most recent call first):
  src/bindings/python/CMakeLists.txt:49 (SWIG_LINK_LIBRARIES)


CMake Error at src/bindings/python/CMakeLists.txt:62 (INSTALL):
  INSTALL TARGETS given target "_openshot" which does not exist in this
  directory.


-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE) 
```

I've verified this being a cmake-3.12 issue by downgrading to cmake-3.11.4 and successfully building libopenshot.

ferdnyc:
Like the libopenshot changes, mostly enhancements to use CMake 3.1+ features
- Enforce the minimum CMake version, and set 3.14 as max tested (as it's what I'm using)
  (Using CMake > 3.14 won't fail, instead it'll automatically set compatibility options for new 3.14+ features)
- Update Doxygen module, as (implicitly) removing 2.x compatibility options breaks it
- Configure for C++11 standard in the recommended CMake manner

However, this also contains critical fixes to the `CMakeLists.txt` files for the bindings, which currently will not build under very recent CMake versions, since `SWIG_ADD_MODULE()` is completely deprecated. However, since its replacement `swig_add_library()` didn't even exist until CMake 3.8, that logic is conditionalized e.g.:
```cmake
	### Add the SWIG interface file (which defines all the SWIG methods)
	if (CMAKE_VERSION VERSION_LESS 3.8.0)
		swig_add_module(pyopenshot python openshot.i)
	else()
		swig_add_library(pyopenshot LANGUAGE python SOURCES openshot.i)
	endif()
```
The SWIG tools didn't create real target names until very recently, but now `swig_add_library()` can — and that's why `swig_add_module()` was dropped — but that's functionality we're disabling anyway for backwards compatibility, so instead we use `${SWIG_MODULE_pyopenshot_REAL_NAME}` and similar as the target name everywhere, as that's supposed to be supported on all CMake 3.1...3.14.

Fixes #140

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 #
 # @section LICENSE
 #
-# Copyright (c) 2008-2014 OpenShot Studios, LLC
+# Copyright (c) 2008-2019 OpenShot Studios, LLC
 # <http://www.openshotstudios.com/>. This file is part of
 # OpenShot Library (libopenshot), an open-source project dedicated to
 # delivering high quality video editing and animation solutions to the
@@ -59,13 +59,17 @@ STRING(REGEX REPLACE "#define[ ]+OPENSHOT_VERSION_MINOR[ ]+([0-9]+);(.*)" "\\1" 
 STRING(REGEX REPLACE "#define[ ]+OPENSHOT_VERSION_BUILD[ ]+([0-9]+);(.*)" "\\1" BUILD_VERSION "${LINE_BUILD}")
 STRING(REGEX REPLACE "#define[ ]+OPENSHOT_VERSION_SO[ ]+([0-9]+);(.*)" "\\1" SO_VERSION "${LINE_SO}")
 
+message(STATUS "MAJOR Version: ${MAJOR_VERSION}")
+message(STATUS "MINOR Version: ${MINOR_VERSION}")
+message(STATUS "BUILD Version: ${BUILD_VERSION}")
+message(STATUS "SO/API/ABI Version: ${SO_VERSION}")
 message(STATUS "Determining Version Number - done")
 
 ################### SETUP PROJECT ###################
 PROJECT(libopenshot LANGUAGES CXX
         VERSION ${MAJOR_VERSION}.${MINOR_VERSION}.${BUILD_VERSION})
 
-message("\
+message("
 Generating build files for OpenShot
   Building ${PROJECT_NAME} (version ${PROJECT_VERSION})
   SO/API/ABI Version: ${SO_VERSION}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,7 @@ message(STATUS "SO/API/ABI Version: ${SO_VERSION}")
 message(STATUS "Determining Version Number - done")
 
 ################### SETUP PROJECT ###################
-PROJECT(libopenshot LANGUAGES CXX
+PROJECT(libopenshot LANGUAGES C CXX
         VERSION ${MAJOR_VERSION}.${MINOR_VERSION}.${BUILD_VERSION})
 
 message("
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,8 +6,8 @@
 #
 # Copyright (c) 2008-2014 OpenShot Studios, LLC
 # <http://www.openshotstudios.com/>. This file is part of
-# OpenShot Library (libopenshot), an open-source project dedicated to 
-# delivering high quality video editing and animation solutions to the 
+# OpenShot Library (libopenshot), an open-source project dedicated to
+# delivering high quality video editing and animation solutions to the
 # world. For more information visit <http://www.openshot.org/>.
 #
 # OpenShot Library (libopenshot) is free software: you can redistribute it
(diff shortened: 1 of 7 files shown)


ferdnyc:
Wow, really? An unknown `cmake_policy()` setting is an _error_!??!

They make CMake so hard to love. \*sigh* I'll fix that up.

ferdnyc:
Ohh-kay. The whole thing _built_ fine, but as Travis was nice enough to point out to me, I never actually tried `make test`. Which breaks. Fixing...

ferdnyc:
That should do it.

CMakeLists.txt
```diff
@@ -53,22 +58,27 @@ STRING(REGEX REPLACE "#define[ ]+OPENSHOT_VERSION_MAJOR[ ]+([0-9]+);(.*)" "\\1"
...
-MESSAGE("--------------------------------------------------------------")
-MESSAGE("Generating build files for ${PROJECT_NAME} (${PROJECT_VERSION})")
+PROJECT(libopenshot LANGUAGES CXX
+        VERSION ${MAJOR_VERSION}.${MINOR_VERSION}.${BUILD_VERSION})
+
+message("
+Generating build files for OpenShot
+  Building ${PROJECT_NAME} (version ${PROJECT_VERSION})
+  SO/API/ABI Version: ${SO_VERSION}
+")
```
jonoomph: @ferdnyc Can we add `ENABLE_LANGUAGE(C)`, just like I did in https://github.com/OpenShot/libopenshot-audio/pull/43 ? Thanks!

ferdnyc: Done, tho I actually added it to the `project()` line instead.

Ironically, it was my adding the `LANGUAGES CXX` specification in `project()` that screwed things up in the first place, since [CMake's docs say](https://cmake.org/cmake/help/v3.14/command/project.html):
> By default `C` and `CXX` are enabled if no language options are given.

(Oops. My fault.)

jonoomph:
Done and done! Let's see what the build servers do now. :pray:

jonoomph:
@ferdnyc Whoops, all 4 builders failed. Here is the output for Linux and Mac:

* [linux-builder-gitlab-output.txt](https://github.com/OpenShot/libopenshot/files/3049795/linux-builder-gitlab-output.txt)
* [mac-builder-gitlab-output.txt](https://github.com/OpenShot/libopenshot/files/3049796/mac-builder-gitlab-output.txt)

jonoomph:
Looks like we need to use ADD_TEST perhaps?

jonoomph:
@ferdnyc Welp, I'm having trouble figuring this one out. I did find a work-around though:
https://github.com/OpenShot/libopenshot/pull/211

PR #211, opened 2019-04-05
Fix cmake 3 error for `make test`

jonoomph:
Experimental fix for libopenshot cmake 3 support


ferdnyc:
Aha, yeah, looks like [they reserved "test" at some point](https://stackoverflow.com/questions/46638704/cmake-error-target-name-test-is-reserved-or-not-valid).

ferdnyc:
(At some point we probably should rewrite the tests to run under [CTest](https://cmake.org/cmake/help/latest/manual/ctest.1.html), and then we probably _will_ end up using the built-in "test" target.)

PR #213, opened 2019-04-06
Fixing python install logic for GitLab

jonoomph:
Fixing python install logic for GitLab (since we've changed the prefix of where it's installed). On older builds, the Python bindings were installed into `/usr/local/`, but now they are correctly installed into our `build/install-x64` path.


jonoomph:
@ferdnyc This was a fallout from some of the cmake3 changes I believe. But it was an easy fix, and I think we are back to 100% build server happiness now!

PR #215, opened 2019-04-07
Fix up JsonCPP discovery messages

ferdnyc:
The JsonCPP discovery previously claimed it "Could NOT find" the system JsonCPP, even when it never checked because `USE_SYSTEM_JSONCPP` was not enabled.

With this change it correctly reports whether it's looking for the system install or just using the built-in copy. For convenience, the message includes the name of the option that controls this behavior.


jonoomph:
Thanks! Nice

PR #216, opened 2019-04-07
Make os_test depend on openshot-test

ferdnyc:
Previously if you ran `make os_test` immediately after generating the build tree, it would run successfully and ignore the fact that the `openshot-test` executable hadn't been built yet.

This change makes the "openshot-test" target a dependency of "os_test", so `make os_test` will trigger the full build process for the test executable, and even the library itself, if they haven't already been run.


ferdnyc:
Scratch this one, #217 is a better solution.

PR #217, opened 2019-04-07
Make os_test use openshot-test target

ferdnyc:
By specifying `COMMAND openshot-test` for the custom target,
the path to the compiled executable is automatically used
and the openshot-test target is automatically made a dependency.


ferdnyc:
(The dependency is important because prior to this change, if you generated the build directory and then immediately ran `make os_test`, it would **_succeed_** despite the fact that the `openshot-test` executable hadn't even been built yet. Now it'll trigger the full build process if necessary.)

PR #219, opened 2019-04-13
Add control over the output color space etc.

SuslikV:
Just wonder will it work or not? (I can't build right now, so will see if automation build will pass this)

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -1190,6 +1190,10 @@ bool FFmpegReader::CheckSeek(bool is_video) {
 	return is_seeking;
 }
 
+#define FIXED_1_0 (1 << 16)
+// Video PC, jpeg, Full color range
+#define Full_Range 1
+
 // Process a video packet
 void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
 	// Calculate current frame #
@@ -1315,8 +1319,23 @@ void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
 		if (openshot::Settings::Instance()->HIGH_QUALITY_SCALING) {
 			scale_mode = SWS_LANCZOS;
 		}
-		SwsContext *img_convert_ctx = sws_getContext(info.width, info.height, AV_GET_CODEC_PIXEL_FORMAT(pStream, pCodecCtx), width,
-															  height, PIX_FMT_RGBA, scale_mode, NULL, NULL, NULL);
+		
+		// In case of 4:4:4 inputs, FFmpeg scaling optimization for fast_bilinear
+		// makes undesirable horizontal chroma simplifications.
+		// Do not reuse chroma for 2 pixels RGB/BGR in this case.
+		scale_mode |= SWS_FULL_CHR_H_INT;
+
+		SwsContext *img_convert_ctx = sws_getContext(info.width, info.height, pCodecCtx->pix_fmt, width,
+				height, PIX_FMT_RGBA, scale_mode, NULL, NULL, NULL);
+
+		// Scaling uses own numeration for color ranges Partial (mpeg, tv) = 0; Full (jpeg, pc) = 1;
+		// thus assuming that Partial color range was set.
+		int s_color_range = 0;
+		if (pCodecCtx->color_range == 2) s_color_range = Full_Range;
+
+		// Set decoding color details before scaling
+		sws_setColorspaceDetails(img_convert_ctx, sws_getCoefficients(pCodecCtx->colorspace), s_color_range,
+				sws_getCoefficients(SWS_CS_ITU709), Full_Range, 0, FIXED_1_0, FIXED_1_0);
 
 		// Resize / Convert to RGB
 		sws_scale(img_convert_ctx, my_frame->data, my_frame->linesize, 0,
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -2004,6 +2004,10 @@ void FFmpegWriter::OutputStreamInfo() {
 	av_dump_format(oc, 0, path.c_str(), 1);
 }
 
+#define FIXED_1_0 (1 << 16)
+// Video PC, jpeg, Full color range
+#define Full_Range 1
+
 // Init a collection of software rescalers (thread safe)
 void FFmpegWriter::InitScalers(int source_width, int source_height) {
 	int scale_mode = SWS_FAST_BILINEAR;
@@ -2020,9 +2024,18 @@ void FFmpegWriter::InitScalers(int source_width, int source_height) {
 		} else
 #endif
 		{
-			img_convert_ctx = sws_getContext(source_width, source_height, PIX_FMT_RGBA, info.width, info.height, AV_GET_CODEC_PIXEL_FORMAT(video_st, video_st->codec), SWS_BILINEAR,
+			img_convert_ctx = sws_getContext(source_width, source_height, PIX_FMT_RGBA, info.width, info.height, video_codec->pix_fmt, SWS_BILINEAR,
 											 NULL, NULL, NULL);
 		}
+
+		// Scaling uses own numeration for color ranges Partial (mpeg, tv) = 0; Full (jpeg, pc) = 1;
+		// thus assuming that Partial color range was set.
+		int d_color_range = 0;
+		if (video_codec->color_range == 2) d_color_range = Full_Range;
+
+		// Set encoding color details before scaling
+		sws_setColorspaceDetails(img_convert_ctx, sws_getCoefficients(SWS_CS_ITU709), Full_Range,
+				sws_getCoefficients(video_codec->colorspace), d_color_range, 0, FIXED_1_0, FIXED_1_0);
 
 		// Add rescaler to vector
 		image_rescalers.push_back(img_convert_ctx);
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -433,6 +433,7 @@ void FFmpegWriter::SetOption(StreamType stream, string name, string value) {
 						av_opt_set_int(c->priv_data, "crf", min(stoi(value), 51), 0); // 0-51
 						if (stoi(value) == 0) {
 							av_opt_set(c->priv_data, "preset", "veryslow", 0);
+							c->pix_fmt = PIX_FMT_YUV444P;
 						}
 						break;
 					case AV_CODEC_ID_H265 :
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -344,7 +344,8 @@ void FFmpegWriter::SetOption(StreamType stream, string name, string value) {
 	// Was option found?
 	if (option || (name == "g" || name == "qmin" || name == "qmax" || name == "max_b_frames" || name == "mb_decision" ||
 				   name == "level" || name == "profile" || name == "slices" || name == "rc_min_rate" || name == "rc_max_rate" ||
-				   name == "crf")) {
+				   name == "crf" || name == "color_range" || name == "color_primaries" || name == "color_trc" ||
+		      		   name == "colorspace")) {
 		// Check for specific named options
 		if (name == "g")
 			// Set gop_size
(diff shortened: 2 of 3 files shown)


SuslikV:
OK. Maybe other time...

SuslikV:
I know nothing about the project, so it's more like concept.

jonoomph:
@SuslikV So did your test work out? What exactly does this PR do? Just curious.

SuslikV:
Nope yet. This info is just to turn your attention that OpenShot clips loaded to YouTube and viewed back in Chrome (on Windows) look on my system different (green and magenta has most noticeable difference). That's all. Maybe it's not correct to specify it like this, and  parameters should be filled somewhere else.

ferdnyc:
@SuslikV I was going to reply here and comment on the PR, but I'm realizing now my comments are less about the code than about your request / colorspace issues in general, so I'll meet you over at OpenShot/openshot-qt#2711

ferdnyc:
(Sorry, still typing. It'll be there eventually.) :wink:

SuslikV:
^_^
```
I'm here and waiting for you
Where are you, I can't find you
I'm here and waiting for you
I'll wait forever for you
```
@ferdnyc hardly I can answer in real time. You may not to hurry.

SuslikV:
OK. Finally, I did some tests and this is not the right place where to set the output color_range, color_primaries, color_trc, colorspace for libopenshot - it just sets flags but not uses this values during encoding. Maybe I will look how it works, and what is needed to be done to set it right.

SuslikV:
The only way I found to set bt709 color space, partial (mpeg/tv) color range output in libopenshot with small amount of code changes is something like:
```
sws_setColorspaceDetails(img_convert_ctx, sws_getCoefficients(SWS_CS_DEFAULT), 1, sws_getCoefficients(SWS_CS_ITU709), 0, 0, 1 << 16, 1 << 16);
```
placed right after the `img_convert_ctx` assigned.

SuslikV:
I will update PR in few days with history modifications, so need to reopen it first. If you don't mind of course. There will be solutions for better colors support of non-subsampled color video import; non-subsampled color full range lossless export (at least H264 encoder); file color details flagging (at least for popular mp4 output).

To use remaining space, I have a question - where OpenShot multiplexing its streams? I need fast_write or fragmented writing of MP4/MOV. Please, give me a point, just to not read whole code here.

SuslikV:
File for experiments (2 sec color range table +2 sec color table) 4:4:4 crf=0, H264 encoder made with similar changes to the libopenshot:

about 1 MB to download, CRC of the zip archive MD5:C49C722835CFDD58E3E948A2C1CC9A15
[TestColorDetails_LosslessCFR0.zip](https://github.com/OpenShot/libopenshot/files/3263466/TestColorDetails_LosslessCFR0.zip)

Don't forget to set _Scale_ to _None_ when previewing the file. Some players by default, will use subsampling /range optimizations, so usually it is hard to get right result from the first try.

Please, check if it works for your builds of OpenShot.

MediaInfo of the file is under the spoiler

<details>

```
General
Complete name                            : O:\Temp\OpenShotCrashTest\TestColorDetails_LosslessCFR0.mp4
Format                                   : MPEG-4
Format profile                           : Base Media
Codec ID                                 : isom (isom/iso2/avc1/mp41)
File size                                : 1.15 MiB
Duration                                 : 4 s 40 ms
Overall bit rate mode                    : Variable
Overall bit rate                         : 2 390 kb/s
Writing application                      : Lavf57.83.100

Video
ID                                       : 1
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : High 4:4:4 Predictive@L5
Format settings                          : CABAC / 16 Ref Frames
Format settings, CABAC                   : Yes
Format settings, RefFrames               : 16 frames
Format settings, GOP                     : M=1, N=12
Codec ID                                 : avc1
Codec ID/Info                            : Advanced Video Coding
Duration                                 : 4 s 40 ms
Bit rate                                 : 2 380 kb/s
Width                                    : 1 280 pixels
Height                                   : 720 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 25.000 FPS
Color space                              : YUV
Chroma subsampling                       : 4:4:4
Bit depth                                : 8 bits
Scan type                                : Progressive
Bits/(Pixel*Frame)                       : 0.103
Stream size                              : 1.15 MiB (100%)
Writing library                          : x264 core 155 r2901 7d0ff22
Encoding settings                        : cabac=1 / ref=16 / deblock=1:0:0 / analyse=0x1:0x131 / me=umh / subme=9 / psy=0 / mixed_ref=1 / me_range=24 / chroma_me=1 / trellis=0 / 8x8dct=0 / cqm=0 / deadzone=21,11 / fast_pskip=0 / chroma_qp_offset=0 / threads=2 / lookahead_threads=1 / sliced_threads=0 / nr=0 / decimate=1 / interlaced=0 / bluray_compat=0 / constrained_intra=0 / bframes=0 / weightp=2 / keyint=12 / keyint_min=1 / scenecut=40 / intra_refresh=0 / rc=cqp / mbtree=0 / qp=0
Color range                              : Full
Color primaries                          : BT.709
Transfer characteristics                 : BT.709
Matrix coefficients                      : BT.709

Audio
ID                                       : 2
Format                                   : AAC
Format/Info                              : Advanced Audio Codec
Format profile                           : LC
Codec ID                                 : mp4a-40-2
Duration                                 : 4 s 27 ms
Bit rate mode                            : Variable
Bit rate                                 : 2 288 b/s
Maximum bit rate                         : 96.0 kb/s
Channel(s)                               : 2 channels
Channel positions                        : Front: L R
Sampling rate                            : 48.0 kHz
Frame rate                               : 46.875 FPS (1024 SPF)
Compression mode                         : Lossy
Stream size                              : 1.13 KiB (0%)
Default                                  : Yes
Alternate group                          : 1


```

</details>

jonoomph:
Looking at this PR now... It's a lot to unpack though... My primary concern is backwards compatibility and stability with respect to default video export options, and openshot-qt support.

src/FFmpegReader.cpp
```diff
@@ -1315,8 +1319,23 @@ void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
...
 		}
-		SwsContext *img_convert_ctx = sws_getContext(info.width, info.height, AV_GET_CODEC_PIXEL_FORMAT(pStream, pCodecCtx), width,
-															  height, PIX_FMT_RGBA, scale_mode, NULL, NULL, NULL);
+		
+		// In case of 4:4:4 inputs, FFmpeg scaling optimization for fast_bilinear
+		// makes undesirable horizontal chroma simplifications.
+		// Do not reuse chroma for 2 pixels RGB/BGR in this case.
+		scale_mode |= SWS_FULL_CHR_H_INT;
+
+		SwsContext *img_convert_ctx = sws_getContext(info.width, info.height, pCodecCtx->pix_fmt, width,
```
jonoomph: pCodecCtx-pix_fmt should be `AV_GET_CODEC_PIXEL_FORMAT(pStream, pCodecCtx)`. This allows us to build against other versions of LibAV and FFmpeg.

src/FFmpegReader.cpp
```diff
@@ -1315,8 +1319,23 @@ void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
...
+				height, PIX_FMT_RGBA, scale_mode, NULL, NULL, NULL);
+
+		// Scaling uses own numeration for color ranges Partial (mpeg, tv) = 0; Full (jpeg, pc) = 1;
+		// thus assuming that Partial color range was set.
+		int s_color_range = 0;
+		if (pCodecCtx->color_range == 2) s_color_range = Full_Range;
```
jonoomph: Is `->color_range` supported in FFmpeg 2.x, 3.x, and 4.x? Just looks like a similar issue to pix_fmt, where FFmpeg refactors it somewhere else at some point.

src/FFmpegWriter.cpp
```diff
@@ -457,6 +459,65 @@ void FFmpegWriter::SetOption(StreamType stream, string name, string value) {
...
+		//   SetOption(openshot.VIDEO_STREAM, "colorspace", str(int(1)) )
+		//
+		// the first one for convenience, the later - for future use.
+		//
+		// To preserve current OpenShot behavior and for compatibility with old hardware and software,
+		// do not set colorspace details if BT.601 standard is in use.
+		// The default color details of ITU-R BT601-6 625, ITU-R BT1358 625, ITU-R BT1700 625 PAL & SECAM, IEC 61966-2-4 xvYCC601
+		// will be applied in this case without setting file container flags (mainly used for SD quality videos).
+#if LIBAVFORMAT_VERSION_MAJOR >= 53
+		else if (c->codec_type == AVMEDIA_TYPE_VIDEO) {
```
jonoomph: codec_type should be `AV_GET_CODEC_TYPE(av_stream)`, again for the same reasons... compatibility with different versions of FFmpeg.

src/FFmpegWriter.cpp
```diff
@@ -2019,10 +2084,19 @@ void FFmpegWriter::InitScalers(int source_width, int source_height) {
 		} else
 #endif
 		{
-			img_convert_ctx = sws_getContext(source_width, source_height, PIX_FMT_RGBA, info.width, info.height, AV_GET_CODEC_PIXEL_FORMAT(video_st, video_st->codec), SWS_BILINEAR,
+			img_convert_ctx = sws_getContext(source_width, source_height, PIX_FMT_RGBA, info.width, info.height, video_codec->pix_fmt, SWS_BILINEAR,
```
jonoomph: Same thing again: `AV_GET_CODEC_PIXEL_FORMAT(video_st, video_st->codec)` instead of pix_fmt

jonoomph:
@SuslikV This mostly looks good to me... but I believe it will fail to compile again certain versions of LibAV and FFmepg. I've added some comments, but they should be quick changes. Nice!

src/FFmpegReader.cpp
SuslikV: Available in all versions up to 1.0 (at least as storage) as far as I remember.

SuslikV:
I'm not common with all variations of LibAV, though. Solution based on FFmpeg mainly.

src/FFmpegWriter.cpp
SuslikV: And again, I don't understand you. The `c->` already points to codec context here, so it has `codec_type` field. For what reason I should re-use the stream's pointer to the same structure? The field itself available in all FFmpeg versions of AVCodecContext.

Enumeration may has different name - but I can use just `0` to make simple and robust comparison (and leave enumeration's name as comment in the code). Is this solution better? No excessive `ifdef` constructions. Usually, when enumeration changes - new values added to the end, so video always will be 0.

jonoomph: `av_stream->codecpar->codec_type`  vs  
`av_stream->codec->codec_type`

For some reason, LibAV/FFmpeg devs love to refactor things and rename things. Which path to use is based in FFmpegUtilities.h, which we only built to get past certain "version-based" differences such as this. Hope that makes sense.

jonoomph:
@SuslikV I'm also not familiar with all variations, but we built FFmpegUtilities.h to deal with the differences as we encounter them. So if you something wrapped by FFmpegUtilities.h, that is why. :wink:

src/FFmpegWriter.cpp
SuslikV: `c` - is pointer to AVCodecContext structure
declaration:
https://github.com/OpenShot/libopenshot/blob/cf9fbf4402e4a154c741c383b24b6df306e27cd7/src/FFmpegWriter.cpp#L327
definition:
https://github.com/OpenShot/libopenshot/blob/cf9fbf4402e4a154c741c383b24b6df306e27cd7/src/FFmpegWriter.cpp#L334
It's a few lines before my code. All structure entries I use is always part of the AVCodecContext.

For what reason you are create the AV_GET_CODEC_TYPE and why should I use it **exactly** here, if I'm already working with the codec context? Because it is still requires to check if the:

`codec_type` is part of the `enum CodecType` (https://ffmpeg.org/doxygen/0.5/structAVCodecContext.html) or
`codec_type` is part of the `enum AVMediaType` (https://ffmpeg.org/doxygen/1.0/structAVCodecContext.html)

while the AV_GET_CODEC_TYPE still uses same codec context pointer (`av_stream->codec->...`) for all LIBAVFORMAT_VERSION_MAJOR = 55 and below.

If you think that using the AV_GET_CODEC_TYPE is robust solution - then it is not (just different place of the parameter). And sure, I think the `#if LIBAVFORMAT_VERSION_MAJOR >= 53` of mine should be removed to not confuse anybody. I'm start thinking that hard coded zero is the best solution here.

jonoomph:
I'm going to close this PR for now (house cleaning), but if you implement the changes mentioned above, please reopen and let's take another look. Thanks.

SuslikV:
Video Editor without color management is like a bird without the song. I would not like to make changes that has wrong assumptions of robustness.

PR #223, opened 2019-04-18
Travis CI: Also run `make install`

ferdnyc:
Taking my own suggestion, this adds a `make install` step to all three (overkill, maybe?) Travis builds, using `DESTDIR=dist/` to keep the files local to the build dir. Meant to bring the CMake-managed install logic under CI verification.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,8 @@ matrix:
       - cmake -D"CMAKE_BUILD_TYPE:STRING=Debug" ../
       - make VERBOSE=1
       - make os_test
-
+      - make install DESTDIR=dist/
+      
     - language: cpp
       name: "FFmpeg 3"
       before_script:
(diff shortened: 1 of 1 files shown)


ferdnyc:
I'm already glad I thought of this, because I scared myself for a second when I saw:

`-- Installing: dist/usr/local/lib/python3/dist-packages/_openshot.so` :scream:

...Then I remembered, the fix for _that_ is in a separate PR which hasn't been merged yet. But point is, we're seeing more of the build chain exercised, which is a good thing!

jonoomph:
Indeed! Yes, I'll merge this very soon.

jonoomph:
Sorry this one took so long... yikes.

ferdnyc:
@jonoomph Speaking of, I don't know if you saw but I finally did get around to accepting those project invitations and setting up my account @ the GitLab instance. Things there seem... very sparse, currently (in the sense that I can see literally _nothing_) so I assume there's setup that needs to be done now that the account's active?

PR #226, opened 2019-05-02
Various cmake improvements for Windows

SuslikV:
Some cmake improvements that simplifies building routine for MinGW (Windows).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,9 @@ list(APPEND OPENSHOT_INCLUDE_DIRS
 #### Enable C++11 (for std::shared_ptr support)
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
+
+#### Optional C++ restrictions
+set(CMAKE_CXX_EXTENSIONS OFF CACHE BOOL "Allow compiler C++ language extensions")
 
 IF (WIN32)
 	SET_PROPERTY(GLOBAL PROPERTY WIN32 "WIN32")


SuslikV:
C++ language extensions usage changes is similar to https://github.com/OpenShot/libopenshot-audio/pull/48

CMakeLists.txt
```diff
@@ -78,12 +78,32 @@ Generating build files for OpenShot
...
-set(CMAKE_CXX_EXTENSIONS OFF)
+
+#### Optional C++ restrictions
+set(CMAKE_CXX_EXTENSIONS OFF CACHE BOOL "Allow compiler C++ language extensions")
 
 IF (WIN32)
 	SET_PROPERTY(GLOBAL PROPERTY WIN32 "WIN32")
 ENDIF(WIN32)
 
+#### Define _WIN32_WINNT for some odd macros
```
jonoomph: Is this an appropriate place for this type of logic? Seems like it would be handled in cmake perhaps, right? In other words, it seems odd OpenShot would set all sorts of _WIN32_WINNT values, just in case some module expects it? Maybe I just don't quite understand.

Do we know of anything that actually requires these? If so, what?

src/CMakeLists.txt
```diff
@@ -399,10 +399,14 @@ add_subdirectory(bindings)
 set(LIB_INSTALL_DIR lib${LIB_SUFFIX}) # determine correct lib folder
 
 # Install primary library
-INSTALL(  TARGETS openshot
```
jonoomph: This also looks suspect to me. Perhaps @ferdnyc would have some thoughts. In theory, cmake should be doing this type of stuff automatically... with respects to handling Mac, Windows, and Linux in a single way... vs having lots of IF WIN, IF MAC type logic all over the place. But I would love to hear what @ferdnyc thinks.

jonoomph:
Added some comments, but in summary, this looks a bit strange since cmake should be making our lives easier not harder. :smiley:

CMakeLists.txt
SuslikV: Example of the error `C:/msys64/mingw64/include/zmq.h:71:2: error: #error You need at least Windows XP target`
https://ci.appveyor.com/project/SuslikV/libopenshot/builds/24238985
the mentioned in the log-file the `zmq.h` defines the  _WIN32_WINNT variable by itself if it is not exist (but in some cases only). And the log from the server version OS (Windows Server 2012 R2).

src/CMakeLists.txt
SuslikV: Don't know what it should, but I don't see my recently compiled .dll file in the destination folder of the installation path.
The idea behind this was to build (and it does) and copy the .dll (not only .dll.a) and insert it into frozen OpenShot, just to not recompile whole project each time I need to check if my new lovely feature works.

SuslikV:
In which way it makes your life harder?

src/CMakeLists.txt
SuslikV: Few thoughts from @ferdnyc , similar code https://github.com/OpenShot/libopenshot-audio/pull/52 ^_^

ferdnyc: Yeah, the DLL not being installed was the missing `RUNTIME DESTINATION` path in the `install()`, which I've submitted PRs to fix in both projects.

If a _single_ destination is specified `install(TARGETS openshot DESTINATION $path)`, everything gets installed there. But if the various products of the target are routed with `ARCHIVE DESTINATION`, `LIBRARY DESTINATION`, etc. then any not specified _aren't_ installed. So we can't leave out `RUNTIME DESTINATION` for DLLs.

But there's absolutely no need to make it conditional by OS.
```
install(TARGETS openshot
		ARCHIVE DESTINATION ${LIB_INSTALL_DIR}
		LIBRARY DESTINATION ${LIB_INSTALL_DIR}
		RUNTIME DESTINATION ${LIB_INSTALL_DIR} )
```
(or, better, `${CMAKE_INSTALL_LIBDIR}` from GnuInstallDirs) can be used on all three OSes.

CMakeLists.txt
ferdnyc: @SuslikV `_WINNT_WIN32` isn't something the user is meant to set themselves, it's detected from the system environment. `zmq.h` only defines it if it's _not_ defined for some reason — in those cases, it'll set it to a reasonable default. Adding some indentation:
```cpp
#if defined _WIN32
    //  Set target version to Windows Server 2008, Windows Vista or higher.
    //  Windows XP (0x0501) is supported but without client & server socket types.
    #ifndef _WIN32_WINNT
        #define _WIN32_WINNT 0x0600
    #endif

    #ifdef __MINGW32__
        //  Require Windows XP or higher with MinGW for getaddrinfo().
        #if (_WIN32_WINNT >= 0x0501)
        #else
            #error You need at least Windows XP target
    #endif
#endif
```
...The only way you could be getting that error message is if `_WIN32_WINNT` is **set** in your environment, and it's set to a value < `0x0501`. Setting it from CMake isn't the right fix at all.

ferdnyc: My guess is, the issue is this, from your AppVeyor output:
```
OS=Windows_NT
```
If the system profile is set up for straight Windows NT (which I think was `0x0400`), that's _not_ greater than the minimum "Windows XP or higher" requirement.

ferdnyc: I edited `src/examples/Example.cpp` to check the built-in `#define`:
```cpp
#include <fstream>
#include <iostream>
#include <memory>
#include "../../include/OpenShot.h"
#include "../../include/CrashHandler.h"

using namespace openshot;


int main(int argc, char* argv[]) {


        cout << "_WIN32_WINNT: " << _WIN32_WINNT << endl;
        return(0);

        Settings *s = Settings::Instance();
        s->HARDWARE_DECODER = 2; // 1 VA-API, 2 NVDEC
        s->HW_DE_DEVICE_SET = 0;
```
And built `openshot-example` on my Win7 VM:
```cmd
> path=C:\msys64\mingw64\bin;C:\msys64\mingw64\lib;%PATH%
> cmake -D"CMAKE_INSTALL_PREFIX:PATH=C:\msys64\home\ferd\libopenshot\build\install-x64" -G "MinGW Makefiles" -D"CMAKE_BUILD_TYPE:STRING=Release" ../
> mingw32-make openshot-example
> copy ..\..\libopenshot-audio\build\libopenshot-audio.dll src\
> src\openshot-example
_WIN32_WINNT: 1537
```
(Which equals `0x0601`.) I'd try the same thing in AppVeyor, find out what `_WIN32_WINNT` is set to. But you can't just force it to a different value, a Windows build environment below version `0x0501` won't have all the required system API support.

SuslikV: @ferdnyc I probably third time say that this is with  Windows Server 2012 R2 image. How it could be **lower** than Windows Server 2008?

src/CMakeLists.txt
SuslikV: @ferdnyc for what reason are you set 3 **same** paths in a single rule?

CMakeLists.txt
SuslikV: I have other question here @jonoomph . Is the zmq really required for release builds of the application, for example?

ferdnyc: > @ferdnyc I probably third time say that this is with Windows Server 2012 R2 image. How it could be **lower** than Windows Server 2008?

I don't know, so test it with something like my `Example.cpp` test. Because, like I said, there's no other way you could be getting that error message. If `_WIN32_WINNT` is _undefined_, `zmq.h` will **set it** to `0x0600`:
```cpp
    #ifndef _WIN32_WINNT
        #define _WIN32_WINNT 0x0600
    #endif
```

So, the problem cannot possibly be that `_WIN32_WINNT` is not set.

ferdnyc: ZeroMQ is definitely required for all builds, it's the messaging library that `libopenshot` uses for logging (and passing those logs to OpenShot).

src/CMakeLists.txt
ferdnyc: They control where different products of the target will be installed, in different scenarios, on different OSes. (We just happen to install them all to the same place.) [CMake install():](https://cmake.org/cmake/help/v3.0/command/install.html)
> Static libraries are always treated as `ARCHIVE` targets. Module libraries are always treated as `LIBRARY` targets. For non-DLL platforms shared libraries are treated as `LIBRARY` targets, except that those marked with the `FRAMEWORK` property are treated as `FRAMEWORK` targets on OS X. For DLL platforms the DLL part of a shared library is treated as a `RUNTIME` target and the corresponding import library is treated as an `ARCHIVE` target.

Technically we only care about `LIBRARY DESTINATION` (for Linux `.so` and OS X `.dylib`) and `RUNTIME DESTINATION` (for Windows `.dll`). The `ARCHIVE DESTINATION` will get the Windows `.dll.a` file, though, and would also get any static libraries built on Linux/Mac. (`FRAMEWORK DESTINATION` is also a possibility if we ever build the MacOS library as a Framework, though I know next-to-nothing about that.)

The `RUNTIME DESTINATION` actually _should be_ `${CMAKE_INSTALL_BINDIR}`, so that Windows DLLs would be installed alongside executables in the `bin/` directory. That's how Windows really expects things to be set up, and it's a "more correct" layout. OpenShot on Windows is ultimately packaged that way by cx_freeze, in fact: all of the `.dll`s are placed in the `C:\Program Files\OpenShot Video Editor\` directory alongside the `.exe`. 

But for now, the DLL is still being routed into the library dir, because the previous single-path `install()` command placed it there, so that's the assumption all of the packaging scripts and etc. were written with.

CMakeLists.txt
SuslikV: @ferdnyc Logs are bloat in OpenShot with useless info. I don't know why they all not excluded during compilation (at least in release mode). It reminds me a big beta-testing rather that working solution.

And as for the mentioned test, I have some difficulties, because I explicitly specifying destination platform as  `-DCMAKE_SYSTEM_NAME=Windows -DCMAKE_SYSTEM_VERSION=6.1` during building and I got result 601h that is quite normal I think.

ferdnyc: That's the thing, `_WIN32_WINNT` has nothing to do with the host system, it's an API level identifier. 

MinGW should be setting that to `0x601` on any 64-bit Windows install, it's in `C:\MSYS64\MinGW64\x86_64-w64-mingw32\include\_mingw.h`. That header's automatically included in all builds. It's installed with the `mingw-w64-x86_64-headers-git` package, which is also part of the `mingw-w64-x86_64-toolchain` group. I believe it's also installed as a dependency of `mingw-w64-x86_64-gcc`. Perhaps you need to update the build environment. But you shouldn't need to set any of these things by hand. 

That's what @jonoomph was referring to about "making things harder" — if these variables aren't set by the environment, it's a sign that the _environment_ isn't set up properly, not that they need to be set in the build config. Doing that is just masking the real problem.

ferdnyc: Oh, and in that build you linked to, you're still setting `_WIN32_WINNT` on the command line:
<pre style="white-space:pre-wrap;"><code style="white-space:pre-wrap;">
[ 78%] Building CXX object src/CMakeFiles/openshot-example.dir/examples/Example.cpp.obj
cd /d C:\projects\libopenshot\build\src && 
C:\mingw-w64\x86_64-7.3.0-posix-seh-rt_v5-rev0\mingw64\bin\g++.exe
  -DIGNORE_JUCE_HYPOT=1 -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16
 -DOPENSHOT_IMAGEMAGICK_COMPATIBILITY=0 -DQT_CORE_LIB -DQT_GUI_LIB
 -DQT_MULTIMEDIAWIDGETS_LIB -DQT_MULTIMEDIA_LIB -DQT_NETWORK_LIB
 -DQT_NO_DEBUG -DQT_WIDGETS_LIB <b>-D_WIN32_WINNT=0x0601</b>
 @CMakeFiles/openshot-example.dir/includes_CXX.rsp -include cmath           -g -ggdb 
 -fopenmp    -std=c++11 -o CMakeFiles\openshot-example.dir\examples\Example.cpp.obj
 -c C:\projects\libopenshot\src\examples\Example.cpp

[ 80%] Linking CXX executable openshot-example.exe
cd /d C:\projects\libopenshot\build\src && 
"C:\Program Files (x86)\CMake\bin\cmake.exe"
 -E cmake_link_script CMakeFiles\openshot-example.dir\link.txt --verbose=1
"C:\Program Files (x86)\CMake\bin\cmake.exe" -E remove -f CMakeFiles\openshot-example.dir/objects.a
C:\mingw-w64\x86_64-7.3.0-posix-seh-rt_v5-rev0\mingw64\bin\ar.exe cr
 CMakeFiles\openshot-example.dir/objects.a @CMakeFiles\openshot-example.dir\objects1.rsp
C:\mingw-w64\x86_64-7.3.0-posix-seh-rt_v5-rev0\mingw64\bin\g++.exe   -include cmath 
          -g -ggdb  -fopenmp    -Wl,--whole-archive CMakeFiles\openshot-example.dir/objects.a
 -Wl,--no-whole-archive  -o openshot-example.exe -Wl,--out-implib,libopenshot-example.dll.a
 -Wl,--major-image-version,0,--minor-image-version,0
 @CMakeFiles\openshot-example.dir\linklibs.rsp
</code></pre>
You'd need to remove that, to test the initial value.

src/CMakeLists.txt
SuslikV: @ferdnyc I see only single path for all (despite it written 3 times). And I think that the 3 paths - is just none existing future.

About OS specific code. I did'n knew why there were different install scripts (libopenshot and libopenshot-audio) for one application, so just escaped unknown for me part with the OS specific code, that's all. Now I'm sure that this is just rudiment of the long development.

CMakeLists.txt
SuslikV: @ferdnyc Sometimes I think that OpenShot was designed to sit only in crystal clear environment of the ideal OS. I will show to you the value of the _WIN32_WINNT for the given environment sooner or later, and maybe you will tell me what to do with it to build the OpenShot (or how to make it to build) in the given environment (if I'm doing it wrong). I know that I'm doing something wrong, because it doesn't work out of the box - it didn't compile (if you try to follow the wiki instructions), dependencies has opened bugs, installation hasn't required libs, compatibility with python version not mentioned, host OS doesn't match destination and so on.

SuslikV: @ferdnyc 
https://ci.appveyor.com/project/SuslikV/libopenshot/builds/25609191
the current value of _WIN32_WINNT is 0x502 so it doesn't causes any errors. Where it was set first time (the log with the mentioned error) I don't know, but it is not `_mingw.h`, because the faulty log uses the `mingw-w64-x86_64-headers-git 7.0.0.5255.83bdce54-1` package and there, the variable surely would be defined as 0x601 - not 0x502, as I have currently (or unknown number that was set in the old log that has the error). Right now the error isn't exist in the mentioned environment, so I will remove this part from the PR.

src/CMakeLists.txt
SuslikV: As for the #262 , when it will be implemented (in case it happens first) and there will be no problems with building in the  environment that was mentioned earlier - I will remove this part from the PR too.

jonoomph: #262 is merged

jonoomph:
Quick question for @SuslikV... with the removal of the _WIN32_WINNT definitions and the merging of #262, is there anything left in this PR which has not been addressed in another PR?

SuslikV:
Not much is left and what is left is not critical. I left it just for beauty of the code. Feel free to close unmerged.

jonoomph:
Sorry to close unmerged :disappointed: But I think we've done some great CMake improvements over the past few weeks & months, and things are feeling really great now! So thanks for your efforts!

SuslikV:
Great is too loud ^_^ , man.

PR #222, opened 2019-04-18
Update Python install path detection

ferdnyc:
This changes the `PYTHON_MODULE_PATH` handling in two ways:

* Makes it a cache variable, so detection is only run once and
  only if it's not already set. This allows it to be overridden
  on the command line, e.g.
  `cmake -DPYTHON_MODULE_PATH:PATH=lib/python3.6/dist-packages`

* Uses the presence of the `pybuild` executable to sense for
  a Debian-derived system (only on UNIX AND NOT APPLE), and
  if found it falls back to the old `getsitepackages()[0]`
  path extraction method.

This works in my tests on Fedora and Ubuntu 18.10, hopefully older/Mintyer will prove the same. `pybuild` is part of dh-python, which synaptic reports as a dependency of `python3-dev` and `python3-all`, so here's hoping...

PR #228, opened 2019-05-03
Fix Python install path detection on Debian

ferdnyc:
Fixes the breakage from #222.


ferdnyc:
Nope, commits are still out of whack.

PR #227, opened 2019-05-02
Revert "Update Python install path detection"

jonoomph:
Reverts OpenShot/libopenshot#222

This broke on both Linux and Mac GitLab builders... so just reverting for now.

Linux Output:
```
-- Removed runtime path from "/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/build/install-x64/../../../../../../../../../usr/local/lib/python3.4/dist-packages/_openshot.so"
-- Installing: /home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/build/install-x64/../../../../../../../../../usr/local/lib/python3.4/dist-packages/openshot.py
$ mv install-x64/lib/python3.4/site-packages/*openshot* install-x64/python
mv: cannot stat ‘install-x64/lib/python3.4/site-packages/*openshot*’: No such file or directory
ERROR: Job failed: exit status 1
```

Mac Output:
```
-- Installing: /Users/jonathan/builds/7d5103a1/0/OpenShot/libopenshot/build/install-x64/../../src/bindings/python/_openshot.so
-- Installing: /Users/jonathan/builds/7d5103a1/0/OpenShot/libopenshot/build/install-x64/../../src/bindings/python/openshot.py
$ mv install-x64/lib/python3.6/site-packages/*openshot* install-x64/python
mv: rename install-x64/lib/python3.6/site-packages/*openshot* to install-x64/python/*openshot*: No such file or directory
ERROR: Job failed: exit status 1
```

Issue #218, opened 2019-04-10
Python Install Path Logic is Invalid


jonoomph:
With previous versions of libopenshot, the install path for our Python bindings were determined with the following code:

```
EXECUTE_PROCESS ( COMMAND ${PYTHON_EXECUTABLE} -c "import site; print(site.getsitepackages()[0])"
OUTPUT_VARIABLE _ABS_PYTHON_MODULE_PATH
OUTPUT_STRIP_TRAILING_WHITESPACE )

Output on my system: /usr/local/lib/python3.5/dist-packages
```
And now we use the following code:

```
execute_process ( COMMAND ${PYTHON_EXECUTABLE} -c "\
from distutils.sysconfig import get_python_lib; \
print( get_python_lib( plat_specific=True, prefix='${CMAKE_INSTALL_PREFIX}' ) )"
OUTPUT_VARIABLE _ABS_PYTHON_MODULE_PATH
OUTPUT_STRIP_TRAILING_WHITESPACE )

Output on my system: lib/python3/dist-packages
```

While the newer version supports custom prefixes, it does not seem to find the correct place on Ubuntu and Linux Mint (and probably other distros as well). Not sure what the best solution is, but I wanted to document this in a bug report, and get some additional eyeballs on it.

jonoomph:
@ferdnyc I would love to hear your thoughts on this. :smile:

ferdnyc:
I'll be able to give more meaningful input in a couple of hours when I'm back home, responding on my phone right now while I'm in the waiting room for an appointment in a few minutes.

Offhand, the GNUInstallDirs suggestion another user made for libopenshot-audio might help here, too, if we applied it. (Which we can, now that we're using CMake 3.)

But I also seem to remember CMake 3 having a slightly _different_ variable for explicitly accessing an **absolute** path, maybe something like `${CMAKE_INSTALL_PATH_FULL}` -- that might work to get the required behavior. I'll check the docs when I get home in an hour or so. (Just finished the appointment, actually.)

ferdnyc:
I was also trying to remember what we do in Fedora, but as I suspected we just explicitly define everything to excess rather than querying for values. Like, you can see the build log for my most recent build here. I'd love to copy and paste the `cmake` command line but my phone locks up every time I try to select text on that page. The relevant point, though, is that it's like 400+ characters and explicitly passes in every path variable CMake recognizes. ☺️

[...]

PR #229, opened 2019-05-03
Don't break Python install path detection on Debian

ferdnyc:
OK, this adds in this fix with the #222 changes, which should protect against #227 

Python: Assume /usr/local prefix on Debian

If we're using the fallback path detection method on Debian, we
shouldn't compute the install path relative to `${CMAKE_INSTALL_PREFIX}`
since it _isn't_. So, we assume it's relative to `/usr/local` instead.

### DIFF
--- a/src/bindings/python/CMakeLists.txt
+++ b/src/bindings/python/CMakeLists.txt
@@ -65,29 +65,39 @@ if (PYTHONLIBS_FOUND AND PYTHONINTERP_FOUND)
 	target_link_libraries(${SWIG_MODULE_pyopenshot_REAL_NAME}
 	                      ${PYTHON_LIBRARIES} openshot)
 
-    ### Check if the following Debian-friendly python module path exists
-	SET(PYTHON_MODULE_PATH "${CMAKE_INSTALL_PREFIX}/lib/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages")
-	if (NOT EXISTS ${PYTHON_MODULE_PATH})
-
-		### Check if another Debian-friendly python module path exists
-		SET(PYTHON_MODULE_PATH "${CMAKE_INSTALL_PREFIX}/lib/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/dist-packages")
-		if (NOT EXISTS ${PYTHON_MODULE_PATH})
-
-			### Calculate the python module path (using distutils)
-			execute_process ( COMMAND ${PYTHON_EXECUTABLE} -c "\
+	### FIND THE PYTHON INTERPRETER (AND THE SITE PACKAGES FOLDER)
+	if (UNIX AND NOT APPLE)
+		### Special-case for Debian's crazy, by checking to see if pybuild
+		### is available. We don't use it, except as a canary in a coal mine
+		find_program(PYBUILD_EXECUTABLE pybuild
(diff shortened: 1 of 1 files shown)


sparfenyuk:
I don't understand why it is so complicated? Can't we just use FIND_PACKAGE(Python 3)?

ferdnyc:
@sparfenyuk 
> I don't understand why it is so complicated? Can't we just use FIND_PACKAGE(Python 3)?

That wouldn't give us the module _install_ path.

We can't use `find_package(Python3)` anyway, because that didn't exist until CMake 3.12, and we need to support all the way down to CMake 3.1. 

(The thing we _are_ using, because it existed **before** CMake 3.12, is `find_package(PythonLibs)` and `find_package(PythonInterp)`. Still has the same problem that it doesn't give us the install path, though, just points us to the interpreter executable and build tools.)

Really, it's only complicated because [the Debian Python people are on crack](https://github.com/OpenShot/libopenshot/issues/218#issuecomment-482893395).

ferdnyc:
(TL;DR version of the linked comment:)

Debian-derived distros follow the [Debian Python Policy](https://debian-python.readthedocs.io/en/latest/debian-policy.html) which says that:

> A special directory is dedicated to public Python modules installed by the local administrator, `/usr/lib/python3/dist-packages` for all Python 3 versions, `/usr/local/lib/python2.Y/dist-packages` for Python 2.6 and later, and `/usr/local/lib/python2.Y/site-packages` for Python 2.5 and earlier.
> 
> For a local installation by the administrator of Python 2.6 and later, a special directory is reserved to Python modules which should only be available to this Python, `/usr/local/lib/python2.Y/site-packages` (and `/usr/local/lib/python3/site-packages` for all Python 3 versions). Unfortunately, for Python 2.5 and earlier this directory is also visible to the system Python.

...But then _their own `distutils`_ returns `/usr/local/lib/python3/dist-packages` as the module instal path for the default `/usr/local/` prefix. You'll notice that directory is not even **mentioned** in the above quote. It's also _not in the `sys.path` of the distro `/usr/bin/python3.x` interpreters_! So if you were to install modules there, they would not be available without path modifications.

jonoomph:
Thanks @ferdnyc! This looks all sorts of crazy (not blaming you, haha). But I'm feeling crazy, so let's give it a try. Hopefully one day someone will stop by and say, you guys are nuts, just use this 1 line solution, lol! :+1:

jonoomph:
Darn, all builders failed initially. Looks like Linux did work okay, with a minor tweak to the gitlab-ci.yml file (just a path changing), but Mac and Windows failed in a stranger way. Might be related to Cmake version, not sure. Details are included in https://github.com/OpenShot/libopenshot/pull/260.

jonoomph:
Maybe we just go with the most "standard / simplest" approach, and let Debian patch things if they want to hard-code it, or use a different method? I can always make things work with respect to the builders... and distros can (and usually do) patch the Python install path.

ferdnyc:
I hate this change. Or it hates me, not even sure which.

All I know is that the current bindings install-path logic is completely broken. The Python logic tries too hard to be too clever, and turns into a mess.  Whereas the the _Ruby_ bindings aren't nearly clever enough — if you build with a relative `CMAKE_INSTALL_PREFIX`, it will try to write into `/usr/` on `make install`! That's just horrible.

I _really_ want to fix that, still, but I feel like there are too many moving parts too far out of my reach to figure out how.

PR #230, opened 2019-05-03
The default return value is present

eisneinechse:
Remove else so that the default return value is used if no other return was used."else if" in line 334 had no else and therefore in some cases no return value was present.

### DIFF
--- a/src/KeyFrame.cpp
+++ b/src/KeyFrame.cpp
@@ -336,9 +336,8 @@ bool Keyframe::IsIncreasing(int index)
 			return false;
 		}
 	}
-	else
-		// return default true (since most curves increase)
-		return true;
+	// return default true (since most curves increase)
+	return true;
 }
 
 // Generate JSON string of this object


jonoomph:
LGTM

jonoomph:
Merged! :+1:

eisneinechse:
There is now a new pull request #231 	
Here:
```
 if (current_value < next_value)
 		return true; // Curve Increasing
 	else
 		return false; // Curve Decreasing
```
is replaced by
```
		if (current_value >= next_value) {
￼				// Decreasing
￼				return false;
￼			}
```
and a default return:
```
	return true;
```
That way the default return value, the normal one, is true and false is the special case. It might even be better to assign the values to a return variable and at the end have a return returnvalue;
In some cases, like critical sections of parallel OpenMP sections it is necessary to have just a single exit point.
I have a bad feeling if there is no return value at the very end of a method. The probability of something being missed at a later point is high.
Anyway, whatever what is going to be used at the end is going to be much easier to read and understand than what it was before.

PR #233, opened 2019-05-08
fix a number of memory leaks

ckirmse:
- some were with libav functions
- same were due to non-virtual destructors

it was previously leaking a full frame (could be several megs based on resolution) plus a bunch of libav structures. Most of the fixes were found using -fsanitize=address on a very simple test program (setting the environment variable ASAN_OPTIONS=fast_unwind_on_malloc=0 when running the program generates much more helpful stack traces too). Tested only with ffmpeg 3.2.

### DIFF
--- a/include/CacheBase.h
+++ b/include/CacheBase.h
@@ -60,6 +60,8 @@ namespace openshot {
 		/// @param max_bytes The maximum bytes to allow in the cache. Once exceeded, the cache will purge the oldest frames.
 		CacheBase(int64_t max_bytes);
 
+		virtual ~CacheBase();
+
 		/// @brief Add a Frame to the cache
 		/// @param frame The openshot::Frame object needing to be cached.
 		virtual void Add(std::shared_ptr<Frame> frame) = 0;
--- a/include/CacheMemory.h
+++ b/include/CacheMemory.h
@@ -71,7 +71,7 @@ namespace openshot {
 		CacheMemory(int64_t max_bytes);
 
 		// Default destructor
-		~CacheMemory();
+		virtual ~CacheMemory();
 
 		/// @brief Add a Frame to the cache
 		/// @param frame The openshot::Frame object needing to be cached.
--- a/include/Clip.h
+++ b/include/Clip.h
@@ -112,7 +112,10 @@ namespace openshot {
 
 		// File Reader object
 		ReaderBase* reader;
-		bool manage_reader;
+
+		/// If we allocated a reader, we store it here to free it later
+		/// (reader member variable itself may have been replaced)
+		ReaderBase* allocated_reader;
 
 		/// Adjust frame number minimum value
 		int64_t adjust_frame_number_minimum(int64_t frame_number);
--- a/include/Clip.h
+++ b/include/Clip.h
@@ -160,7 +160,7 @@ namespace openshot {
 		Clip(ReaderBase* new_reader);
 
 		/// Destructor
-		~Clip();
+		virtual ~Clip();
 
 		/// @brief Add an effect to the clip
 		/// @param effect Add an effect to the clip. An effect can modify the audio or video of an openshot::Frame.
--- a/include/ClipBase.h
+++ b/include/ClipBase.h
@@ -69,6 +69,7 @@ namespace openshot {
 
 		/// Constructor for the base clip
 		ClipBase() { };
+		virtual ~ClipBase();
 
 		// Compare a clip using the Position() property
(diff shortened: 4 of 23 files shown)


jeffski:
Good to see some work on this, I've had issues when rendering on low memory environments, anything that can improve this is awesome. The builds failed for some reason, hope that can be fixed.

ckirmse:
also sorry I did not run the official tests before I sent this PR. I'm running it now locally and it's failing for me too. I'll work on that.

jonoomph:
Thanks for this! I think I'll pull this into an official branch, so our builders (Mac, Windows, and Linux) can all give it a try, and make sure this works across OSes.

ckirmse:
Sounds great! I wouldn't be surprised if we want/need to tweak things to support various versions of ffmpeg/libav just right, so please send suggestions my way and hopefully we can get things working well. I tried to make ownership of various objects as clear as possible to reduce future confusion, but it was a little tough, particularly as someone from the outside who doesn't know the code base. Feedback/suggestions welcome.

PS using the sanitize tool, I also found that FFmpegWriter.cpp write_audio_packets reads past the end of its input buffer too, but it was a little too convoluted for me to fix yet.

jonoomph:
Changing this to merge into a new branch, so I can test on all builders and test things

jonoomph:
I am seeing a consistent crash after exporting (I've reproduced on FFmpeg3 on Windows and Linux so far)

jonoomph:
I'm debugging it now

jonoomph:
Here is the crash (on Linux):
```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
0x0000000003a2e2e0 in ?? ()
(gdb) bt
#0  0x0000000003a2e2e0 in ?? ()
#1  0x00007fffbb030c24 in openshot::Timeline::ClearAllCache (this=0x343fd30)
    at /home/jonathan/apps/libopenshot-git/src/Timeline.cpp:1480
Python Exception <class 'RuntimeError'> Type does not have a target.: 
#2  0x00007fffbba7b994 in _wrap_Timeline_ClearAllCache (args=)
    at /home/jonathan/apps/libopenshot-git/build/src/bindings/python/openshotPYTHON_wrap.cxx:39230
```

jonoomph:
My first reaction is to not delete the final_cache in Timeline.
```
	delete final_cache;
	final_cache = NULL;
```

ckirmse:
Well... final_cache was allocated in Timeline's constructor, so it needs to be deleted somewhere, and the destructor would usually be the right place. However, if Timeline's SetCache is called then that would break  things. I did not see that _actually_ getting used anywhere.

jonoomph:
It is called from openshot-qt in a couple places (such as before an export)

jonoomph:
Actually, the line which crashes is `clip->Reader()->GetCache()` inside the `Timeline::ClearAllCache()` method. We are trying to return the address of final_cache most likely, but this would be of a Clip's reader, and not the Timeline's reader.

jonoomph:
However, on a side note, otherwise things are working well with this PR!

ckirmse:
ok, so that's mostly good :). Yeah the relationship between the ClearAllCache() and Close() functions around Timeline and Readers and FrameMappers was a bit hard for me to understand. I'd be very happy if you could help fix that cleanup if you understand that better, or if you have a specific test/situation that fails that you can point me to then I'm absolutely willing to help.

jonoomph:
In Timeline::Close(), we destroy the FrameMapper instances, and then later try and access them. How about this: https://github.com/OpenShot/libopenshot/commit/bd21d1a751b3d1375cae37747e6d193e2e38d17c

jonoomph:
One more tweak: https://github.com/OpenShot/libopenshot/commit/968e472c73bccac9179a139fa7584a05acf8a95b (which fixes a crash on openshot-qt Exit. Essentially, the timeline's Cache can be managed by itself (by default), or by the user (if they call SetCache). This fixes that logic, and prevents crashes (at least in my testing so far)

ckirmse:
ok, yeah that looks really good to me. That should solve both problems you describe.

Heads up the comment on Timeline.cpp line 924 (before Timeline::SetCache) incorrectly says it's getting the cache.

jonoomph:
@ckirmse I've made a few more fixes (caused by ~Timeline() and ~Clip() destructors... causing random failures on Travis CI). Please take a look at the memory-fixes branch (and see if it all works for you): https://github.com/OpenShot/libopenshot/commits/memory-fixes. Thanks!

PR #236, opened 2019-05-14
Fixing Memory Leaks (lots of them)

jonoomph:
Thanks to @ckirmse (and @eisneinechse) for the help with identifying memory leaks, adding virtual destructors, testing and fixing some crashes. Original PR: https://github.com/OpenShot/libopenshot/pull/233. As far as I can tell, things seem stable and non-crashy with this branch, and all unit tests work now.


ckirmse:
ok just ran your latest code on my project here successfully and without leaks. It looks good!

jonoomph:
Thanks! Merged into `develop`.

Issue #239, opened 2019-05-30
Example.cpp: Memory usage grows out of control if run with longer video


ferdnyc:
@jonoomph — If you have a moment, a quick question about the libopenshot example code.

Interested in getting a better picture of when, where, and how libopenshot allocates memory during the encoding process (see OpenShot/openshot-qt#2811), I modified `src/examples/Example.cpp` with useful local paths, different codecs, a longer output stream (1000 frames instead of 100), and the correct hardware decoder type for my Nvidia GTX 710 GPU (VDPAU decoder), but no other substantial changes. Here's a complete diff from the checked-in source to my version:
```diff
diff --git a/src/examples/Example.cpp b/src/examples/Example.cpp
index 8033968..e4f3305 100644
--- a/src/examples/Example.cpp
+++ b/src/examples/Example.cpp
@@ -37,19 +37,19 @@ using namespace openshot;
 int main(int argc, char* argv[]) {
 
 	Settings *s = Settings::Instance();
-	s->HARDWARE_DECODER = 2; // 1 VA-API, 2 NVDEC
+	s->HARDWARE_DECODER = 6; // 1 VA-API, 2 NVDEC, 6 VDPAU, 7 QSV
 	s->HW_DE_DEVICE_SET = 1;
 
-    FFmpegReader r9("/home/jonathan/Videos/sintel_trailer-720p.mp4");
+    FFmpegReader r9("/home/ferd/rpmbuild/REPOS/openshot-qt/bugtesting/memory-tests/BBB-2m-clip1_00:00:00-00:02:02.mp4");
     r9.Open();
     r9.DisplayInfo();
 
     /* WRITER ---------------- */
-    FFmpegWriter w9("/home/jonathan/metadata.mp4");
+    FFmpegWriter w9("/var/tmp/example-out.mp4");
 
     // Set options
-    w9.SetAudioOptions(true, "libmp3lame", r9.info.sample_rate, r9.info.channels, r9.info.channel_layout, 128000);
-    w9.SetVideoOptions(true, "libx264", r9.info.fps, 1024, 576, Fraction(1,1), false, false, 3000000);
+    w9.SetAudioOptions(true, "aac", r9.info.sample_rate, r9.info.channels, r9.info.channel_layout, 128000);
+    w9.SetVideoOptions(true, "libx264", r9.info.fps, 1920, 1080, Fraction(1,1), false, false, 3000000);
 
     w9.info.metadata["title"] = "testtest";
[...]

PR #240, opened 2019-05-30
free cache in FrameMapper::Close()

ckirmse:
- this hugely reduces the memory used by rendering a timeline
  with a lot of clips
- could be related to issue #239

### DIFF
--- a/src/FrameMapper.cpp
+++ b/src/FrameMapper.cpp
@@ -651,6 +651,16 @@ void FrameMapper::Close()
 		// Close internal reader
 		reader->Close();
 
+		// Clear the fields & frames lists
+		fields.clear();
+		frames.clear();
+
+		// Mark as dirty
+		is_dirty = true;
+
+		// Clear cache
+		final_cache.Clear();
+
 		// Deallocate resample buffer
 		if (avr) {
 			SWR_CLOSE(avr);


jonoomph:
This looks good to me! I'm gonna test it out real quick though.

jonoomph:
LGTM! Worked great.. nice fix!

ferdnyc:
> * could be related to issue #239

I suspect it's not, only because `Example.cpp` only contains a single clip, and (obviously, therefore) doesn't call `Close()` until the end of the encode process.

It's still a great fix, and hopefully will have some effect on OpenShot/openshot-qt#2811, but #239 here can't really be `Close()`-related/dependent.

ferdnyc:
Oh, wait, _FrameMapper_... not the Reader. Whoops. Ignore my previous comment, then. :confused:

ferdnyc:
Unfortunately, though, this change does not reduce the out-of-control memory growth observed in #239. :disappointed:

ferdnyc:
Oh, wait... no, yes it does! At least when the hardware decoder is enabled, yes. Resident memory consumption stays at around 800MB constant! Killer fix, @ckirmse ! :+1: 

I'll do a clean rebuild and double-check whether the software-decoder path is unaffected, as it seemed to be in my previous test.

ferdnyc:
Huh. So, this is odd. With this fix in place, the hardware-decoded path is indeed fixed. But, with the hardware decoder disabled, the `openshot-example` process grows at a constant rate throughout the entire export process. There's still a leak in the software decoder path, looks like.

ckirmse:
Is your test a single clip on a timeline? Because due to the frame caches, generally speaking the memory usage will go up throughout the processing of a clip, up to the max size of the frame caches. I haven't looked at the frame cache size limits personally.

ferdnyc:
> Is your test a single clip on a timeline? Because due to the frame caches, generally speaking the memory usage will go up throughout the processing of a clip, up to the max size of the frame caches. I haven't looked at the frame cache size limits personally.

No timeline at all, `Example.cpp` just uses a single `FFMpegReader` to generate frames that it dumps into an `FFMpegWriter`.

(I've updated my previous comment since posting it.)

ckirmse:
fwiw I'm using the software decoding path and I tested a timeline with this fix that had 137 clips (each about 10s long) and memory did not climb beyond a fixed amount right at the beginning (memory usage of the process never went over 900MB).

ferdnyc:
> fwiw I'm using the software decoding path and I tested a timeline with this fix that had 137 clips (each about 10s long) and memory did not climb beyond a fixed amount right at the beginning (memory usage of the process never went over 900MB).

Wow, that's amazing! It sounds like this will go a long way towards addressing the process memory-growth I documented in OpenShot/openshot-qt#2811 then.

ferdnyc:
Huh! Well, this is very screwy.

I was trying to run `openshot-example` under `memusage` so I could generate graphs of the memory growth like I did over in OpenShot/openshot-qt#2811 , and what I've discovered is that _this_ leak disappears when the process is shimmed by `memusage`. 

I have the **OLD** code (git 9ffd6a6, specifically) checked out, and the exact same `src/openshot-example` binary will grow nonstop if I just run it in the shell. But, `memusage -p /tmp/old.png -t -T -x 1200 -y 800 src/openshot-example` exhibits no memory growth, the graph looks like this:
old

(Which is also how the graph would look when _not_ running under memusage, with this fix applied.)

So, turns out this one's a bit of a heisenbug after all, which (partly) explains my earlier confusion. (The rest of the explanation is that I'm me, so confusion is expected.)

PR #241, opened 2019-06-03
Silence deprecated warnings in ffmpeg 3.x

eisneinechse:
The deprecated calls are and will be present in ffmpeg 3.x and later version use other calls. Therefore, the warning is useless and obscures only important warnings.

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -992,14 +992,7 @@ AVStream *FFmpegWriter::add_audio_stream() {
 		throw InvalidCodec("A valid audio codec could not be found for this file.", path);
 
 	// Create a new audio stream
-#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
-_Pragma ("GCC diagnostic push")
-_Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
-#endif
 	AV_FORMAT_NEW_STREAM(oc, audio_codec, codec, st)
-#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
-_Pragma ("GCC diagnostic pop")
-#endif
 
 	c->codec_id = codec->id;
 #if LIBAVFORMAT_VERSION_MAJOR >= 53
@@ -1082,14 +1075,7 @@ AVStream *FFmpegWriter::add_video_stream() {
 		throw InvalidCodec("A valid video codec could not be found for this file.", path);
 
 	// Create a new video stream
-#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
-_Pragma ("GCC diagnostic push")
-_Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
-#endif
 	AV_FORMAT_NEW_STREAM(oc, video_codec, codec, st)
-#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
-_Pragma ("GCC diagnostic pop")
-#endif
 
 	c->codec_id = codec->id;
 #if LIBAVFORMAT_VERSION_MAJOR >= 53
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -992,7 +992,14 @@ AVStream *FFmpegWriter::add_audio_stream() {
 		throw InvalidCodec("A valid audio codec could not be found for this file.", path);
 
 	// Create a new audio stream
+#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
+_Pragma ("GCC diagnostic push")
+_Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
+#endif
 	AV_FORMAT_NEW_STREAM(oc, audio_codec, codec, st)
+#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
+_Pragma ("GCC diagnostic pop")
+#endif
 
 	c->codec_id = codec->id;
 #if LIBAVFORMAT_VERSION_MAJOR >= 53
@@ -1075,7 +1082,14 @@ AVStream *FFmpegWriter::add_video_stream() {
 		throw InvalidCodec("A valid video codec could not be found for this file.", path);
 
 	// Create a new video stream
+#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
+_Pragma ("GCC diagnostic push")
+_Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
+#endif
 	AV_FORMAT_NEW_STREAM(oc, video_codec, codec, st)
+#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
+_Pragma ("GCC diagnostic pop")
+#endif
 
 	c->codec_id = codec->id;
 #if LIBAVFORMAT_VERSION_MAJOR >= 53
--- a/include/FFmpegUtilities.h
+++ b/include/FFmpegUtilities.h
@@ -209,9 +209,12 @@
 		#define AV_FORMAT_NEW_STREAM(oc, st_codec, av_codec, av_st) 	av_st = avformat_new_stream(oc, NULL);\
 			if (!av_st) \
 				throw OutOfMemory("Could not allocate memory for the video stream.", path); \
-			c = avcodec_alloc_context3(av_codec); \
-			st_codec = c; \
-			av_st->codecpar->codec_id = av_codec->id;
+			_Pragma ("GCC diagnostic push"); \
+			_Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\""); \
+			avcodec_get_context_defaults3(av_st->codec, av_codec); \
+			c = av_st->codec; \
+			_Pragma ("GCC diagnostic pop"); \
+			st_codec = c;
 		#define AV_COPY_PARAMS_FROM_CONTEXT(av_stream, av_codec) avcodec_parameters_from_context(av_stream->codecpar, av_codec);
 	#elif LIBAVFORMAT_VERSION_MAJOR >= 55
 		#define AV_REGISTER_ALL av_register_all();


ferdnyc:
@eisneinechse 

Can you give an example of what deprecation warnings you're getting, and with what FFmpeg version? 

The call you've wrapped is a macro, defined in `include/FFmpegUtilities.h`, and it _shouldn't_ be using deprecated calls. I'm concerned that the warning may be a valid sign that a bug was introduced recently. (Specifically in 833fcb8, which adds a call to `avcodec_get_context_defaults3` — that function is not only deprecated as far back [as FFmpeg 3.1](https://ffmpeg.org/doxygen/3.1/group__lavc__core.html#ga1dd4bf43e6a4ec8b8d76bd3673e5e73a), but the deprecation message is fairly dire):
> This function should not be used, as closing and opening a codec context multiple time is not supported. A new codec context should be allocated for each new use.

eisneinechse:
Example of warning:
/home/peterm/git_remove/libopenshot/src/FFmpegWriter.cpp:36:17: note: #pragma message: You are compiling with experimental hardware encode
 #pragma message "You are compiling with experimental hardware encode"
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/peterm/git_remove/libopenshot/src/../include/ChannelLayouts.h:32:0,
                 from /home/peterm/git_remove/libopenshot/src/../include/Frame.h:59,
                 from /home/peterm/git_remove/libopenshot/src/../include/CacheBase.h:32,
                 from /home/peterm/git_remove/libopenshot/src/../include/CacheMemory.h:34,
                 from /home/peterm/git_remove/libopenshot/src/../include/ReaderBase.h:36,
                 from /home/peterm/git_remove/libopenshot/src/../include/FFmpegWriter.h:39,
                 from /home/peterm/git_remove/libopenshot/src/FFmpegWriter.cpp:31:
/home/peterm/git_remove/libopenshot/src/FFmpegWriter.cpp: In member function ‘AVStream* openshot::FFmpegWriter::add_audio_stream()’:
/home/peterm/git_remove/libopenshot/src/../include/FFmpegUtilities.h:212:41: warning: ‘AVStream::codec’ is deprecated [-Wdeprecated-declarations]
    avcodec_get_context_defaults3(av_st->codec, av_codec); \
                                         ^
/home/peterm/git_remove/libopenshot/src/../include/FFmpegUtilities.h:212:41: note: in definition of macro ‘AV_FORMAT_NEW_STREAM’
    avcodec_get_context_defaults3(av_st->codec, av_codec); \
                                         ^~~~~
...

This calls were already replaced by the calls that are used for ffmpeg 4.x but the new version has a memory leak. The old version doesn't leak memory. The old version is also still valid in ffmpeg 3.x but is marked as deprecated.
The old version should in the end be replaced by the new version used in ffmpeg 4.x when the memory leak is fixed but for now the old version fixes the memory leak and is valid and I don't think that there will be a ffmpeg 3.x version that will have this old version removed.
The warnings are only suppressed in ffmpeg 3.x.
I get this warnings with ffmpeg 3.4

ferdnyc:
@eisneinechse Thanks much! Hah, the deprecation wasn't even where I thought it was. (It's related, because the deprecation of `AvStream::codec` is the _reason_ `avcodec_get_context_defaults3()` is no longer supposed to be used... but point is, there are bigger fish here.

\*shrug\* It's not up to me, of course, but I would be wary of removing those warnings. Every piece of information I can find about the `AVStream::codec` deprecation leads me to believe that it was a response to a pretty serious issue with the previous API (and the way users were abusing it), and having code that uses it in libopenshot feels like something, at the VERY least, we should be reminded of frequently.

I mean, just read about the [MythTV developers' crusade against `AVStream::codec`](https://code.mythtv.org/trac/ticket/13186#comment:5). Peter Bennett put a _lot_ of effort into trying to excise it from their code base, seemingly For Reasons™. (Emphasis his.)

> Patch 005 - deprecated AVStream::codec and avcodec_close
> 
> **This patch has been updated. The original patch crashed on some streams that do not have codecs (e.g. MHEG). It is now changed to use stream->codecpar wherever possible, and to check whether an AVCodecContext exists in cases where it may have failed to be created. There is still a problem that stream->codec is used by some MythTV modifications to FFMPEG code. As of this patch it is still using stream->codec in those cases.**
> 
> In the old method, ffmpeg automatically creates an AVCodecContext for every AVStream. The context can be opened and closed again any number of times. This is deprecated and the application code is now responsible for creating and freeing an AVCodecContext when needed. The AVCodecContext that the application creates may not be opened and closed more than once. The deprecated function avcodec_close must not be called on an AVCodecContext that you create, instead you have to free the AVCodecContext (using avcodec_free_context) and allocate a new one if you want to read the stream again.
> 
> In many cases, reference fields in AVStream::codec are now available in AVStream::codecpar. Where possible the code now uses those fields, to avoid having to create an AVCodecContext.
> 
> "codec" no longer shows in the online documentation for AVStream. It is only defined in ffmpeg if you define FF_API_LAVF_AVCTX when compiling ffmpeg. See external/FFmpeg/libavformat/version.h, which has a list of defines that will be set in future to disable this and others. For more information google FF_API_LAVF_AVCTX. Instead of AVStream::codec, the AVStream::codecpar structure contains the information you need to allocate your own AVStream.
> 
> Old code for AVStream::codec
> 
> AVCodecContext *avctx = stream->codec;
> new code using AVStream::codecpar
> 
> AVCodec *pCodec = avcodec_find_decoder(stream->codecpar->codec_id);
> AVCodecContext *avctx = avcodec_alloc_context3(pCodec);
> avcodec_parameters_to_context(avctx, stream->codecpar);
> av_codec_set_pkt_timebase(avctx, stream->time_base);
> The avctx must saved and the same value used for subsequent calls with that stream. Existing code which uses stream->codec must instead keep track of the axctx value to use each time a packet is received.
> 
> Old Code for avcodec_close:
> 
> avcodec_close(st->codec);
> New code using avcodec_free_context:
> 
> avcodec_free_context(&avctx);
> Note the old code must still be used for any AVCodecContext obtained from stream->codec.

...But like I said, it's not up to me anyway. And _any_ efforts to clean up the tree and the build process are always extremely welcome!

eisneinechse:
@ferdnyc 
Agreed. I think that we should be reminded to fix the new way to do it frequently and use that instead of the old deprecated way..
My concern is that the warnings flood the output and important warnings vanish from site. Maybe an option would be to silence the warning but output a message, like the one that shows that hardware acceleration is compiled or not compiled, for both ffmpeg 4 and ffmpeg 3.
The version that is used for compilations with ffmpeg 4.x could then also have that message which might get this leak fixed faster. As soon as the leak is fixed the version for ffmpeg 4.x could then be used for ffmpeg 3.x too.

ferdnyc:
> @ferdnyc
> Agreed. I think that we should be reminded to fix the new way to do it frequently and use that instead of the old deprecated way..
> My concern is that the warnings flood the output and important warnings vanish from site. Maybe an option would be to silence the warning but output a message, like the one that shows that hardware acceleration is compiled or not compiled, for both ffmpeg 4 and ffmpeg 3.

I could get behind that, sure. As long as it's kept track of, that's my real concern. (TBH it would also be nice to silence the endless flood of deprecation warnings when you use a system-installed jsoncpp, instead of the bundled ancient one.)

> The version that is used for compilations with ffmpeg 4.x could then also have that message which might get this leak fixed faster. As soon as the leak is fixed the version for ffmpeg 4.x could then be used for ffmpeg 3.x too.

That's the part I guess I'm not clear on... is this an FFmpeg leak, or a libopenshot leak? If it's a leak in old FFmpeg, I doubt anyone's going to be fixing it. But if it's a libopenshot issue... it just seems weird to me, that it would affect the >3.2 code, but not the 4.x code. Especially since `AVstream::codec` was already long since deprecated, by 3.4.

There was also that _massive_ memory leak that @ckirmse fixed recently in #240 , which was also affecting the 4.0 code. That patch went in _after_ 833fcb8 — so with that fixed, are we sure there still even is a leak in 3.2+, using `avcodec_alloc_context3()`?

There's also the fact that, in the _same_ patch where the >3.2 code was reverted to use the deprecated `avcodec_get_context_defaults3()` method of accessing the context, the cleanup code was _also_ updated to use `avformat_free_context()` when destroying the writer. These changes both came in that 833fcb8 commit:

```diff
diff --git a/include/FFmpegUtilities.h b/include/FFmpegUtilities.h
index 0d12ba7..7e3e007 100644
--- a/include/FFmpegUtilities.h
+++ b/include/FFmpegUtilities.h
@@ -209,9 +209,9 @@
 		#define AV_FORMAT_NEW_STREAM(oc, st_codec, av_codec, av_st) 	av_st = avformat_new_stream(oc, NULL);\
 			if (!av_st) \
 				throw OutOfMemory("Could not allocate memory for the video stream.", path); \
-			c = avcodec_alloc_context3(av_codec); \
-			st_codec = c; \
-			av_st->codecpar->codec_id = av_codec->id;
+			avcodec_get_context_defaults3(av_st->codec, av_codec); \
+			c = av_st->codec; \
+			st_codec = c;
 		#define AV_COPY_PARAMS_FROM_CONTEXT(av_stream, av_codec) avcodec_parameters_from_context(av_stream->codecpar, av_codec);
 	#elif LIBAVFORMAT_VERSION_MAJOR >= 55
 		#define AV_REGISTER_ALL av_register_all();
diff --git a/src/FFmpegWriter.cpp b/src/FFmpegWriter.cpp
index 072ac6d..ddcd4e1 100644
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -884,9 +884,8 @@ void FFmpegWriter::flush_encoders() {
 }
 
 // Close the video codec
-void FFmpegWriter::close_video(AVFormatContext *oc, AVStream *st) {
-	AV_FREE_CONTEXT(video_codec);
-	video_codec = NULL;
+void FFmpegWriter::close_video(AVFormatContext *oc, AVStream *st)
+{
 #if IS_FFMPEG_3_2
 	//  #if defined(__linux__)
 		if (hw_en_on && hw_en_supported) {
@@ -900,10 +899,8 @@ void FFmpegWriter::close_video(AVFormatContext *oc, AVStream *st) {
 }
 
 // Close the audio codec
-void FFmpegWriter::close_audio(AVFormatContext *oc, AVStream *st) {
-	AV_FREE_CONTEXT(audio_codec);
-	audio_codec = NULL;
-
+void FFmpegWriter::close_audio(AVFormatContext *oc, AVStream *st)
+{
 	// Clear buffers
 	delete[] samples;
 	delete[] audio_outbuf;
@@ -942,12 +939,6 @@ void FFmpegWriter::Close() {
 	if (image_rescalers.size() > 0)
 		RemoveScalers();
 
-	// Free the streams
-	for (int i = 0; i < oc->nb_streams; i++) {
-		av_freep(AV_GET_CODEC_ATTRIBUTES(&oc->streams[i], &oc->streams[i]));
-		av_freep(&oc->streams[i]);
-	}
-
 	if (!(fmt->flags & AVFMT_NOFILE)) {
 		/* close the output file */
 		avio_close(oc->pb);
@@ -957,8 +948,9 @@ void FFmpegWriter::Close() {
 	write_video_count = 0;
 	write_audio_count = 0;
 
-	// Free the context
-	av_freep(&oc);
+	// Free the context which frees the streams too
+	avformat_free_context(oc);
+	oc = NULL;
 
 	// Close writer
 	is_open = false;
```

But AFAIK, `avformat_free_context()` is what it should've been using all along to properly clean up the context, including in the 3.2+ code. So I'm wondering if those issues being fixed now means that it's safe to go back to using `avcodec_alloc_context3()` in the 3.2+ code as well?

eisneinechse:
@ferdnyc 
Unfortunately I don't have a lot of time to spare right now.
I don't like to silence the warning and would very much like to resolve the leak in a proper manner fast. The suppression was only intended temporarily and only because the warning is so verbose.
It would be great of someone could check if the patches from @ckirmse would already solve the memory leak, maybe I will find some time but I think the best person to check that would be ckirmse. The easiest way is probably compile with ffmpeg as the proper way is used there and no change to the source is needed.
These leaks were present way before I started to contribute. They should be addressed soon but at the time I started the main goal was to make openshot stable and the smaller leaks were kind of forgotten.

eisneinechse:
@ferdnyc 
I did some test under ffmpeg 3 (Linux).
The memory leak is about the same with the new (correct) way and the old way. I see no leak with a 720p input but with a 4k 60 fps input video file there is still a leak (a small one). But I did only run openshot normally and watched the memory usage of the system.
From this I would say the new way would work as good as the old deprecated one BUT someone should check it more thoroughly.
AND the problem is not visible, maybe because it is small, when the video is 720p, it is easier visible with 4k input video files.
People with many small 4k video clips may have the problem, people with only a few and 720p video clips won't see a problem.
So, maybe we should after more memory tests use the new correct way again for ffmpeg 3 (and forget about silencing the warnings). I guess the real test is using it and waiting for feedback from the users.

eisneinechse:
These were the changes I used for my tests that did not show any changes in memory leaks:
```
diff --git a/include/FFmpegUtilities.h b/include/FFmpegUtilities.h
index 7e3e007..0d12ba7 100644
--- a/include/FFmpegUtilities.h
+++ b/include/FFmpegUtilities.h
@@ -209,9 +209,9 @@
                #define AV_FORMAT_NEW_STREAM(oc, st_codec, av_codec, av_st)     av_st = avformat_new_stream(oc, NULL);\
                        if (!av_st) \
                                throw OutOfMemory("Could not allocate memory for the video stream.", path); \
-                       avcodec_get_context_defaults3(av_st->codec, av_codec); \
-                       c = av_st->codec; \
-                       st_codec = c;
+                       c = avcodec_alloc_context3(av_codec); \
+                       st_codec = c; \
+                       av_st->codecpar->codec_id = av_codec->id;
                #define AV_COPY_PARAMS_FROM_CONTEXT(av_stream, av_codec) avcodec_parameters_from_context(av_stream->codecpar, av_codec);
        #elif LIBAVFORMAT_VERSION_MAJOR >= 55
                #define AV_REGISTER_ALL av_register_all();

```

eisneinechse:
@ferdnyc 
@jonoomph 
Unfortunately the memory leak is still there. It manifests itself when an export is finished or canceled.It is not there with the old, deprecated way. It can be best observed when exporting to a 4k video file, canceling it and export again (over and over again). Each time some memory isn't freed.

ckirmse:
Like everyone else here I don't have a ton of free time right now but I love libopenshot and I may have caused whatever problem this is so I am motivated. Can someone explain what they're seeing what is wrong and how I can easily reproduce it? If so, I will dedicate some time to testing it and trying to fix it.

eisneinechse:
@ckirmse 
The problem is not a real problem. The warnings produced by the calls a very verbose. All I wanted was to silence or limit the flood of warnings by silencing them and now showing a message that there is a problem with the new version because there is a memory leak.
The call that is used now in ffmpeg 3.x (and before) is deprecated and should be replaced with the version that is used in the ffmpeg 4.x version.
So the plan would be to find a fix to the memory leak using the new call and then using that solution in ffmpeg 3.x too. At some point the older call will no longer work for ffmpeg > 4.x . It should be fine for ffmpeg 3.x and lower.

eisneinechse:
@ckirmse 
The memory leak with the new call shows when an export is finished. Not all reserved memory is freed. The best way to observe it is to export to a 4k video over and over again. Every time the memory shrinks. On 720p video exports the memory loss is so small that nobody notices it but the higher the resolution the bigger the leak. And if someone only export once it will also not be noticed.

ckirmse:
Are you sure the issue is with libopenshot and not openshot-qt? I'm not using openshot-qt. With libopenshot, I've created a timeline with over 100 clips and exported it, with the address sanitizer (for leak tracking) enabled, and it showed no leaks. Is it possible there are gui-related leaks with exports?

eisneinechse:
@ckirmse 
I think there is a smaller memory leak in the GUI.
The memory leak I am referring to is one that is only a single leak that is manifesting itself when the export is stopped, either through a finished export or a canceled export. At the end of the export something is not freed. AND it is only there when the new version is used not with the deprecated version.
So to trigger it it is not necessary to have a lot of input files but to export multiple times. That's why I don't see it, I only export once and after that I close openshot-qt, and I export to 720p not 4k.
The magnitude of the leak is dependent on the output resolution. So I guess it is/are some frame(s) or more likely packet(S) that is/are not freed. Why it shows when using avcodec_alloc_context3 but not when using avcodec_get_context_defaults3 to format a new stream is not clear to me.
The memory leak is only at the end of the export.
I simply used the version of AV_FORMAT_NEW_STREAM that is used for ffmpeg 4.x in ffmpeg 3.x and the memory leak is there. The only difference between a version of libopenshot that has the leak and one that doesn't have the leak is the definition of AV_FORMAT_NEW_STREAM.

ferdnyc:
@eisneinechse That's a lot of testing, thanks! Do you know if FFmpeg 4.0 builds show the same leak as >3.2, when using the "new-style" (`avcodec_alloc_context3()`) form of `AV_FORMAT_NEW_STREAM`? (If you tested that.)

If they **do**, then presumably our code is just failing to do everything it should be doing, when it comes to freeing all of the contexts being allocated... a bug it would be good to find.

If _not_, then it sounds like this is just an FFmpeg leak in their 3.x code, which was fixed in 4.0. In which case, I'm personally fine with using the `avcodec_get_context_default3()` code that doesn't leak as a workaround, and shutting up their deprecation warnings (with a `#pragma` note about it, including that we're using the deprecated code because the new code leaks in >3.2).

eisneinechse:
@ferdnyc I think it leaks in ffmpeg 4 too but those test were a long time ago and I will have to re-do them to be sure.
As I wrote before I am using openshot-qt to export just one video at a time and that's why I didn't get the leak myself. And with 720p the leak is so small that it is not noticeable.
But that leak has to be fixed. I totally agree.
As soon as I have time and remember where my ffmpeg 4 installation was I will test it. But if you have an installation with ffmpeg 4 you can easily test it yourself. Simple export the same (small) video over and over again while looking at the used memory, preferably export to 4k.
The thing that make me think is that something is not freed because one thing is different when **opening** the stream. Maybe that something is allocated at the start that is not freed not at the end of the export.

ferdnyc:
I mean, my biggest concern is that, with the "new-style" `AV_FORMAT_NEW_STREAM` (whether it's in 3.x or 4.0 FFmpeg), _two_ contexts are being allocated. One for the audio stream, one for the video stream. But then `FFmpegWriter::Close()` is only ever closing _one_, which it refers to as "the" context.

It's a bit confusing because there are two types of contexts here: One is the AVFormatContext, which is what's passed in to `AV_FORMAT_NEW_STREAM` as `oc`. The other is the AVCodecContext, which is created by `avcodec_alloc_context3()` (when used). Even [the FFmpeg 4.0 docs](https://ffmpeg.org/doxygen/4.0/group__lavc__core.html#gae80afec6f26df6607eaacf39b561c315) say, regarding AVCodecContexts:
> The resulting struct should be freed with `avcodec_free_context()`.

Which we're never doing.

eisneinechse:
@ferdnyc Maybe that's the problem.

eisneinechse:
@ferdnyc I think that is the problem. The second one is from getting the stream, I think. You have to look what is used for ffmpeg 2. I took that from an example at that time and was wondering myself why one has to alloc another context. But at that time the documentation for ffmpeg 3 was "not perfect", most of the examples had problems.

ferdnyc:
FFmpeg's moving-target API not being documented properly!? **#UNPOSSABLE!** :laughing:

eisneinechse:
@ferdnyc 
the original was 
#define AV_GET_CODEC_CONTEXT(av_stream, av_codec) av_stream->codec
I think we have to improve that, not AV_FORMAT_NEW_STREAM

Great, now I spilled my coffee!

eisneinechse:
At that time I thought "OK it works for now, that will do for now". I shouldn't have done that. ANd then it worked for me so I forgot. Now I remember. I glad that now many people look into it, I don't have too much time right now. And when someone have seen code for the nth time (s)he will not see the bug at the n+1th time.

eisneinechse:
@ferdnyc the same AV_GET_CODEC_CONTEXT is used in FFmpegReader and it does not have a leak there. Strange.

eisneinechse:
One thing: If the context would be the problem why is the memory leaks size dependent on the resolution of the video? It probably has to be something that is a frame or packet.

ferdnyc:
> One thing: If the context would be the problem why is the memory leaks size dependent on the resolution of the video? It probably has to be something that is a frame or packet.

I guess I agree that's _confusing_... but if it **wasn't** the context, then why would switching from using `avcodec_allocate_context3()` to `avcodec_get_context_default3()` make the leak go away? Could there be something allocated in the context that's resolution-dependent?

eisneinechse:
@ferdnyc I completely agree.
And I am totally confused.
The only thing I can think of is that by setting the defaults something later on is not allocated or the allocation is correct only when the defaults are set. Well, I am confused.
It's just the size is kind of big for a context and also the dependence on the resolution is strange. Maybe that's something for valgrind.

eisneinechse:
@ferdnyc 
If avcodec_get_context_defaults3 occurs after avcodec_alloc_context3(codec), it is superfluous since avcodec_alloc_context3 already sets up the defaults. If it occurs in other cases it must change to avcodec_free_context and avcodec_alloc_context3 since reusing an AVCodecContext is no longer supported.
from:
https://code.mythtv.org/trac/ticket/13186

eisneinechse:
This might help to find the problem.
https://code.mythtv.org/trac/attachment/ticket/13186/006_more_avstream_codec.patch

src/FFmpegWriter.cpp
```diff
@@ -992,7 +992,14 @@ AVStream *FFmpegWriter::add_audio_stream() {
 		throw InvalidCodec("A valid audio codec could not be found for this file.", path);
 
 	// Create a new audio stream
+#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
+_Pragma ("GCC diagnostic push")
+_Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
+#endif
 	AV_FORMAT_NEW_STREAM(oc, audio_codec, codec, st)
+#if (IS_FFMPEG_3_2 && (LIBAVFORMAT_VERSION_MAJOR < 58))
+_Pragma ("GCC diagnostic pop")
```
ferdnyc: Completely unrelated to the previous discussion about `AVFormat::Context`, a question: Is there any reason these couldn't just be,
```cxx
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
and
```cxx
#pragma GCC diagnostic pop
```
instead? 

AIUI, `_Pragma()` exists to be used in things like macros, because you can't macro a `#keyword`-style preprocessor directive. But we're not in a macro here, we're not even in a header. This is right in the C++ source, where `#pragma` should be fine, and seems clearer to me.

eisneinechse:
That way the warning is still there in ffmpeg 4 in case someone thinks about going to the old way in ffmpeg 4 too.
I used the _Pragma because it is standardized and I wanted to include it in the definition in FFmpegUtilities.h wher it would have been inside a define but it did not work out.

eisneinechse:
@ferdnyc To be honest the lower limit is useless because in ffmpeg 2 it is not deprecated, it simply doesn't matter there.
And #pragma would also be fine but _Pragma is C++11 standard and #pragma is is quasi standard. I wouldn't mind if #pragma is used. I would much more like if the _Pragma could be included in FFmpegUtilities.h and work. My attempts failed. Maybe someone else has more luck. That way the pragma would only be in the ffmpeg3.2 part where it is used and deprecated.
I don't want to hide the problem and I would have never thought about using it if I wouldn't have found multiple occurrences of silenced warnings in ffmpeg sources.

ferdnyc:
> And #pragma would also be fine but _Pragma is C++11 standard and #pragma is is quasi standard.

Is it really? Huh. My impression is that they were equivalent. _Individual_ pragmas are compiler-dependent, so it's a crap shoot regardless of which form you use, but the C++11 standard [talks about](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf#page=422) `_Pragma()` as though it's just an alternate form of `#pragma` useful in macros, which is how I've been thinking of it. I may have to rethink that!

As far as incorporating it into `FFmpegUtilities.h`, with macro expansion that gets tricky since deprecation warnings are displayed at the point of use, so the pragma needs to go _with_ the macro — it can't modify the macro definition in the header, since then it'll only be processed once during the `#include` and won't still be in effect when the macro is later expanded in the C++ source. So, I think it'd have to be something like this:

```cpp
#define AV_FORMAT_NEW_STREAM(oc, st_codec, av_codec, av_st) 	av_st = avformat_new_stream(oc, NULL); \
	if (!av_st) \
		throw OutOfMemory("Could not allocate memory for the video stream.", path); \
	_Pragma ("GCC diagnostic push"); \
	_Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\""); \
	avcodec_get_context_defaults3(av_st->codec, av_codec); \
	_Pragma ("GCC diagnostic pop"); \
	c = av_st->codec; \
	st_codec = c;
```

ferdnyc:
Oops! Actually the pop has to be lower, because `c = av_st->codec;` also warns.

But, this worked for me, compiling against FFmpeg 4.1.3 with the corresponding macro modified to use `avcodec_get_context_defaults3()`:
```cpp
#define AV_FORMAT_NEW_STREAM(oc, st_codec, av_codec, av_st)     av_st = avformat_new_stream(oc, NULL);\
        if (!av_st) \
                throw OutOfMemory("Could not allocate memory for the video stream.", path); \
        _Pragma ("GCC diagnostic push"); \
        _Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\""); \
        avcodec_get_context_defaults3(av_st->codec, av_codec); \
        c = av_st->codec; \
        _Pragma ("GCC diagnostic pop"); \
        st_codec = c;
```
Without the pragmas, I got deprecating warnings. With them, blissful silence.

eisneinechse:
I could have sworn I did exactly that and it did not work. Probably something in the pragmas that was wrong.
Thanks, I think that would be the way to go.

eisneinechse:
I just now modified FFmpegUtilities.h in 3.2 part but I still get the warnings.
Strange

eisneinechse:
That's the diff:
~~~
diff --git a/include/FFmpegUtilities.h b/include/FFmpegUtilities.h
index 7e3e007..4ac8b15 100644
--- a/include/FFmpegUtilities.h
+++ b/include/FFmpegUtilities.h
@@ -209,8 +209,11 @@
                #define AV_FORMAT_NEW_STREAM(oc, st_codec, av_codec, av_st)     av_st = avformat_new_stream(oc, NULL);\
                        if (!av_st) \
                                throw OutOfMemory("Could not allocate memory for the video stream.", path); \
+                       _Pragma ("GCC diagnostic push"); \
+                       _Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\""); \
                        avcodec_get_context_defaults3(av_st->codec, av_codec); \
                        c = av_st->codec; \
+                       _Pragma ("GCC diagnostic pop"); \
                        st_codec = c;
                #define AV_COPY_PARAMS_FROM_CONTEXT(av_stream, av_codec) avcodec_parameters_from_context(av_stream->codecpar, av_codec);
        #elif LIBAVFORMAT_VERSION_MAJOR >= 55
~~~

eisneinechse:
@ferdnyc  OK, now I tried g++ 9 and it works. I guess g++ 7 has a bug with _Pragma

ferdnyc:
Huh. I don't have a 3.x to test against, unfortunately. But I may be able to scare one up. I'll give it another try, see if I missed something. Will probably take me a couple of hours, though, I have a pair of jsoncpp PRs I need to finish up first.

ferdnyc:
Oh, just saw your later update. (Refresh before replying, refresh before replying!) Well, that blows! :disappointed: Hmmmm.

eisneinechse:
It's not even gcc, it's only g++ before 9.0, I guess.

eisneinechse:
clang doesn't have the bug either from what I read.

ferdnyc:
OOC, does generating the build tree with `cmake -DCMAKE_CXX_FLAGS="-Wno-deprecated"` shut them up? I suppose there's always that route, as overkill and brute-force as it may be...

eisneinechse:
I wouldn't do that. There are some really important warnings in there. Maybe a scaled down version of my patch. Suppress for ffmpeg lower than 4.0 and use the #pragma version as gcc lower than 4.6 (or 4.2?) has problems with _Pragma.

ferdnyc:
(If it does, then it should be able to extend `src/CMakeLists.txt` so that it only does that when building with GCC versions < 9, and/or only if the user requests it with a `cmake -DIGNORE_DEPRECATED:BOOL=1` type dealie.)

eisneinechse:
I would like to keep it simple as the call should be replaced with a fixed version of what is used now with ffmpeg 4. I see this only as temporary.

ferdnyc:
Oh yeah, [here's the bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55578).

How about this? (From that bug report.)
```
cmake -DCMAKE_CXX_FLAGS="-no-integrated-cpp"
```
That's reported to trick g++ into not screwing up the preprocessing of those `_Pragma()` directives by having the preprocessor and compiler run together instead of separately, so they should be handled as expected.

(I have to read up on what _other_ side effects it would have, but for the moment I'm just curious whether it really does work. I tried it on my system w/ gcc 9 / FFMpeg 4.1, and it didn't seem to affect anything, which is good.)

eisneinechse:
OK, I will have to test it.

ferdnyc:
Ah, no, I got that backwards:

> **-no-integrated-cpp**     
> Performs a compilation in two passes: preprocessing and compiling.  This option allows a user supplied "cc1", "cc1plus", or "cc1obj" via the -B option.  The user supplied compilation step can then add in an additional preprocessing step after normal preprocessing but before compiling.  The default is to use the integrated cpp (internal cpp)
> The semantics of this option will change if "cc1", "cc1plus", and "cc1obj" are merged.

So, it should be safe and pretty side-effect free.

ferdnyc:
(It doesn't sound like an alternate preprocessor _has_ to be supplied with `-B`, though, and I guess that's the trick: It keeps using its own preprocessor, but doesn't screw up the pragmas. Or I'm misinterpreting that bug report.)

eisneinechse:
It works on g++ 7.4.0

ferdnyc:
> It works on g++ 7.4.0

giphy

eisneinechse:
I will check with 8.3.0 too, just to be sure.
Maybe you should submit a PR with that. Maybe we can even use that switch only for g++ lower than 9.0?

ferdnyc:
> Maybe we can even use that switch only for g++ lower than 9.0?

That's what I'm thinking, yeah. I just have to check how we can do compiler version lookups in a way that's CMake 3.1 compatible. Some of that stuff didn't exist back in the dark ages.

eisneinechse:
I guess the downside is that the compilation is slower. But that should not be the big problem.
And .... g++ 8.3.0 works too. :-)

jonoomph:
@eisneinechse Does this PR need to be updated before merging? Does it still reflect what was agreed in the conversation above? I see some diffs / patches mentioned in the conversation and just want to verify.

jonoomph:
LGTM! Good job working through this one!

PR #246, opened 2019-06-09
Fix version macros in bindings

ferdnyc:
The `#define`s for the version information in `include/Version.h` have semicolons at the end, exactly the way `#define`s shouldn't:
https://github.com/OpenShot/libopenshot/blob/0d4ea7fe71e88bcee4a7fd1404bd52c8e2169997/include/Version.h#L37-L42

This doesn't _really_ affect the C++, but when those macros get translated into Python by the bindings, it very much does affect things:
```console
$ # Before
$ python3 -c 'import openshot; print(openshot.OPENSHOT_VERSION_ALL)'
0;.2;.3;
```
This PR removes the semicolons, and adjusts the regular expressions in `CMakeLists.txt` accordingly.
```console
$ # After
$ PYTHONPATH=/home/ferd/rpmbuild/REPOS/libopenshot/build/src/bindings/python
$ python3 -c 'import openshot; print(openshot.OPENSHOT_VERSION_ALL)'
0.2.3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@ message(STATUS "Determining Version Number (from Version.h file)")
 #### Get the lines related to libopenshot version from the Version.h header
 file(STRINGS ${CMAKE_CURRENT_SOURCE_DIR}/include/Version.h
      OPENSHOT_VERSION_LINES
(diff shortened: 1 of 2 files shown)


ferdnyc:
Closing in favor of #253

PR #247, opened 2019-06-11
Add 2 new multiplexing presets for mp4, mov files

SuslikV:
Add 2 new multiplexing presets for mp4, mov files:
* mp4_faststart
* mp4_fragmented

The Preset usage from openshot-qt export.py as follows (example):
w.SetOption(openshot.VIDEO_STREAM, "muxing_preset", "mp4_faststart")

YouTube suggest to use streamable file formats to process the uploading
videos faster (by starting its processing when upload not yet complete)
MP4, MOV files export requires additional dictionary keys to be set in
this case.

MP4/MOV is popular, so instead of using external utils, add support for
streamable export directly to OpenShot.

Cons (-):
- The fragmented output requires modern player to playback, otherwise
only first 4-8sec of footage will playback.
- Placing 'moov' box to the begging of the file doing more write operations
to the disk (box moved when file finished).

Pros (+):
+ Fragmented selfcontained mp4 file export allows to
preview the file when encoding is not yet complete, also it requires
slightly less memory for writing, because there is no need to store
whole 'moov' box of mp4 when encoding frames.
+ YouTube starts its processing as soon as first few MB of file was
transferred.

Default setting is none of the muxing options is set.

**Future use and development.**
Later, any custom option may be set by extending _name_ variable
to _name_="muxing_custom", the _value_ to two sub-strings like
_value_="key,value" or similar that can be easily parsed inside SetOption.

Right now no output errors logged because the mentioned dictionary
keys is safe to ignore if set fails. There is no need to throw exception
on this.

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -38,6 +38,9 @@ using namespace openshot;
 #pragma message "You are compiling only with software encode"
 #endif
 
+// Multiplexer parameters temporary storage
+AVDictionary *mux_dict = NULL;
+
 #if IS_FFMPEG_3_2
 int hw_en_on = 1;					// Is set in UI
 int hw_en_supported = 0;	// Is set by FFmpegWriter
@@ -464,6 +467,17 @@ void FFmpegWriter::SetOption(StreamType stream, string name, string value) {
 
 		ZmqLogger::Instance()->AppendDebugMethod("FFmpegWriter::SetOption (" + (string)name + ")", "stream == VIDEO_STREAM", stream == VIDEO_STREAM, "", -1, "", -1, "", -1, "", -1, "", -1);
 
+	// Muxing dictionary is not part of the codec context.
+	// Just reusing SetOption function to set popular multiplexing presets.
+	} else if (name == "muxing_preset") {
+		if (value == "mp4_faststart") {
+			// 'moov' box to the beginning; only for MOV, MP4
+			av_dict_set(&mux_dict, "movflags", "faststart", 0);
+		} else if (value == "mp4_fragmented") {
+			// write selfcontained fragmented file, minimum length of the fragment 8 sec; only for MOV, MP4
+			av_dict_set(&mux_dict, "movflags", "frag_keyframe", 0);
+			av_dict_set(&mux_dict, "min_frag_duration", "8000000", 0);
+			}
 	} else {
 		throw InvalidOptions("The option is not valid for this codec.", path);
 	}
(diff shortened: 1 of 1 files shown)


jeffski:
Nice to see faststart added.

src/FFmpegWriter.cpp
```diff
@@ -511,17 +525,29 @@ void FFmpegWriter::WriteHeader() {
...
+	if (is_mp4 || is_mov)
+		av_dict_copy(&dict, mux_dict, 0);
+
+	if (avformat_write_header(oc, &dict) != 0) {
 		throw InvalidFile("Could not write header to file.", path);
 	};
 
+	// Free multiplexing dictionaries sets
+	if (dict) av_dict_free(&dict);
+	if (mux_dict) av_dict_free(&mux_dict);
```
ferdnyc: I'm also not sure about freeing `mux_dict` here. Should that be moved to `FFmpegWriter::Close`, maybe? If you're just going to free `mux_dict` right away here, then it really _is_ temporary storage and there's no reason not to pass it in to `avformat_write_header()` and let it get overwritten.

I get where the semantics get weird, because you want to pass in an `AVDictionary **` to `avformat_write_header()`, but you only want it to contain `mux_dict` IFF the export format is the right type. So, how about doing this?
```cpp
AVDictionary **dictp = NULL;
if (mux_dict && (is_mp4 || is_mov))
  dictp = &mux_dict;
if (avformat_write_header(oc, dictp) != 0) { ... }
if (mux_dict) av_dict_free(&mux_dict);
```
That way, you'll still pass a `NULL` into `avformat_write_header()` if there's no dict or it's the wrong format, same as before. If everything _is_ right to set the muxing options, then you'll pass in a pointer to the `mux_dict`, but without having to needlessly copy it first.

ferdnyc: (And in that case, you can disregard my previous comments about not labeling it temporary storage or adding a comment to explain the copying.)

src/FFmpegWriter.cpp
```diff
@@ -511,17 +525,29 @@ void FFmpegWriter::WriteHeader() {
...
+	// Set multiplexing parameters
+	AVDictionary *dict = NULL;
+
+	bool is_mp4 = strcmp(oc->oformat->name, "mp4");
+	bool is_mov = strcmp(oc->oformat->name, "mov");
+	// Set dictionary preset only for MP4 and MOV files
+	if (is_mp4 || is_mov)
+		av_dict_copy(&dict, mux_dict, 0);
+
+	if (avformat_write_header(oc, &dict) != 0) {
```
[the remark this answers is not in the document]
SuslikV: I didn't analyze the return values in the code, because it is not critical for performance or what ever. If the options cannot be used, I don't mind. The original thought was not to mix my own muxing dictionary with parameters that in future may be written somewhere else.

src/FFmpegWriter.cpp
SuslikV: The dictionary only needed to pass the options into the function, nothing else. If there was other way to say to FFmpeg that I need faststart or fragmented writing with my MP4 export then code would be completely different. I don't know about other solutions. Maybe someone with better knowledge of FFmpeg and OpenShot can do it better.

SuslikV:
Why presets instead of allow to user to set required keys from the start? Because users usually don't know what exactly should be done to make their lovely mp4 files "streamable". Nothing else was on my mind. If preset options not optimal then they should be changed.

src/FFmpegWriter.cpp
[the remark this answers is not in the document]
ferdnyc: Oh, no, I totally understand about not checking the output dictionary, we don't care about "options that were not found". And I'm not at all concerned about performance, or even memory — I'm sure the AVDictionary is small, and even with the copy they're both getting freed right away. I'm sure it's not an issue.

> The original thought was not to mix my own muxing dictionary with parameters that in future may be written somewhere else

OK, so there may be a need to populate `dict` with other keys besides the ones in `mux_dict`. That makes sense.

My only thought was whether there might be any situation where `FFmpegWriter::WriteHeader()` ends up getting called _again_ later on, after a restart or something... in which case, would there be any use in keeping the `mux_dict` around so those options don't have to be re-set? But I'm not sure that can actually happen, anyway, and even if it does we're probably better off making them call `FFmpegWriter::SetOption() again to re-select the presets... if they still want them applied. Works for me!

jonoomph:
This LGTM as far as I can tell. I like having these new options available. I'm a little nervous about freeing the dict where it currently is, but if nothing is breaking here, no reason to rock the boat I suppose.

PR #249, opened 2019-06-11
make use of crop_x, crop_y, crop_with, crop_height keyframes

chad3814:
@jonoomph let me know if these were implemented somewhere else, but I couldn't find it.

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -713,6 +713,11 @@ string Clip::PropertiesJSON(int64_t requested_frame) {
 	root["has_audio"] = add_property_json("Enable Audio", has_audio.GetValue(requested_frame), "int", "", &has_audio, -1, 1.0, false, requested_frame);
 	root["has_video"] = add_property_json("Enable Video", has_video.GetValue(requested_frame), "int", "", &has_video, -1, 1.0, false, requested_frame);
 
+	root["crop_x"] = add_property_json("Crop X", crop_x.GetValue(requested_frame), "float", "", &crop_x, 0.0, 0.0, false, requested_frame);
+	root["crop_y"] = add_property_json("Crop Y", crop_y.GetValue(requested_frame), "float", "", &crop_y, 0.0, 0.0, false, requested_frame);
+	root["crop_width"] = add_property_json("Crop Width", crop_width.GetValue(requested_frame), "float", "", &crop_width, 0.0, 1.0, false, requested_frame);
+	root["crop_height"] = add_property_json("Crop Height", crop_height.GetValue(requested_frame), "float", "", &crop_height, 0.0, 1.0, false, requested_frame);
+
 	root["wave_color"] = add_property_json("Wave Color", 0.0, "color", "", &wave_color.red, 0, 255, false, requested_frame);
 	root["wave_color"]["red"] = add_property_json("Red", wave_color.red.GetValue(requested_frame), "float", "", &wave_color.red, 0, 255, false, requested_frame);
 	root["wave_color"]["blue"] = add_property_json("Blue", wave_color.blue.GetValue(requested_frame), "float", "", &wave_color.blue, 0, 255, false, requested_frame);
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -713,8 +713,8 @@ string Clip::PropertiesJSON(int64_t requested_frame) {
 	root["has_audio"] = add_property_json("Enable Audio", has_audio.GetValue(requested_frame), "int", "", &has_audio, -1, 1.0, false, requested_frame);
 	root["has_video"] = add_property_json("Enable Video", has_video.GetValue(requested_frame), "int", "", &has_video, -1, 1.0, false, requested_frame);
 
-	root["crop_x"] = add_property_json("Crop X", crop_x.GetValue(requested_frame), "float", "", &crop_x, 0.0, 0.0, false, requested_frame);
-	root["crop_y"] = add_property_json("Crop Y", crop_y.GetValue(requested_frame), "float", "", &crop_y, 0.0, 0.0, false, requested_frame);
+	root["crop_x"] = add_property_json("Crop X", crop_x.GetValue(requested_frame), "float", "", &crop_x, -1.0, 1.0, false, requested_frame);
+	root["crop_y"] = add_property_json("Crop Y", crop_y.GetValue(requested_frame), "float", "", &crop_y, -1.0, 1.0, false, requested_frame);
 	root["crop_width"] = add_property_json("Crop Width", crop_width.GetValue(requested_frame), "float", "", &crop_width, 0.0, 1.0, false, requested_frame);
 	root["crop_height"] = add_property_json("Crop Height", crop_height.GetValue(requested_frame), "float", "", &crop_height, 0.0, 1.0, false, requested_frame);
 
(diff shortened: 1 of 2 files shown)


chad3814:
created a bug where videos are getting cropped wrong

chad3814:
okay this works for me

ferdnyc:
I notice you're changing the default `crop_gravity`... will this affect OpenShot in any way?

Hm. I guess not, as OpenShot only ever makes use of crop as an _effect_, not as a clip property. (Which I guess is why this was never implemented.)

chad3814:
@ferdnyc yeah, it was never implemented, so seems like the defaults should be changeable, also this puts the crop gravity inline with the layout gravity, which seems logical to me.

jonoomph:
Thanks @chad3814! This looks great. Yeah, this went unimplemented until now. I'm hoping that this makes 0 performance difference... I'm hopefull.

jonoomph:
Oh, can you add these 2 keyframes (crop_x, crop_y) to the PropertiesJSON method? This will return them to openshot-qt, and they should appear in the UI, and allow you to play with them. I think I removed them from this method since they were not being used.

jonoomph:
Something like this perhaps...
```
	root["crop_x"] = add_property_json("Crop X", crop_x.GetValue(requested_frame), "float", "", &crop_x, 0.0, 1.0, false, requested_frame);
	root["crop_y"] = add_property_json("Crop Y", crop_y.GetValue(requested_frame), "float", "", &crop_y, 0.0, 1.0, false, requested_frame);
```

chad3814:
thanks for the flurry of PR mergers @jonoomph!

src/Clip.cpp
```diff
@@ -713,6 +713,11 @@ string Clip::PropertiesJSON(int64_t requested_frame) {
 	root["has_audio"] = add_property_json("Enable Audio", has_audio.GetValue(requested_frame), "int", "", &has_audio, -1, 1.0, false, requested_frame);
 	root["has_video"] = add_property_json("Enable Video", has_video.GetValue(requested_frame), "int", "", &has_video, -1, 1.0, false, requested_frame);
 
+	root["crop_x"] = add_property_json("Crop X", crop_x.GetValue(requested_frame), "float", "", &crop_x, 0.0, 0.0, false, requested_frame);
+	root["crop_y"] = add_property_json("Crop Y", crop_y.GetValue(requested_frame), "float", "", &crop_y, 0.0, 0.0, false, requested_frame);
```
ferdnyc: Typo, I think. The 6th and 7th parameters to `add_property_json()` set the range for the parameter (min and max , respectively), so they can't both be the same value (`0.0`) or the parameter is unmodifiable.

Typically it'd be either `0.0`, `1.0` or `-1.0`, `1.0` depending how the value is used. @jonoomph suggested `0.0`, `1.0`, so I assume that's correct and the origin is the upper left corner. If the origin point were the center, `-1.0`, `1.0` would be more common. (Although it's worth discussing whether, for OpenShot, the origin _should_ be the center, as that's probably what people would be expecting?)

src/Clip.cpp
```diff
@@ -76,9 +76,9 @@ void Clip::init_settings()
 	wave_color = Color((unsigned char)0, (unsigned char)123, (unsigned char)255, (unsigned char)255);
 
 	// Init crop settings
-	crop_gravity = GRAVITY_CENTER;
-	crop_width = Keyframe(-1.0);
-	crop_height = Keyframe(-1.0);
+	crop_gravity = GRAVITY_TOP_LEFT;
+	crop_width = Keyframe(1.0);
+	crop_height = Keyframe(1.0);
```
ferdnyc: If these parameters are going to be exported to OpenShot, as I said I think `GRAVITY_CENTER` is how most people would assume cropping works (especially since they won't have the ability to modify `crop_gravity`), with the range for `crop_x` and `crop_y` being `-1.0` to `1.0`.

If that's _not_ going to be the default, then I think `crop_gravity` needs to be exported as well, so it can be set from the GUI. 

If everyone looking to remove black bars from a video (most common use case for cropping) will have to set `crop_x` and `crop_y` to move the origin point, instead of just reducing either `crop_width` or `crop_height` to reduce the video size in the appropriate dimension, that doesn't seem especially user-friendly. (Assuming I've understood properly how these parameters work?

Like, say you've got a `1920`x`1080` video file, and the contents are a `1920` x `800` film matted from 2.4:1 to 16:9 — _super_ common case. Trimming those black bars **should** be as simple as setting `crop_height` to `0.74` (`800 / 1080`) and done. Everything else can be left alone. That's how I think anyone would _expect_ it to work, no? 

If the origin is the upper-left corner, then you have to both set `crop_height` to `0.74`, _and_ set `crop_y` to `.13` (`((1080-800) / 2)/1080`) to move the origin point down to the upper-left corner of the video content.

ferdnyc:
This will be great to have in OpenShot. Having crop properties on the Clip will let us close at least a half-dozen bugs over there that all go the same way:

> #### A wild BUG REPORT has been filed!
> > "The Crop effect doesn't work right."
> 
> "The Crop effect works fine, it just doesn't do what you think it does. It crops output, not input." 
> 
> > "Well, how do I crop the input video?"
> 
> "OK, here's where it gets complicated..."

ferdnyc:
Or... wait, I realize now that maybe I _am_ misunderstanding `crop_gravity`. So I'll delete my previous comment, and just ask this way, using concrete examples:

* What _would_ the `crop_x`, `crop_y`, `crop_width`, and `crop_height` parameters be, to crop a `1920` x `1080` video to `1920` x 800`, removing `240` pixels from both the top and bottom?

*  What would the parameters be if you wanted to crop off the top `120` pixels of a `1280`x`720` video, so the output frame is `1280`x`600`?

chad3814:
Thanks @ferdnyc 

> * What _would_ the `crop_x`, `crop_y`, `crop_width`, and `crop_height` parameters be, to crop a `1920` x `1080` video to `1920` x `800`, removing `140` pixels each from the top and bottom?

with `crop_gravity` set to `GRAVITY_TOP_LEFT`, then `crop_x` would `0`, `crop_y` would be `140 / 1080`, `crop_width` would be `1` and `crop_height` would be `800 / 1080`

> * What would the parameters be if you wanted to crop off the top `120` pixels of a `1280`x`720` video, so the output frame is `1280`x`600`?

you could set `crop_gravity` to `GRAVITY_TOP_LEFT`, and then `crop_x` would be `0`, `crop_y` would be `120 / 720`, `crop_width` would be `1` and `crop_height` would be `600 / 720`. My mind works best in `GRAVITY_TOP_LEFT`, but all four are percentages.

ferdnyc:
> Thanks @ferdnyc
> 
> > * What _would_ the `crop_x`, `crop_y`, `crop_width`, and `crop_height` parameters be, to crop a `1920` x `1080` video to `1920` x `800`, removing `140` pixels each from the top and bottom?
> 
> with `crop_gravity` set to `GRAVITY_TOP_LEFT`, then `crop_x` would `0`, `crop_y` would be `140 / 1080`, `crop_width` would be `1` and `crop_height` would be `800 / 1080`

Hmmm. It'd really be cool if _symmetric_ cropping — in either dimension — didn't require you to adjust `crop_x` or `crop_y` at all, since I think that'll be like 90% of users' cropping needs.

jonoomph:
LGTM! I look forward to playing with this soon!

PR #250, opened 2019-06-12
Fix ImageMagick 7 support (#60)

steils:
Should be backward compatible with ImageMagick 6.

PR #252, opened 2019-06-13
Add backwards-compatible Imagemagick 7 support

ferdnyc:
This PR is an alternative to #250 by @steils — honestly, it was never my intention to step on anyone's toes with this, and I wouldn't have continued with it if it wasn't already in progress by the time @steils submitted #250. But, it was, and I discovered a couple of things along the way that I thought made it worth continuing with. But I apologize to @steils for duplicating their work.

Just as in #250, this adds ImageMagick 7 compatibility to libopenshot without sacrificing ImageMagick 6 support. In the case of this PR, a new header `imclude/MagickUtilities.h` is created to hold the compatibility `#define`s.

Out of an overabundance of caution, I also wrapped the three headers and the three source files that are ImageMagick-only in `#ifdef USE_IMAGEMAGICK` statements. Even though the various places they're used are also wrapped, and therefore they shouldn't ever be accessed anyway unless ImageMagick support is enabled.

The image-conversion code in `src/Frame.cpp` received the only major changes, which warrant explanation.

When I was reading through the ImageMagick APIs looking for ways to make that code version-agnostic, I discovered `ExportImagePixels()` in the `MagickCore` API, which contains basically _all_ of the same code. That function existed in IM6 and continues to exist in IM7. The API of that function is _**unchanged**_ from IM6 to IM7. So instead of doing the export by hand (and having to account for changes in the underlying API), it uses `MagickCore::ExportImagePixels()` which accounts for all of the API changes for us. 

The big revelation was when I discovered `image->constImage()`, which lets me convert the `Magick::Image` we're working with into a `const MagickCore::Image` that I can pass into `MagickCore::ExportImagePixels()`. 

All in all this means that I was able to replace all of this:
```cpp
    // Iterate through the pixel packets, and load our own buffer
        // Each color needs to be scaled to 8 bit (using the ImageMagick built-in ScaleQuantumToChar function)
        int numcopied = 0;
    Magick::PixelPacket *pixels = new_image->getPixels(0,0, new_image->columns(), new_image->rows());
    for (int n = 0, i = 0; n < new_image->columns() * new_image->rows(); n += 1, i += 4) {
        buffer[i+0] = MagickCore::ScaleQuantumToChar((Magick::Quantum) pixels[n].red);
        buffer[i+1] = MagickCore::ScaleQuantumToChar((Magick::Quantum) pixels[n].green);
        buffer[i+2] = MagickCore::ScaleQuantumToChar((Magick::Quantum) pixels[n].blue);
        buffer[i+3] = 255 - MagickCore::ScaleQuantumToChar((Magick::Quantum) pixels[n].opacity);
        numcopied+=4;
    }
```
with just:
```cpp
        MagickCore::ExceptionInfo exception;
        // TODO: Actually do something, if we get an exception here
        MagickCore::ExportImagePixels(new_image->constImage(), 0, 0, new_image->columns(),
                new_image->rows(), "RGBA", Magick::CharPixel, buffer, &exception);
```

(As you can see, there's a TODO in the code pointing out that `MagickCore::ExportImagePixels()` will populate an `exception` struct if it encounters any problems. Currently the code ignores that possibility, which it should not, but I wasn't clear enough on how exception-handling is actually done in the Frame code to feel safe doing anything with it.)

The new code compiles cleanly under IM6. It compiles cleanly under IM7. The tests in `tests/ImageWriter.cpp`, which DO test this code path via `ImageWriter::WriteFrame()`, run successfully when built against either version of ImageMagick.

Unrelatedly, but because I was in there and caught it, the include guard in `include/effects/Mask.h` was using the identifier `OPENSHOT_WIPE_EFFECT_H`. Which only wasn't a problem because we don't _have_ a wipe effect. Still, I changed it to `OPENSHOT_MASK_EFFECT_H` for safety.

### DIFF
--- a/include/Frame.h
+++ b/include/Frame.h
@@ -53,14 +53,14 @@
 #include <memory>
 #include <unistd.h>
 #include "ZmqLogger.h"
-#ifdef USE_IMAGEMAGICK
-	#include "Magick++.h"
-#endif
 #include "ChannelLayouts.h"
 #include "AudioBufferSource.h"
 #include "AudioResampler.h"
 #include "Fraction.h"
 #include "JuceHeader.h"
+#ifdef USE_IMAGEMAGICK
+	#include "MagickUtilities.h"
+#endif
 
 #pragma SWIG nowarn=362
 using namespace std;
--- a/include/ImageReader.h
+++ b/include/ImageReader.h
@@ -28,6 +28,9 @@
 #ifndef OPENSHOT_IMAGE_READER_H
 #define OPENSHOT_IMAGE_READER_H
 
+// Require ImageMagick support
+#ifdef USE_IMAGEMAGICK
+
 #include "ReaderBase.h"
 
 #include <cmath>
@@ -36,9 +39,9 @@
 #include <omp.h>
 #include <stdio.h>
 #include <memory>
-#include "Magick++.h"
 #include "CacheMemory.h"
 #include "Exceptions.h"
+#include "MagickUtilities.h"
 
 using namespace std;
 
@@ -113,4 +116,5 @@ namespace openshot
 
 }
 
-#endif
+#endif //USE_IMAGEMAGICK
+#endif //OPENSHOT_IMAGE_READER_H
--- a/include/ImageWriter.h
+++ b/include/ImageWriter.h
@@ -32,9 +32,10 @@
  * along with OpenShot Library. If not, see <http://www.gnu.org/licenses/>.
  */
 
-
 #ifndef OPENSHOT_IMAGE_WRITER_H
 #define OPENSHOT_IMAGE_WRITER_H
+
+#ifdef USE_IMAGEMAGICK
 
 #include "ReaderBase.h"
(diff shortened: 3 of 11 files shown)


ferdnyc:
(Note: Because the git history got a little wonky when I had to force-push a fix, if this gets merged I'd suggest making use of the option to squash the PR into a single commit.)

steils:
> But I apologize to @steils for duplicating their work.

Not at all, Frank. Actually I think that healthy rivalry is good for all, and will improve the quality overall.

>  MagickCore::ExportImagePixels(new_image->constImage(), 0, 0, new_image->columns(),
                new_image->rows(), "RGBA", Magick::CharPixel, buffer, &exception);

Nice catch! Honestly I don't like these `MAGICK_*` macros, the "ifdef" approach looks easier to read for me. But I'll take this `MagickCore::ExportImagePixels` to #250. Thanks.

jonoomph:
This looks great @ferdnyc! Also, thanks @steils for your work as well.

PR #234, opened 2019-05-10
Fix path to Settings.h

jeffski:
When I build libopenshot and then `sudo make install` it, when I build my own software that uses the library I get an error:

```
/usr/local/include/libopenshot/effects/../OpenMPUtilities.h:35:10: fatal error: ../include/Settings.h: No such file or directory
```

It works with the new path in the PR.

PR #253, opened 2019-06-16
Move version info ownership to CMake

ferdnyc:
This PR pulls the trigger on something I've wanted to take care of for a _while_ now: Making the project's root `CMakeLists.txt` the canonical, authoritative source of the version number data. Expect an equivalent PR for `libopenshot-audio` as well.

All of the messy, backwards parsing of `include/Version.h` that was being done in `CMakeLists.txt` in order to set up the version strings for the build goes away completely. Instead, after the version information is set (directly), using simple CMake commands and variables, the final version header is generated from its template file:
```cmake
configure_file(include/OpenShotVersion.h.in include/OpenShotVersion.h @ONLY)
```

This does necessitate a bit of header reworking, since the version header will now live in the _build_ directory instead of the _source_ directory. That means that all relative-path includes of `../Version.h` or `../../../include/Version.h` or whatever are gone, replaced with just `#include "OpenShotVersion.h"` (The name-change is to avoid possible collisions with same-named header files.)

The build now _relies_ on the include path being set properly for all targets, to include the `${CMAKE_BINARY_DIR}/include/` path where `OpenShotVersion.h` will live.

Honestly, I'd like to fix _all_ of the header files along those same lines, but for now I'm contenting myself with this one. But as I mentioned in #234, it would really be good for the build if we got rid of all the relative-path `#include` statements, and made sure we manage our include directories properly. I have some pretty solid plans for the best way to do that, but it'll mean another huge PR that touches every single file in the tree, so I'm holding off for now.

#### Other notes
* Obviously, this means that any version number changes would now be made at the top of `CMakeLists.txt` at the root of the repository, and only there. The version header (template) in `include/OpenShotVersion.h.in` should normally not require editing.
* This PR also eliminates `include/Tests.h`, which hasn't been necessary for some time, The proper `TEST_MEDIA_PATH` will be set from `tests/CMakeLists.txt` using a preprocessor definition which it adds to `CXX_FLAGS`.
* This PR invalidates #246, which I will close unmerged.


chad3814:
This is great @ferdnyc! I'd really love to see a PR with all the includes fixed, but know that's a huge change.

SuslikV:
It is not a huge at all. The huge change is when you modify rendering core or replace web Timeline with something else.

ferdnyc:
@SuslikV There are different kinds of huge. You're right that it's nothing in terms of lines-of-code or whatever, but a PR to fix how the includes are handled in the build would have to modify _every_ source file, which is indeed a huge task to coordinate. (And means **every** other PR has to be updated and possibly conflict-resolved after it gets merged.)

Your two examples, OTOH, are actually pretty self-contained — the rendering core in `src/windows/{export,video_widget,preview_thread}.py`, and the web timeline in `src/windows/views/timeline_webview.py` and `src/timeline`. Yes, replacing them _wholesale_ would of course be a huge task and a major undertaking, but ideally the replacements could be developed alongside the existing implementations, then switched over to without too much disruption to the rest of the codebase. (...Hey, I said "ideally".)

Of course, if you really wanted to break the world, just decide to replace PyQt5 with wxWidgets, or something like that. Then you basically have to start over from scratch. (Though, wxPython version 4 is coming along, and includes a wx.html2 WebView module, so Python and the web timeline could actually be retained.)

ferdnyc:
(Wait, I forgot this is libopenshot — you mentioned the web timeline and threw me off, since there is no web timeline in libopenshot. So, replacing that would be easy!) :laughing:

SuslikV:
No, it wouldn't. _libopenshot_ is empty vessel without _Timeline_. I wish to have Grail instead. ^_^ So, this PR just small step ahead. It doesn't change logic of the application, it doesn't change any functions. It changes only some internal stuff. Thus, the priority could be lowered (down to the ground floor) and it is unknown will it be even merged. The lib itself has number of unresolved bugs. Nice to see this PR - but it isn't game changer

<details>

(like one of the PRs that resolved huge memory consumption - RAM manufactures were not in mood with this solution...). Right now I see no way to enhance OpenShot without the core modifications. I hardly believe that the core will be rewritten (and _Timeline_ of openshot-qt replaced). This is dead end in the development. IMHO. Ouch!.. I'm not a programmer by the way.

</details>

When I say _libopenshot_ - I only mean OpenShot as whole project, of course. You got me.

I see 19 PRs right now - this isn't huge number at all. The maintainer of the project can ask each author to update their PRs before merge, in case he has no time to do it by himself.

ferdnyc:
Of _course_ it isn't a game-changer. Why would anyone put in the effort required to work on "game-changer" PRs, when "it is unknown will it be even merged." — a statement that applies to all PRs submitted here, not just this one — and even one-line PRs have to sit weeks before getting looked at?

Little, incremental housekeeping changes like this are already a struggle, and require an abundance of patience that, speaking for myself, at times can be hard to find. But I genuinely feel that they make the build more robust, which makes the code easier to work on. So I keep flinging them up here.

SuslikV:
A lot of effort or not - it is interesting question for what reason you are making PRs to this project?
For me, it is more like: "Hey, look, I can make it a bit better. Do you like it?"

jonoomph:
@SuslikV Ouch indeed! @ferdnyc is awesome and every PR is highly appreciated. I typically work in cycles, merging PRs more frequently is a priority, and I'm working to improve in that area. I also spend a good amount of time supporting customers, users, and trying to earn enough money to fund this project out of my own pocket. I use OpenShot all the time. I love it to death. It's useful for tons of different types of projects. Could it be better? Yes. OpenShot's GitHub PRs should be a place of improvement and support though, and not this.

jonoomph:
@ferdnyc Can you take a look at the conflicts on this one?  I've changed the target of this PR to use a new branch... so I can run this through all the builders and verify nothing explodes. :smile: But looks like some of the other PR merges have conflicted with this one.

SuslikV:
@jonoomph , Like for one-person maintainer, the PRs merged quite fast. No jokes. But number of Windows users may surprise you (a lot of work ahead, sure). Application is simple to use for the beginners, IMHO. This is good sign.

jonoomph:
I'm gonna work on this PR today... try and resolve these conflicts

jonoomph:
@ferdnyc I am moving this to another PR (sorry for the confusion): https://github.com/OpenShot/libopenshot/pull/263. Merging this has become a little complicated on my side, so I've manually merged them into another branch.

PR #254, opened 2019-06-20
Add -no-integrated-cpp for G++ < 9

ferdnyc:
This switches on the compiler option `-no-integrated-cpp` for GCC versions < 9, to work around a bug in the handling of `_Pragma()` directives in macros. This is necessary to make the "preferred" form of the PR #241 fixes (to silence deprecation warnings on some versions of FFmpeg) work with those GCC versions.

See [GCC bug 55578](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55578) and the discussion at #241 starting from [this comment](https://github.com/OpenShot/libopenshot/pull/241#issuecomment-503786187) for more details.

I tested this on my system, `-no-integrated-cpp` was inserted into the compiler command line with GCC 7.3.0, but was not with GCC 9.1.1. It should never be inserted with any compiler that doesn't have the short identifier "GNU", so it's safe around other build environments.

Facilitates #241

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,13 @@ Generating build files for OpenShot
   SO/API/ABI Version: ${SO_VERSION}
 ")
 
+#### Work around a GCC < 9 bug with handling of _Pragma() in macros
+#### See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55578
+if ((${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU") AND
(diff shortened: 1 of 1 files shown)


ferdnyc:
Cool. The Travis builds show that `-no-integrated-cpp` works (or, at the very least, doesn't _hurt_ anything) in GCC 5.4.0 as well, @eisneinechse .

jonoomph:
I've been out of town and missed all this fun!

jonoomph:
LGTM! I'll let this move through all our build servers and make sure nothing explodes

jonoomph:
Passed all builders!

PR #262, opened 2019-06-27
Properly install DLL on Win32

ferdnyc:
Adding a RUNTIME DESTINATION to the install for the library target will install the DLL on Windows systems, so it no longer has to be manually copied in the `.gitlab-ci.yml` script.

(I feel better about OpenShot/libopenshot-audio#52 now that I notice the libopenshot install had the same bug, and was just manually working around it.)


jonoomph:
Thanks @ferdnyc, this looks great!

PR #266, opened 2019-07-03
Default values for ZMQ logging calls

ferdnyc:
This is part of an effort I'm broadly calling "Don't make programmers write awkward code". It doesn't actually have anything to do with `-Wall`, it just addresses something that would keep catching my eye as I was visiting bits of code that _were_ flagged by `-Wall`.

### Don't make programmers write awkward code
If you poke around inside libopenshot's code for any length of time, you can't help but notice all those calls to `ZmqLogger::Instance()->AddDebugMethod()` that look like this:

```cpp
ZmqLogger::Instance()->AppendDebugMethod("Timeline::apply_effects (Does effect intersect)",
 "effect->Position()", effect->Position(), "does_effect_intersect", does_effect_intersect,
 "timeline_frame_number", timeline_frame_number, "layer", layer,"", -1, "", -1); 
```

The method takes 13 total arguments (an opening message, then six `(name, value)` tuples), but sometimes you just don't have that much data to log. So, the other 2, or 4, or **twelve** get filled in with dummy values, and the code for the class knows to disregard those:
https://github.com/OpenShot/libopenshot/blob/1b19ae7fe4a6ced31e7eaebf885194cc4a0a23d3/src/FFmpegWriter.cpp#L557

#### Let the compiler be the compiler
Except... this is C++11, where parameters can  have default values, so why should the _caller_ have to fill in all those dummy empty strings and `-1` values? That's a job for the compiler!

So, I added default values to every argument except the first, using the exact same dummy data the caller would be filling in anyway. Now, the caller has the _option_ to stop feeding arguments when they no longer have any to provide, and the compiler will take over and fill in the dumb defaults.

Then, I ran a recursive regexp across the source tree, repeatedly replacing `, "", -1);` with `);` on lines containing `AppendDebugMethod` until I ran out of matches. End result? That line above becomes:
```cpp
 ZmqLogger::Instance()->AppendDebugMethod("FFmpegWriter::WriteHeader"); 
```
the first one at the top becomes:
```cpp
 ZmqLogger::Instance()->AppendDebugMethod("Timeline::apply_effects (Does effect intersect)",
 "effect->Position()", effect->Position(), "does_effect_intersect", does_effect_intersect,
 "timeline_frame_number", timeline_frame_number, "layer", layer); 
```
And so on.

#### Tastes great, but exactly as filling as before

This _feels_ a little dangerous (though for the life of me I can't understand why, even though **I** feel it too!), but it's really not. The more you think about it, the more obvious it becomes that the code after this change is functionally identical _in **every** possible way_ to how it was before. What GCC produces should be practically indistinguishable from before. The source is just a lot more readable, and more comfortable to write.

There are probably even more clever tricks that could be performed with templates and [parameter packs](https://en.cppreference.com/w/cpp/language/parameter_pack), to do type-checking and enforce the pairing of `(name, value)` tuples so that a caller can't do something like this:
```cpp
 ZmqL::I()->AppendDebugMethod("My So-Called Logfile: TNG", "the value is..."); /// ...YES???
```
But, there was nothing stopping them doing that _before_, so the question of enforcing argument pairing is completely orthogonal to the question of providing default values, really.

### DIFF
--- a/include/ZmqLogger.h
+++ b/include/ZmqLogger.h
@@ -45,8 +45,6 @@
 #include "JuceHeader.h"
 
 
-using namespace std;
-
 namespace openshot {
 
 	/**
@@ -58,11 +56,11 @@ namespace openshot {
 	class ZmqLogger {
 	private:
 		CriticalSection loggerCriticalSection;
-		string connection;
+		std::string connection;
 
 		// Logfile related vars
-		string file_path;
-		ofstream log_file;
+		std::string file_path;
+		std::ofstream log_file;
 		bool enabled;
 
 		/// ZMQ Context
@@ -108,19 +106,19 @@ namespace openshot {
 		void Close();
 
 		/// Set or change connection info for logger (i.e. tcp://*:5556)
-		void Connection(string new_connection);
+		void Connection(std::string new_connection);
 
 		/// Enable/Disable logging
 		void Enable(bool is_enabled) { enabled = is_enabled;};
 
 		/// Set or change the file path (optional)
-		void Path(string new_path);
+		void Path(std::string new_path);
 
 		/// Log message to all subscribers of this logger (if any)
-		void Log(string message);
+		void Log(std::string message);
 
 		/// Log message to a file (if path set)
-		void LogToFile(string message);
+		void LogToFile(std::string message);
 	};
 
 }
--- a/include/ZmqLogger.h
+++ b/include/ZmqLogger.h
@@ -96,12 +96,13 @@ namespace openshot {
 		static ZmqLogger * Instance();
 
 		/// Append debug information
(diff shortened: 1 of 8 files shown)


ferdnyc:
> So, I added default values to every argument except the first, using the exact same dummy data the caller would be filling in anyway.

Actually not _quite_ true, since I specified the dummy `float` arguments as `-1.0` instead of just `-1`, like in the calls. Those won't be looked at regardless, if the corresponding strings are empty, and it felt safer not to rely on implicit type conversion for the defaults in the declaration.

jonoomph:
Love it! This was always a bit of a hack, and could probably use some more work in the future, but this is much cleaner than before. Thx!

PR #268, opened 2019-07-07
Install documentation into DOCDIR/API, if built

ferdnyc:
This adds the same documentation install logic used in libopenshot-audio (since OpenShot/libopenshot-audio#56, plus the modifications in OpenShot/libopenshot-audio#59), so that running `make install` AFTER `make doc` will pick up the generated Doxygen output an install it into `${CMAKE_INSTALL_DOCDIR}/API`, which is `${CMAKE_INSTALL_PREFIX}/share/doc/libopenshot/API` on most systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,3 +99,12 @@ add_subdirectory(tests)
 ################### DOCUMENTATION ###################
 # Find Doxygen (used for documentation)
 include(cmake/Modules/UseDoxygen.cmake)
+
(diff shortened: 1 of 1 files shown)


ferdnyc:
(BTW: This shouldn't require any modifications to the existing documentation-publishing process for the openshot.org website, since it pulls the docs from the _build_ dir, without installing them. This PR only affects the install destination, builddir paths are unchanged.)

jonoomph:
LGTM!

PR #269, opened 2019-07-09
Fix resvg rendering issues for common transitions

jonoomph:
When using Resvg to render SVG files to pixels, we were accidentally rendering transparency into our original aspect ratio, and thus, common/SVG-based transitions appeared to be letter boxed. This is essentially a regression caused by our implementation of Resvg.

There are times during preview (at smaller resolutions) when transitions appear to be 1 pixel offset, caused by some aliasing. I'm not sure if this is specifically an resvg problem, or something larger related to source aspect ratios / Qt scaling / etc... Some discussion has taken place on this issue here: https://github.com/RazrFalcon/resvg/issues/153.

### DIFF
--- a/src/QtImageReader.cpp
+++ b/src/QtImageReader.cpp
@@ -229,10 +229,15 @@ std::shared_ptr<Frame> QtImageReader::GetFrame(int64_t requested_frame)
 			path.find(".svgz") != std::string::npos) {
 			ResvgRenderer renderer(QString::fromStdString(path));
 			if (renderer.isValid()) {
+				// Scale SVG size to keep aspect ratio, and fill the max_size as best as possible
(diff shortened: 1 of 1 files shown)


ferdnyc:
@jonoomph Have you seen the issue about chroma inversion with the AppImage builds that include ReSVG? I was able to reproduce it with a downloaded AppImage, I _haven't_ been able to reproduce it yet with ReSVG itself, which makes me think that either it's a bug on the libopenshot side, or that simply updating ReSVG on the AppImage builders might fix it. Last I checked it was old enough that ReSVG has seen a lot of development since. (Unless that's changed, last I checked it was not too long after the 2.4.4 release.

Anyway, the chroma inversion issue is OpenShot/openshot-qt#2697. It affected _only_ the Linux AppImage builds, in my testing, not the Windows builds bundled with ReSVG.

PR #273, opened 2019-07-11
Clean up allocated memory in JSON code

### LINKED ISSUE #271: CharReader* reader need delete
Upgrade jsoncpp to 1.8.4 , all readers need delete
	
        Json::CharReader* reader(rbuilder.newCharReader());
	string errors;
	bool success = reader->parse( value.c_str(), 
                 value.c_str() + value.size(), &root, &errors );
         delete reader;   // this line

ferdnyc:
It was pointed out that I was being sloppy, in the updated JSON code, and not freeing any of the `Json::CharReader` objects that get allocated all over the place now. This created very, _very_ small memory leaks (less than 2 Kb per allocation) that nevertheless could add up over time.

Running `valgrind` on just the `tests/openshot-test` executable previously (meaning, since the JsonCPP upgrade, but before this PR) would flag 61 instances of a `Json` function leaking memory. After this change, those are **all** eliminated, and leaked memory as reported by valgrind is reduced by nearly 7 Kb. (Hey, I said they were **SMALL** leaks.)

Fixes #271 reported by @laochen.

### DIFF
--- a/src/CacheDisk.cpp
+++ b/src/CacheDisk.cpp
@@ -495,6 +495,8 @@ Json::Value CacheDisk::JsonValue() {
 	string errors;
 	bool success = reader->parse( json_ranges.c_str(),
 	                 json_ranges.c_str() + json_ranges.size(), &ranges, &errors );
+	delete reader;
+
 	if (success)
 		root["ranges"] = ranges;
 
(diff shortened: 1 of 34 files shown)


ferdnyc:
I'm going to merge this one.
* Travis passed it
* I built it locally with `-DCMAKE_CXX_FLAGS="-Wall"` and nothing was flagged there either; any use-after-free type issues would surely have shown up
* it effectively has a second pair of eyes in the form of @laochen, who reported the issue it addresses

All in all that should make it very safe to merge.

jonoomph:
@ferdnyc LGTM! Nice

PR #276, opened 2019-07-12
SWIG: Warnings cleanup

ferdnyc:
This PR contains a bunch of changes to the SWIG generation, to make it possible to run with `CMAKE_CXX_FLAGS="-Wall -Wextra"` (and not be bombarded by errors from the generated code).

* Remove a SWIG pragma from `Frame.h` (gcc warns on it)
* Place the equivalent `%warnfilter` in the `openshot.i` files
* Populate `openshot.i` `PROPERTY` `GENERATED_COMPILE_OPTIONS` with flags to disable warning spew in generated SWIG code, if `-Wall` is used

(Also, remove 'using namespace std' from `Frame.h`, and add `std::` prefixes to necessary types.)

### DIFF
--- a/include/Frame.h
+++ b/include/Frame.h
@@ -65,9 +65,6 @@
 	#include "MagickUtilities.h"
 #endif
 
-#pragma SWIG nowarn=362
-using namespace std;
-
 namespace openshot
 {
 	/**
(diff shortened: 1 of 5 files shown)


jonoomph:
LGTM!

PR #277, opened 2019-07-12
FindOpenShotAudio version reporting/validation, other fixes

ferdnyc:
This PR is an overhaul of `FindOpenShotAudio.cmake` with things I needed for my local testing, but it should be more useful overall (fingers crossed that it works everywhere it's supposed to).

The major changes are:

1. **Version checking!** `find_package(OpenShotAudio 0.1.8 REQUIRED)` will now check that the version is at least `0.1.8`, and abort (because of the `REQUIRED`) if not.
    The version is parsed from the `JuceHeader.h` file, and it would be _best_ with a merge of my PR OpenShot/libopenshot-audio#60, to generate that header during build. But, since we've had "0.1.8" hardcoded in there for a long time, it's fine even right now.

    Other features:
    - Supports `EXACT` (but only for M.N.B, e.g. `0.1.8-dev1` matches `0.1.8` EXACT)
    - Outputs the following: (see below for examples)
      - Version requested, if set
      - Version found, if successfully parsed from header
      - Compatibility status, if version set on both ends

2. Accept `LIBOPENSHOT_AUDIO_DIR` cache variable, in addition to the environment variable. (This allows `cmake -DLIBOPENSHOT_AUDIO_DIR=...` to set the path, instead of having to export an environment variable.) The environment variable will still be given preference, if set.

3. Do the same two-stage search for **both** headers and libs.
   (Avoids this situation...)
   - `LIBOPENSHOT_AUDIO_INCLUDE_DIRS=/usr/include/libopenshot-audio`
   - `LIBOPENSHOT_AUDIO_LIBRARY=/home/ferd/build/`

(A lot of this logic was largely inspired by CMake's `FindPythonLibs.cmake`)

Examples of the version-checking in action:
```cmake
# 1. With "0.1.8-dev1" set in header file
find_package(OpenShotAudio REQUIRED) # Success

# 2. With "0.1.8-dev2" set in header file
find_package(OpenShotAudio 0.1.8 REQUIRED) # Success

# 3. With no parseable string in header file
find_package(OpenShotAudio 0.1.8 REQUIRED) # Success

# F1. With "0.1.8-dev3" set in header file
find_package(OpenShotAudio 0.1.7 EXACT REQUIRED) # FAIL

# F2. With "0.1.8-dev1" set in header file
find_package(OpenShotAudio 0.1.9 REQUIRED) # FAIL
```

#### Successful checks
1. `find_package(OpenShotAudio REQUIRED)` with "0.1.8-dev1" set in header file
```
-- Found OpenShotAudio: /usr/lib/libopenshot-audio.so (found version "0.1.8-dev1") 
```
2. `find_package(OpenShotAudio 0.1.8 REQUIRED)` with "0.1.8-dev2" set in header file
```
-- Found OpenShotAudio: /usr/lib/libopenshot-audio.so (found suitable version "0.1.8-dev2",
 minimum required is "0.1.8") 
```
3. `find_package(OpenShotAudio 0.1.8 REQUIRED)` with no parseable string in header file
```
-- Found OpenShotAudio: /usr/lib/libopenshot-audio.so (Required is at least version "0.1.8") 
```
#### Failures
1. `find_package(OpenShotAudio 0.1.7 EXACT REQUIRED)` with "0.1.8-dev3" set in header file
```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenShotAudio: Found unsuitable version "0.1.8-dev3", but
  required is exact version "0.1.7" (found /usr/lib/libopenshot-audio.so)
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:376 (_FPHSA_FAILURE_MESSAGE)
  cmake/Modules/FindOpenShotAudio.cmake:112 (find_package_handle_standard_args)
  src/CMakeLists.txt:115 (FIND_PACKAGE)


-- Configuring incomplete, errors occurred!
```

2. `find_package(OpenShotAudio 0.1.9 REQUIRED)` with "0.1.8-dev1" set in header file
```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenShotAudio: Found unsuitable version "0.1.8-dev1", but
  required is at least "0.1.9" (found /usr/lib/libopenshot-audio.so)
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:376 (_FPHSA_FAILURE_MESSAGE)
  cmake/Modules/FindOpenShotAudio.cmake:112 (find_package_handle_standard_args)
  src/CMakeLists.txt:115 (FIND_PACKAGE)


-- Configuring incomplete, errors occurred!
```

### DIFF
--- a/cmake/Modules/FindOpenShotAudio.cmake
+++ b/cmake/Modules/FindOpenShotAudio.cmake
@@ -5,8 +5,6 @@
 #  LIBOPENSHOT_AUDIO_INCLUDE_DIRS - The juce.h include directories
 #  LIBOPENSHOT_AUDIO_LIBRARIES - The libraries needed to use juce
 
-include(GNUInstallDirs)
-
 if("$ENV{LIBOPENSHOT_AUDIO_DIR}" AND NOT "${OpenShotAudio_FIND_QUIETLY}")
     message(STATUS "Looking for OpenShotAudio in: $ENV{LIBOPENSHOT_AUDIO_DIR}")
 endif()
--- a/cmake/Modules/FindOpenShotAudio.cmake
+++ b/cmake/Modules/FindOpenShotAudio.cmake
@@ -5,34 +5,114 @@
 #  LIBOPENSHOT_AUDIO_INCLUDE_DIRS - The juce.h include directories
 #  LIBOPENSHOT_AUDIO_LIBRARIES - The libraries needed to use juce
 
-message("$ENV{LIBOPENSHOT_AUDIO_DIR}")
+include(GNUInstallDirs)
 
-# Find the libopenshot-audio header files
-find_path(LIBOPENSHOT_AUDIO_INCLUDE_DIR JuceHeader.h
-			PATHS $ENV{LIBOPENSHOT_AUDIO_DIR}/include/libopenshot-audio/
-			/usr/include/libopenshot-audio/
-			/usr/local/include/libopenshot-audio/ )
+if("$ENV{LIBOPENSHOT_AUDIO_DIR}" AND NOT "${OpenShotAudio_FIND_QUIETLY}")
+    message(STATUS "Looking for OpenShotAudio in: $ENV{LIBOPENSHOT_AUDIO_DIR}")
+endif()
 
-# Find the libopenshot-audio.so (check env var first)
-find_library(LIBOPENSHOT_AUDIO_LIBRARY
-		NAMES libopenshot-audio openshot-audio
-		PATHS $ENV{LIBOPENSHOT_AUDIO_DIR}/lib/ NO_DEFAULT_PATH)
+# Find the libopenshot-audio header files (check env/cache vars first)
+find_path(
+  LIBOPENSHOT_AUDIO_INCLUDE_DIR
+  JuceHeader.h
+  HINTS
+    ENV LIBOPENSHOT_AUDIO_DIR
+  PATHS
+    ${LIBOPENSHOT_AUDIO_DIR}
+  PATH_SUFFIXES
+    include/libopenshot-audio
+    libopenshot-audio
+    include
+  NO_DEFAULT_PATH
+)
 
-# Find the libopenshot-audio.so / libopenshot-audio.dll library (fallback)
-find_library(LIBOPENSHOT_AUDIO_LIBRARY
-			NAMES libopenshot-audio openshot-audio
-			HINTS $ENV{LIBOPENSHOT_AUDIO_DIR}/lib/
-			/usr/lib/
-			/usr/lib/libopenshot-audio/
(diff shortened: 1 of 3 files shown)


jonoomph:
@ferdnyc This is **super** cool! You are a true CMake ninja! :1st_place_medal:

ferdnyc:
@jonoomph 

I'm learning my way around it. Wait 'til you see the group of changes I'm working on next. CMake could be doing **SO** much more for us than it is right now, we just have to get out of its way a little (like with the project versioning). But I think we're in good shape now to really start doing that.

Having to support CMake versions all the way down to 3.1 is a challenge at times, but it's probably good and keeps me honest. (Or, occasionally, keeps me sticking ugly things like `HEX_VERSION_OVERRIDE` into our build files,  just to appease older CMake.)

I **do** keep having the experience of visiting the CMake site, reading about some cool feature in the online documentation — which by default shows the docs for the latest release, currently 3.15 — then switching to the CMake 3.1 version and having the cool thing I just read about vanish right before my eyes. _"Noooo!"_

ferdnyc:
In fact, this is a good time to mention: A bunch of the cool next-set-of-changes things I'm working on _may_ require bumping our minimum version slightly, to CMake 3.2. Turns out they made a **lot** of improvements to the way build parameters like compile options and compiler features are propagated between targets, from CMake 3.1 to 3.2. 

But that should be OK, because Xenial shipped with CMake 3.2.2, so even our oldest supported configuration doesn't really _need_ CMake 3.1 compatibility. 

Xenial actually got updates through [CMake 3.5.1](https://launchpad.net/ubuntu/xenial/+source/cmake), so if we make a CMake upgrade part of the Xenial pre-build steps we could possibly raise our minimum version even higher. Which is something I _may_ end up lobbying for, at some point, because IIRC CMake 3.5 was where they really went all-in on the build target system for dependencies (in the `FindWhateverLib.cmake` scripts), and targets are the key to unlocking **all** of CMake's real power. 

But I'll check exactly what a 3.5 minimum version would buy us and decide whether I think it's really necessary, first. 

Worst-case scenario is that our build files will have to check whether each `find_package()` call is exporting targets for the results, and if it isn't (older version) we just create them ourselves using the variables it sets. 

We'll definitely have to do that _anyway_, for some dependencies, because even though CMake 3.5 was (I think) when they decided that `find_package()` should export its results in the form of build targets instead of cache variables, a lot of the bundled `FindWhateverLib.cmake` scripts didn't get upgraded to actually create targets until _much_ later.

PR #278, opened 2019-07-16
Updating MSYS2, New Cmake builder syntax

jonoomph:
Moving 64-bit and 32-bit Windows builders to AWS, and updating all dependencies and build tools. MinGW Makefiles is no longer working, and it appears things have changed to MSYS Makefiles now. FYI: This will speed up our Windows builds quite a bit! (libopenshot)


SuslikV:
Ouch! You have updated the wiki build instructions, I have used old ones. So strange to see `mingw32-make` together with the `MSYS Makefiles`.

ferdnyc:
@SuslikV Well, at the time you used them, the old ones were the new ones. The wiki updates just happened earlier today. (But as I said in #275, I'll make sure those get synced into `INSTALL-WINDOWS.md` in the source tree, too.)

PR #281, opened 2019-07-16
Fixing python3.6 to 3.7 Windows reference

jonoomph:
This is related to our updated Windows builder I'm working on. Trying to get all the GitLab CI working again, with a few new file-paths.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -78,7 +78,7 @@ windows-builder-x64:
     - cd build
     - cmake -D"CMAKE_INSTALL_PREFIX:PATH=$CI_PROJECT_DIR\build\install-x64" -G "MSYS Makefiles" -DCMAKE_MAKE_PROGRAM=mingw32-make -D"CMAKE_BUILD_TYPE:STRING=Release" ../
     - mingw32-make install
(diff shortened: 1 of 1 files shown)


ferdnyc:
@jonoomph 

P.S> If #277 gets merged, along with other goodies it'll no longer be required to set `LIBOPENSHOT_AUDIO_DIR` as an environment variable, `-DLIBOPENSHOT_AUDIO_DIR=/path/to/it` on the CMake command line will be just as effective.

jonoomph:
That would be great! find_path is just frustrating on Windows, and just doesn't seem to work unless I use environment variables to stuff in there. I would love to get rid of that.

ferdnyc:
As far as I _know_ the new `FindOpenShotAudio.cmake` should work equally well on Windows, we'll have to test it with the new builder environment & cmake versions in use. I tested it with my by-now-outdated MSYS2 builds.

 I'm pretty confident at this point that I can **make** it work, anyway, even if it needs a bit more tweaking. 

Also, it'll still use the envvar if it's set, the cmake variable is just an additional option, so _worst_ case is that Windows still has to use the environment variable form, while the rest can dispense with it.

PR #282, opened 2019-07-22
Sync INSTALL-WINDOWS.md with wiki

ferdnyc:
The changes I had to make to the MarkDown install docs to get Doxygen's markdown parser to process them correctly were, due to an oversight on my part, never copied into the wiki versions of those docs.

@jonoomph recently spent considerable time updating the wiki copy of the [Windows Build Instructions](https://github.com/OpenShot/libopenshot/wiki/Windows-Build-Instructions), so as a result the two documents diverged even further.

I've taken the Wiki doc, applied the formatting fixes from the repo's `doc/` directory version, done some additional reformatting to hopefully find a baseline that both parsers can deal with, added some additional tweaks, and made both documents identical. This check-in is the `INSTALL-WINDOWS.md` end of that process, I'll update the Wiki page directly.

I'll also work on syncing the other two platforms' docs, ideally _before_ either is edited further, this time.

### DIFF
--- a/doc/INSTALL-WINDOWS.md
+++ b/doc/INSTALL-WINDOWS.md
@@ -106,140 +106,166 @@ git clone https://github.com/OpenShot/libopenshot-audio.git
 
 The source code is divided up into the following folders.
 
-### build/
+*   `build/`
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
Actually, witrhdrawing this for now, as there are still changes that need to be reconciled.

PR #283, opened 2019-07-23
Adding logging support for resvg

jonoomph:
Adding logging support for resvg (which will output on stderr). I am currently troubleshooting an issue with resvg and Windows 10 + missing fonts. The resvg logging needs to be enabled to better troubleshoot issues with it.

### DIFF
--- a/src/ZmqLogger.cpp
+++ b/src/ZmqLogger.cpp
@@ -30,6 +30,10 @@
 
 #include "../include/ZmqLogger.h"
 
+#if USE_RESVG == 1
+	#include "ResvgQt.h"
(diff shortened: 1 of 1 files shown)


src/ZmqLogger.cpp
```diff
@@ -30,6 +30,10 @@
 
 #include "../include/ZmqLogger.h"
 
+#if USE_RESVG == 1
+	#include "ResvgQt.h"
+#endif
```
ferdnyc: At the risk of being pedantic, shouldn't this be in `ZmqLogger.h`?

PR #285, opened 2019-07-26
Label "Enable Audio/Video" property values as Auto / Off / On

ferdnyc:
Currently "Enable Audio" and "Enable Video" default to `-1`, and can be modified to either `0` or `1`.

Those values are slightly cryptic for the average user, so this PR uses `add_property_choice_json()` to display  those values using the descriptive labels "Auto", "Off", and "On" (respectively).

image

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -716,6 +716,14 @@ string Clip::PropertiesJSON(int64_t requested_frame) {
 	root["has_audio"] = add_property_json("Enable Audio", has_audio.GetValue(requested_frame), "int", "", &has_audio, -1, 1.0, false, requested_frame);
 	root["has_video"] = add_property_json("Enable Video", has_video.GetValue(requested_frame), "int", "", &has_video, -1, 1.0, false, requested_frame);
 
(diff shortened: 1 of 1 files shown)


SuslikV:
Nice. Hope, someday the selected option in the right-click menu becomes checkable (exclusive selection).

ferdnyc:
@SuslikV Mmm, that wouldn't be a bad way to denote existing keyframe positions (in addition to the green row highlight) — and then on other frames where the value is interpolated, that could be indicated by none of the three options selected. (Even though, for a tri-state property like either of these, the value at every frame is **always** going to match one of the three choices.)

Would you mind filing a new issue for that feature request, so we don't lose track of it if/when this gets merged?

ferdnyc:
...I just realized I had to figure out which repo that would even need to go into. :laughing:  I **think** it could be done just as an openshot-qt feature. I hope so, anyway, because to implement it in libopenshot I'm pretty sure we'd have to extend the JSON dictionary for the properties, in order to pass the options' "checked"/"unchecked" status.

jonoomph:
LGTM!

PR #288, opened 2019-07-30
Fix crash during seeking at the start of the file

SuslikV:
Attempt to fix the crash that was detected during investigation of the https://github.com/OpenShot/openshot-qt/issues/2828

This change doesn't fixes the desync issue mentioned in the report. It aims only at the crash resolve.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -921,6 +921,11 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
 						continue;
 					}
 
+					// Packet may become NULL on Close inside Seek if CheckSeek returns false
+					if (!packet)
+						// Jump to the next iteration of this loop
+						continue;
+
 					// Get the AVFrame from the current packet
 					frame_finished = GetAVFrame();
 
@@ -956,6 +961,11 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
 						// Jump to the next iteration of this loop
 						continue;
 					}
+
+					// Packet may become NULL on Close inside Seek if CheckSeek returns false
+					if (!packet)
+						// Jump to the next iteration of this loop
+						continue;
 
 					// Update PTS / Frame Offset (if any)
 					UpdatePTSOffset(false);


ferdnyc:
Ugh, that code is a _horror show_. `CheckSeek()` will _do_ a `Seek()`, which in turn might just decide to close and reopen the file? 

I'm worried this still isn't enough to protect against all the possible side-effects of something that drastic, but it's certainly saf**er**. Good catch!

SuslikV:
I designed this as band aid. I'm not quite sure that this is right solution. I have explanation for this:

It is hard to tell what was main idea when seeking algorithm was developed. I only see result and I know that it has issues. For example, when I exclude some optimizations from the current implementation of the seeking algorithm - it fails to seek the file mentioned in the linked report, but works well for all other files (it may stuck for a while, but seeking is possible).

ferdnyc:
Sorry, realized I never followed up on this.

It may be a band-aid, but those are useful when you're bleeding. If `packet` can be `nullptr`, then we shouldn't be doing anything with it, so this is safe(<b>r</b>) and makes sense to me. I'm going to merge it.

CC: @eisneinechse , @jonoomph

eisneinechse:
@ferdnyc  If not already known: The seek() function in ffmpeg gets slower when frames later in a stream are to be found. This is especially easily visible when watching a *.flv files and jumping to a late frame, like 2 h in the clip. This is in every program I used not just Openshot.
Because video player don't have to jump to a random frame like Openshot has one can watch such a file without problem as long as no jump to a random frame is made. Openshot jumps to frame for playback and exporting. Therefore, when exporting a *.flv file the export speed gets slower and slower with later frames.
I resolve this by running the flv file through ffmpeg (the program) and convert it to mp4 where the seek() function is much faster (this can also easily be observed when using a video player and jumping to random frames).
This is why the optimizations are made and this is why sometimes the search has to be reset to the first frame because seek() won't find the frame if it is already after the wanted frame.
This is because with "newer" codecs the place where a frame should be is not easily to locate; vbr, ...

ferdnyc:
> @ferdnyc If not already known: The seek() function in ffmpeg gets slower when frames later in a stream are to be found. This is especially easily visible when watching a *.flv file and jumping to a late frame, like 2 h in the clip. This is in every program I used not just Openshot.

Oh, no, absolutely... _believe_ me, I'm very familiar with that phenomenon. :grin:  Seems like it depends on the file format somewhat, though, and especially how well-indexed the file is.

For example, I have 20 _Gigabyte_ movie files in MKV containers where I can start playback in VLC, then click anywhere along the timeline — whether it's 10 minutes forward, or 10 minutes before the end of the 2-hour runtime — and it'll jump there virtually instantaneously. And if the file has chapter markers, jumping directly to any one of those **is** instantaneous.

But then there are other files, especially if they use older codecs/containers, that are effectively unseekable. AVIs with broken indexes have to be reindexed before they'll seek correctly, though at least that can be a one-time, relatively automated process. And I have a lot of older QuickTime files where trying to seek within them is basically useless. Even if it does manage to continue playing the video back from the new location, the audio will be completely out of sync (or even _gone_ completely) from that point. But if you play through from the beginning, everything's fine when you reach that point.

No question it's a hard problem, and probably one that there _isn't_ any perfect solution to — other than maybe to just automatically transcode (or at least remux) bad files into a more compatible format _for_ the user. (Instead of waiting until they ask for support, and then advising them to do exactly that. Which we do, quite often.)

> This is why the optimizations are made and this is why sometimes the search has to be reset to the first frame because seek() won't find the frame if it is already after the wanted frame.

Oh, no, that part I totally understand. And I can believe there's no quicker/cleaner way to reset back to the start of the file other than to just completely close the file and reopen it again. 

My only concern is the side effects that can have (as we've seen here). Because in addition to resetting a lot of the C++ data structures, `Close()` frees a **lot** of FFmpeg's C-style pointers, even while the rest of the code is potentially juggling them as well.

`last_video_frame` just gets `.reset()`, which is NBD. And a lot of the caches and lists just get `.clear()`ed, which is also safe:

https://github.com/OpenShot/libopenshot/blob/857297abdf3b3c4d8067b5f32559d10269b2d7e0/src/FFmpegReader.cpp#L606-L609

https://github.com/OpenShot/libopenshot/blob/857297abdf3b3c4d8067b5f32559d10269b2d7e0/src/FFmpegReader.cpp#L611-L623

But `packet` gets set to `NULL` (which caused the bug @SuslikV fixed in this PR), and all of the FFmpeg contexts are not just closed but have their pointers `av_freep()`'d:

https://github.com/OpenShot/libopenshot/blob/857297abdf3b3c4d8067b5f32559d10269b2d7e0/src/FFmpegReader.cpp#L582-L604

https://github.com/OpenShot/libopenshot/blob/857297abdf3b3c4d8067b5f32559d10269b2d7e0/src/FFmpegReader.cpp#L625-L627

That part feels super scary, especially in a multithreaded environment. And _especially_-especially since `Seek()` calls `Close()` and `Open()` separately in series:

https://github.com/OpenShot/libopenshot/blob/857297abdf3b3c4d8067b5f32559d10269b2d7e0/src/FFmpegReader.cpp#L1731-L1733

So they could potentially happen sort of at the code's leisure, with plenty of time for context switches and thread preemptions along the way, I think?

There's also the fact that `Open()` is just as heavy an operation as `Close()`, and it re-makes a lot of the same setup decisions, re-scans and updates the metadata, and re-resizes the caches — all stuff it did the first time around. (Heck, it'll even re-attempt hardware decoding again, even if it already failed on the previous attempt because the input codec isn't compatible.)

Do you think it would be possible (and, separately, do you think it would be _useful_) if we had a separate `Reopen()` method for use in `Seek()` (and potentially elsewhere), one that's specifically geared to closing and reopening an input file all in one shot, reusing all of the same parameters instead of recomputing them like the full `Open()` will do every time? (Which isn't me trying to put you on the hook for _writing_ `FFmpegReader::Reopen()` — I'm just interested in your opinion on whether it makes sense, as you know this code far better than I do.)

My only concern here is to try and root out areas where we may be exposed to problems that ultimately manifest as user-visible bugs, I'm not _**at all**_ criticizing the current code. (Or, if I am, I'm certainly not criticizing the _coding_. Any codebase that evolves over the course of several years, and works against an API as flaky and ever-shifting as FFmpeg's, will always offer opportunities for refactoring — even if the code was structured perfectly when it was originally written. Heck, hardware decoding is a perfect example. `Open()` was a **lot** less heavy before that was added to the mix.)

eisneinechse:
@ferdnyc  Sorry I did not want to create the impression that I am married to that code. And I think the idea is great to implement a Reopen(). This might be kind of hard, though, and it might take some time. I am all for it.
BTW If someone wants to see the difference between a clip that is searchable and one that is not just recode with svt-av1 and use the standard option to get a non-searchable clip; for searchable clips an option is needed:
-irefresh-type 2
Here a line out of a script I use to recompile. The script is quite long and I haven't done anything in weeks (stress is the reason there too):
ffmpeg 2>/dev/null -nostats -hide_banner -i "$1" -nostdin -f rawvideo -pix_fmt yuv420p - | /usr/local/bin/SvtAv1EncApp -i stdin -w $ARG_W -h $ARG_H -enc-mode $ARG_M -rc 2 -intra-period $ARG_LAD -lad $ARG_LAD -tbr $ARG_TBR -n $ARG_F -fps-num $ARG_FPS_N -fps-denom $ARG_FPS_D -irefresh-type 2 -b output.ivf -lp $ARG_LP -tile-rows $ARG_ROW -tile-columns $ARG_COL

ferdnyc:
(For any other Fedora users playing along at home...)

`SvtAv1EncApp` and its companion `SvtAv1DecApp` can apparently be installed via the `svt1-av1` package in the standard repos.

It appears to be an entirely un-UNIX-like mess of nonsense which comes with no man pages and accepts _only_ the argument `-help` for a help listing (not `-h` (obviously as that's "height"), `--help`, `--usage`, or anything else vaguely standard). 

`-help` causes it to dump a three-column, tab-formatted table of arguments with "Description" text like "ConstrainedIntra" and "AltRefStrength" (**#LiterallyNotDescriptive**) that doesn't fit in an 80-character-wide terminal even though 95% of its lines _would_, because whatever idiots created it decided to _end_ each line with a bunch of unnecessary spaces and tabs. **#ForSomeReason!**

(Best of all, the formatting is _still_ broken anyway, because someone missed a memo:)
```
-search-w                	SearchAreaWidth          	Single input             	
-search-h                	SearchAreaHeight         	Single input             	
-num-hme-w               	number_hme_search_region_in_width	Single input             	
-num-hme-h               	NumberHmeSearchRegionInHeight	Single input             	
-hme-tot-l0-w            	HmeLevel0TotalSearchAreaWidth	Single input             	
-hme-tot-l0-h            	HmeLevel0TotalSearchAreaHeight	Single input             	
```

Basically, this code has **#FreshHell** written all over it, and `ffmpeg` can't grow built -in support quick enough. There is [a plugin, with patches available](https://github.com/OpenVisualCloud/SVT-AV1/tree/master/ffmpeg_plugin) to integrate `libsvtav1` (as well as `libsvthevc`), just needs to get applied to the FFmpeg build. 

I may see about tossing patches over to the RPM Fusion maintainer, since the encoder is part of Fedora already. We _just_ rebuilt for the Fedora 31 branch, though, so it may take some time.

eisneinechse:
Here's the best "manual" I could find:
https://github.com/OpenVisualCloud/SVT-AV1/blob/master/Docs/svt-av1_encoder_user_guide.md

There is also one for the decoder
https://github.com/OpenVisualCloud/SVT-AV1/blob/master/Docs/svt-av1_decoder_user_guide.md

ferdnyc:
I submitted a patch (OpenVisualCloud/SVT-AV1#563) to fix their horrid `SvtAv1EncApp -help` formatting. (Making it one of 47 open PRs, all submitted just since February 9. Most of them within the past month!)

PR #289, opened 2019-08-01
Move #endif

eisneinechse:
Move av_hwdevice_get_hwframe_constraints inside the part of the code only compiled when testing VAAPI constraint detection.


jonoomph:
Thanks, LGTM! :+1:

PR #290, opened 2019-08-02
Python install path: remove site-packages detection

ferdnyc:
This is a hail-mary _attempt_ to fix Python install-path logic for non-Debian systems, without breaking things for the Debian folks. Because currently, the Debian special-casing is causing the **wrong** install path to be chosen on Fedora and other `lib64`-using systems, a regression from previous behavior (up to and including 2.4.4-release) where the proper `/usr/lib64/...` path was correctly chosen.

The path `${CMAKE_INSTALL_PREFIX}/lib/pythonM.N/site-packages` is not Debian-specific, and will exist on too many systems where it should _not_ be assumed to be the correct install path.

### DIFF
--- a/src/bindings/python/CMakeLists.txt
+++ b/src/bindings/python/CMakeLists.txt
@@ -72,26 +72,21 @@ if (PYTHONLIBS_FOUND AND PYTHONINTERP_FOUND)
 	                      ${PYTHON_LIBRARIES} openshot)
 
 	### Check if the following Debian-friendly python module path exists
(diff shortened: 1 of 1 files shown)


jonoomph:
Is this one safe to merge? I know we've attempted this a few times, and every time it broke something on our build servers or with local builds... always something. I want to believe though!

ferdnyc:
Only way to be sure is to try, but I think we HAVE to merge it because right now, in attempting to accommodate Debian, that first path is messing up every other distro. And AIUI the _second_ path (the one I didn't delete) is the right-er one for Debian, anyway. So it _should_ be OK. But like I said, only way to find out...

jonoomph:
Makes sense! Let's give it a try.

PR #293, opened 2019-08-04
Remove "dummy" args from ZmqLogger stragglers

ferdnyc:
I somehow missed a few calls, in #266.


ferdnyc:
This is really just cleanup from #266, so I'm going to merge it to complete that effort.

PR #295, opened 2019-08-05
Fix include path list, for subdirectories

ferdnyc:
The include paths listed in `OPENSHOT_INCLUDE_DIRS` by the root `CMakeLists.txt` all need to be absolute, so that `src/CMakeLists.txt` and the rest don't interpret them relative to _their_ current directory.


ferdnyc:
Going to commit this, because it currently doesn't affect anything — include paths aren't used properly, for our own headers. (This is one of the things preventing that.)

Issue #204, opened 2019-03-18
Remove 'using namespace' from include files


schauveau:
The libopenshot header files contain multiple `using namespace` statements for `openshot` and `std`. This is a bad design for several reasons that I will not enumerate here. 
Look for instance in https://stackoverflow.com/questions/5849457/using-namespace-in-c-headers and https://www.acodersjourney.com/top-10-c-header-file-mistakes-and-how-to-fix-them/

ferdnyc:
Just for the record, I spent a very, very long night taking a look at this issue and, well... the results were less than satisfying. I did manage to get all of the `using namespace` directives out of the headers, but the only way I could get the SWIG wrappers to compile was to stick `using namespace openshot;` _into_ the `openshot.i` files instead, and even then the wrappers didn't come out right. (I have no idea if the library even came out right, TBH.)

The results of my efforts are in my [namespace-cleanup](https://github.com/ferdnyc/libopenshot/tree/namespace-cleanup) branch if anyone's interested. What's there _does_ compile, at least for me. But If you try to run OpenShot with it, it'll die right at startup as soon as it tries to create the Timeline object. It doesn't understand `ChannelLayout`, which comes from the `include/ChannelLayouts.h` file. 

`ChannelLayout` is an enum stuffed full of strings that FFmpeg `#define`s. Those in turn come from `include/FFmpegUtilities.h` which _doesn't_ have a `namespace openshot {}` wrapped around it, and it breaks if I add one. I think it's because the FFmpeg headers are included under an `extern "C"`. Something about that seems to be throwing SWIG off its game, now that everything isn't polluted into all of the wrong namespaces, and I'm not sure how to fix it.

ferdnyc:
I took _another_ run at this, and by jove I think I've got it!

Still testing, but libopenshot compiled _and_ passed its tests, and I ran OpenShot against the corresponding `openshot.py` / `_openshot.so`. So far everything seems to be in order. Despite there being not a _single_ `using namespace …` statement anywhere in the headers, **or** in SWIG's `openshot.i` files.

ferdnyc:
My [std-prefixes](https://github.com/ferdnyc/libopenshot/tree/std-prefixes) branch contains the results of that effort, though once I'm more confident that it's _really_ OK I plan to submit it as a PR, at which point I'll link it to this issue.

PR #296, opened 2019-08-05
Remove all 'using namespace' directives from headers

ferdnyc:
This branch is the culmination of a _lot_ of slogging through our source files. (Not to mention, at least two failed prior attempts to achieve the same thing.) However, it does (at least in my testing) finally achieve a lofty goal: Removing **all** traces of `using namespace ...` from the entire `include/**/*.h` tree.

While this was a chore, _and_ makes our code a bit more verbose, it was (IMHO) a worthwhile goal as `using namespace...` at file scope in headers is [considered](https://stackoverflow.com/questions/5849457/using-namespace-in-c-headers) to be [extremely](https://stackoverflow.com/questions/1265039/using-std-namespace) bad [practice](https://www.acodersjourney.com/top-10-c-header-file-mistakes-and-how-to-fix-them/).

(It can be done in a _public_ wrapper header, one that's meant to be used by the _consumer_ of the library — both `JuceHeader.h` and Qt's headers optionally add a `using namespace...` statement for their own namespaces — but that shouldn't be relied on by the library _itself_, nor should `using namespace std;` **ever** be done at file scope in a header. See the links above for copious justifications why that's a horrible thing.)

Because we're using SWIG, which is _much_ pickier than `g++` when it comes to namespacing, even after I got libopenshot to compile without `using namespace...` statements in its headers, I had to do a lot _more_ work to get the bindings to build cleanly. Mostly what it required was a _lot_ of prefixing for `std::`, `juce::` _and_ `openshot::` types that we tend to use unprefixed. With very few exceptions that was only necessary in the headers, though. 

(And where those newly-prefixed types were in declarations, sometimes also in the corresponding implementation. That was to make Doxygen happy, if nothing else, since it sometimes has trouble matching up e.g. `Frame::Frame(int64_t number, int width, int height, std::string color)` in the header file with `Frame::Frame(int64_t number, int width, int height, string color)` in the `.cpp` source.)

Doing that, however, let me eliminate the double-mapping in the SWIG files for `std::shared_ptr<openshot::Frame>` _and_ `std::shared_ptr<Frame>` (we only need the former now, as it should be), and it allowed me to remove the `namespace std {}` around our SWIG templates.

So far in my testing, everything's worked just as before with this, even running OpenShot with it. But, I've created this PR from a branch in the project repo, so that we can test it with the builders first.

Fixes: #204


SuslikV:
It not so bad as was said. And `uses` always placed last of all headers included. And if not in the header file - why not to move it into the body?

ferdnyc:
> It not so bad as was said. And `uses` always placed last of all headers included. 

Oh, cool. I'll inform Bjarne Stroustrup that it's no big deal to pollute callers' namespaces with all of `std::`. (It is most definitely a bad thing.)

Or do you mean that the code changes weren't that bad? Trust me, they took me basically an _entire day_ (just for this **third** attempt), and at points along the way the SWIG compile was spewing literally THOUSANDS of lines of errors. Its generated code is very fragile, and when it breaks it breaks big.

> And if not in the header file - why not to move it into the body?

They're already there:
```console
$ grep 'using namespace' src/**/*.cpp
src/AudioBufferSource.cpp:using namespace std;
src/AudioBufferSource.cpp:using namespace openshot;
src/AudioReaderSource.cpp:using namespace std;
src/AudioReaderSource.cpp:using namespace openshot;
src/AudioResampler.cpp:using namespace std;
src/AudioResampler.cpp:using namespace openshot;
src/CacheBase.cpp:using namespace std;
src/CacheBase.cpp:using namespace openshot;
src/CacheDisk.cpp:using namespace std;
src/CacheDisk.cpp:using namespace openshot;
src/CacheMemory.cpp:using namespace std;
src/CacheMemory.cpp:using namespace openshot;
src/ChunkReader.cpp:using namespace openshot;
src/ChunkWriter.cpp:using namespace openshot;
src/ClipBase.cpp:using namespace openshot;
src/Clip.cpp:using namespace openshot;
src/Color.cpp:using namespace openshot;
src/Coordinate.cpp:using namespace std;
src/Coordinate.cpp:using namespace openshot;
src/CrashHandler.cpp:using namespace std;
src/CrashHandler.cpp:using namespace openshot;
src/DecklinkInput.cpp:using namespace std;
src/DecklinkOutput.cpp:using namespace std;
src/DecklinkReader.cpp:using namespace openshot;
src/DecklinkWriter.cpp:using namespace openshot;
src/DummyReader.cpp:using namespace openshot;
src/EffectBase.cpp:using namespace openshot;
src/EffectInfo.cpp:using namespace openshot;
src/effects/Bars.cpp:using namespace openshot;
src/effects/Blur.cpp:using namespace openshot;
src/effects/Brightness.cpp:using namespace openshot;
src/effects/ChromaKey.cpp:using namespace openshot;
src/effects/ColorShift.cpp:using namespace openshot;
src/effects/Crop.cpp:using namespace openshot;
src/effects/Deinterlace.cpp:using namespace openshot;
src/effects/Hue.cpp:using namespace openshot;
src/effects/Mask.cpp:using namespace openshot;
src/effects/Negate.cpp:using namespace openshot;
src/effects/Pixelate.cpp:using namespace openshot;
src/effects/Saturation.cpp:using namespace openshot;
src/effects/Shift.cpp:using namespace openshot;
src/effects/Wave.cpp:using namespace openshot;
src/examples/ExampleBlackmagic.cpp:using namespace openshot;
src/examples/Example.cpp:using namespace openshot;
src/examples/ExampleFaststart.cpp:using namespace openshot;
src/examples/ExampleOrig.cpp:using namespace openshot;
src/examples/MemTest.cpp:using namespace openshot;
src/FFmpegReader.cpp:using namespace openshot;
src/FFmpegWriter.cpp:using namespace openshot;
src/Fraction.cpp:using namespace openshot;
src/Frame.cpp:using namespace std;
src/Frame.cpp:using namespace openshot;
src/FrameMapper.cpp:using namespace std;
src/FrameMapper.cpp:using namespace openshot;
src/ImageReader.cpp:using namespace openshot;
src/ImageWriter.cpp:using namespace openshot;
src/KeyFrame.cpp:using namespace std;
src/KeyFrame.cpp:using namespace openshot;
src/PlayerBase.cpp:using namespace openshot;
src/Point.cpp:using namespace std;
src/Point.cpp:using namespace openshot;
src/Profiles.cpp:using namespace openshot;
src/QtImageReader.cpp:using namespace openshot;
src/QtPlayer.cpp:using namespace openshot;
src/ReaderBase.cpp:using namespace openshot;
src/RendererBase.cpp:using namespace openshot;
src/Settings.cpp:using namespace std;
src/Settings.cpp:using namespace openshot;
src/TextReader.cpp:using namespace openshot;
src/Timeline.cpp:using namespace openshot;
src/WriterBase.cpp:using namespace openshot;
src/ZmqLogger.cpp:using namespace std;
src/ZmqLogger.cpp:using namespace openshot;
```

And like I said, there's _nothing_ wrong with that. 

But when those statements are placed in our headers, it forces the same namespace pollution on any code that's going to be **linked** with libopenshot, which has the potential to cause real problems for those callers.

It also means that _our_ header files' references to things in those namespaces are ambiguous, relying on that namespacing to resolve the names. What if a calling program does a `using namespace boost::vector;` _before_ `#include "OpenShot.h"`? Now our `vector<Frame>` has two possible meanings and the compiler has no good way to choose between them. Whereas if it's written correctly, as `std::vector<openshot::Frame>`, there's no room for ambiguity.

The OpenShot API is much more robust and better suited to use by external callers, with those `using` directives sanitized out of there.

SuslikV:
> They're already there...

Why `std::`
https://github.com/OpenShot/libopenshot/blob/066e2558eee55c8b48ed3ed0e8e055f8e61eb3bf/src/CacheBase.cpp#L61
was added here?

ferdnyc:
> > They're already there...
> 
> Why `std::`
> was added here?

\*shrug\* `std` is a _very_ big, cluttered namespace, and while it's **acceptable** to use `using namespace std;` in source files (meaning, it doesn't have the same _dangers_ associated with it, compared to using it in a header, it's still generally not considered very good programming practice by many people.

(See e.g. these two stack overflow Q&As:)
* [Using std namespace](https://stackoverflow.com/questions/1265039/using-std-namespace)
* [Why is “using namespace std;” considered bad practice?](https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice)

And, certainly, it's never _wrong_ to access functions in the `std` namespace by prefixing them, rather than unprefixed via `using namespace std;` So I guess I've been trying to "encourage" the code in that direction. 

That's ultimately a question of style, though. It's not as critical as the removal of the using directives from the headers. And I'd certainly remove the `std::` prefixing in the .cpp files from this PR, if requested.

Actually, though, I'm glad you pointed that out, because there are two things I missed in that code. I failed to also prefix the `atoll` that comes a few lines later:

https://github.com/OpenShot/libopenshot/blob/066e2558eee55c8b48ed3ed0e8e055f8e61eb3bf/src/CacheBase.cpp#L69-L75

_and_  in C++11, that line can use the `std::string`-conversion functions, instead of going through the C string type. So, that could more cleanly be written as:

```cpp
max_bytes = std::stoll(root["max_bytes"].asString());
```

(Actually, I don't quite understand why it wasn't written like this, anyway:
```cpp
max_bytes = atoll(root["max_bytes"].asCString());
```
...maybe `asCString()` is a newer addition to `Json::Value`.)


Anyway, since `atoll` appears in several other source files, I'll create a separate PR to convert all of them over to `std::stoll`.

ferdnyc:
BTW, if anyone is interested in testing this change end-to-end, official builds of OpenShot linked with a libopenshot built off of this branch are available:
* Linux: http://gitlab.openshot.org/OpenShot/openshot-qt/-/jobs/12823/artifacts/file/build/OpenShot-v2.4.4-dev2-1566722386-05a53784-f17a7ed4-x86_64.AppImage
* MacOS: http://gitlab.openshot.org/OpenShot/openshot-qt/-/jobs/12824/artifacts/file/build/OpenShot-v2.4.4-dev2-1566722496-05a53784-f17a7ed4-x86_64.dmg
* Windows 64-bit: http://gitlab.openshot.org/OpenShot/openshot-qt/-/jobs/12825/artifacts/file/build/OpenShot-v2.4.4-dev2-1566722515-67934675-f17a7ed4-x86_64.exe
* Windows 32-bit: http://gitlab.openshot.org/OpenShot/openshot-qt/-/jobs/12826/artifacts/file/build/OpenShot-v2.4.4-dev2-1566722514-67934675-f17a7ed4-x86.exe

(It appears the Linux and MacOS builders missed my most recent merge commit (6793467) to resolve a small conflict, while the Windows builders built off the up-to-date tree. Not sure exactly why that would happen, but it's not important. Nothing in that latest commit has any impact on this PR whatsoever.)

ferdnyc:
Note: It's very likely that other PRs will cause conflicts with this one when merged, or vice versa. (I know for sure #305 will.)

I guess my preference would be, let's merge those others first, and then I'll clean up any conflicts created here.

ferdnyc:
> I'll have to mirror them somewhere else and post new links, sorry.

Finally got around to doing that.

Linux: https://drive.google.com/open?id=144dTaqUDCK9TdOP3S5WFwRsq-CUipNyQ
MacOS: https://drive.google.com/open?id=16ZsQiUVhjMOpv_7uToesILYLhIELEXd6
Win64: https://drive.google.com/open?id=1YgLXlka3ZvNSHmY3crbpv7Do3r94KrvL
Win32: https://drive.google.com/open?id=1LJxd228IdRM1lieyQsxch-s3yHmmIhJC

jonoomph:
Okay, I can wait until most other PRs are merged before revisiting this one. However, it's probably gonna have a bunch of conflicts. But this is AWESOME!

ferdnyc:
Huh, that was actually really painless. So far.

I suspect other still-pending PRs will cause bigger conflicts. Though I suppose we could merge this now (if it passes), and leave _those_ PRs with all the conflicts instead... :smiling_imp:

ferdnyc:
Whoops, that broke? Hmm, I'll figure out why...

ferdnyc:
Should be fixed, the issue was actually in other new code.

ferdnyc:
I have no idea why the first check is stuck on "pending", since it clearly finished successfully, as clicking on it shows. Hopefully GitHub will catch on eventually.

ferdnyc:
Resolved knock-on issues after the merge of #305.

ferdnyc:
Surprisingly enough, @jonoomph, this branch remains conflict-free, and I think we're actually in a good position to merge it. I'm going to manually merge `develop` into it from the command line, which will trigger another run on the GitLab builders. But if that passes, I think it's g2g if we feel like living dangerously.

jonoomph:
Here we go... merging

jonoomph:
:crossed_fingers: :crossed_fingers: :crossed_fingers:

jonoomph:
I'm always afraid I'll accidentally choose the middle finger icon, which looks almost the same. lol

ferdnyc:
> I'm always afraid I'll accidentally choose the middle finger icon, which looks almost the same. lol

...Would that be incorrect, though? :rofl:

ferdnyc:
I'm so glad this is finally in. Now I can resurrect my changes to the include-directory management in CMake, and lose all of those `#include "../../../include/blah.h"` paths from both our `.cpp` and `.i` files.

It _works_, doing it that way, but it means we're not forcing ourselves to use the headers the same way library consumers will... and that's a dangerous game. A library's implementation should double as its own reference API consumer.

So, more breakage to come! Stay tuned. :angel:

PR #297, opened 2019-08-06
Update table in HW-ACCEL.md for Doxygen

ferdnyc:
By using actual Unicode characters for the various ✔️ and ❌ marks in the table, Doxygen will format it properly.
Also:
- Changed to superscripted numerals, for the various callouts
- Reformatted notes into a corresponding ordered list
- Added invisible spaces after the marks without a callout, to balance centering
- Added center-alignment markers to appropriate table columns

### DIFF
--- a/doc/HW-ACCEL.md
+++ b/doc/HW-ACCEL.md
@@ -7,20 +7,22 @@ our support for this in the future!
 
 The following table summarizes our current level of support:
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
This is a pure documentation change, and since there have been no objections (and the current state of the `HW-ACCEL.md` formatting in Doxygen, _without_ this PR, is absolutely abysmal), I'm going to go ahead and merge it.

PR #300, opened 2019-08-07
Rename AV_GET_CODEC_TYPE to AVSTREAM_GET_CODEC_TYPE

### LINKED ISSUE #299: codec_type has number of references in FFmpeg, while usage specific in libopenshot is not mentioned
Logically the `AV_GET_CODEC_TYPE`  always points to part of `AVCodecContext` structure through the next code of libopenshot (audio and video):
https://github.com/OpenShot/libopenshot/blob/05c21ae2d5adabf36916ce6804a08fa54d587450/src/FFmpegWriter.cpp#L1080
https://github.com/OpenShot/libopenshot/blob/05c21ae2d5adabf36916ce6804a08fa54d587450/src/FFmpegWriter.cpp#L1213

but uses different places where the data from `AVCodecContext` structure is referenced. Here I mean, that the source of the `codec_type` pointer inside libopenshot - only `AVCodecContext` structure (which it belongs too).

I think the `AV_GET_CODEC_TYPE` should point only to `codec_type` (enumeration) as part of the `AVCodecContext` structure, no matter where this structure pointer is mentioned.
**Thus,**
current `AVStream` structure specific reference to `codec_type` should be named as `AV_GET_CODEC_TYPE_AVSTREAM` (instead of just `AV_GET_CODEC_TYPE`)
**or**
new definition added as `AV_GET_CODEC_TYPE_CODEC_CONTEXT` that points exclusively to the `AVCodecContext`'s corresponding field.

The AVCodecContext Struct itself is using different types of enumerations (`CodecType` and `AVMediaType`) for the same pointer (`codec_type`), for FFmpeg below v2.x:
https://ffmpeg.org/doxygen/0.5/structAVCodecContext.html
https://ffmpeg.org/doxygen/1.0/structAVCodecContext.html

while in the code, for FFmpeg 2.x, it just points to "some" data of unknown type (I hope, that FFmpeg below 2.x is not possible to use with the libopenshot and nobody will try it ever, so this is minor part of the issue).
https://github.com/OpenShot/libopenshot/blob/05c21ae2d5adabf36916ce6804a08fa54d587450/include/FFmpegUtilities.h#L266

ferdnyc:
There are a lot of ways to access `codec_type` in FFmpeg, and the macro is only useful for the path that goes through `AVStream`. (It's necessary due to `AVStream::codec` being replaced with `AVStream::codecpar` at some point in FFmpeg's evolution.)

Fixes #299

### DIFF
--- a/include/FFmpegUtilities.h
+++ b/include/FFmpegUtilities.h
@@ -284,5 +284,7 @@
 		#define AV_COPY_PARAMS_FROM_CONTEXT(av_stream, av_codec)
 	#endif
 
+	// Deprecated macro aliases, will be removed in a future release
+	#define AV_GET_CODEC_TYPE(av_stream) _Pragma ("GCC warning \"AV_GET_CODEC_TYPE is deprecated, use AVSTREAM_GET_CODEC_TYPE\"") AVSTREAM_GET_CODEC_TYPE(av_stream)
 
 #endif
--- a/include/FFmpegUtilities.h
+++ b/include/FFmpegUtilities.h
@@ -157,7 +157,7 @@
     	#define AV_FREE_FRAME(av_frame) av_frame_free(av_frame)
 		#define AV_FREE_PACKET(av_packet) av_packet_unref(av_packet)
 		#define AV_FREE_CONTEXT(av_context) avcodec_free_context(&av_context)
-		#define AV_GET_CODEC_TYPE(av_stream) av_stream->codecpar->codec_type
+		#define AVSTREAM_GET_CODEC_TYPE(av_stream) av_stream->codecpar->codec_type
 		#define AV_FIND_DECODER_CODEC_ID(av_stream) av_stream->codecpar->codec_id
 		auto AV_GET_CODEC_CONTEXT = [](AVStream* av_stream, AVCodec* av_codec) { \
 			AVCodecContext *context = avcodec_alloc_context3(av_codec); \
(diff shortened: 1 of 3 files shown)


SuslikV:
Small remark, `AVStream::codec` was not just renamed - the pointer to `AVCodecContext` struct was replaced with the new pointer to the `AVCodecParameters` struct in FFmpeg (two different structures). While the `codec_type` from `AVCodecContext` was untouched (as for today), and thus can be referenced directly from the libopenshot code.

ferdnyc:
@SuslikV You're right, not renamed, [replaced](https://ffmpeg.org/pipermail/ffmpeg-cvslog/2016-April/099152.html). I've updated the description, thanks.

ferdnyc:
I was thinking about renaming this, since `AVSTREAM_` is a prefix used by FFmpeg's own headers... but so is `AV_` (_extensively_), so that's a larger issue that would be better to address en masse rather than one-off, if we did choose to address it.

ferdnyc:
I did add an alias with a deprecation warning, though, just in case. Now if `AV_GET_CODEC_TYPE()` is used anywhere, it'll produce a warning:
```
$ make FFmpegWriter.o
Building CXX object src/CMakeFiles/openshot.dir/FFmpegWriter.cpp.o
FFmpegWriter.cpp:740:13: warning: AV_GET_CODEC_TYPE is deprecated, use
 AVSTREAM_GET_CODEC_TYPE
  740 |  if (info.has_audio && audio_codec && AV_GET_CODEC_TYPE(audio_st) ==
 AVMEDIA_TYPE_AUDIO && AV_GET_CODEC_ATTRIBUTES(audio_st, audio_codec)->frame_size <= 1)
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
(I'm not sure why the "arrow" points to the wrong place on the line, but meh.)

ferdnyc:
I've reconsidered this, see [this comment] (https://github.com/OpenShot/libopenshot/issues/299#issuecomment-520110403) for explanation.

PR #302, opened 2019-08-10
QtImageReader: Safer ReSVG file extension checks

ferdnyc:
This changes the type of the private `path` variable to QString internally (which we were converting it to for a bunch of operations anyway), and uses QString's more robust string-manipulation and
-comparison methods to ensure that only filenames that _end in_ an `.svg`/`.svgz` extension (case-insensitively) are recognized as SVG files.

Previously, a filename such as "Title.svg.png" would be interpreted as an SVG file.

The change in the `path` type is purely internal, all of the class's interfaces remain the same, and accept `std::string path`. We simply convert it to/from QString at opposite places in the code, now.

### DIFF
--- a/include/QtImageReader.h
+++ b/include/QtImageReader.h
@@ -67,7 +67,7 @@ namespace openshot
 	class QtImageReader : public ReaderBase
 	{
 	private:
-		string path;
+		QString path;
 		std::shared_ptr<QImage> image;			///> Original image (full quality)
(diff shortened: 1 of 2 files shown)


jonoomph:
Are we sure this works with SWIG? I'll go ahead and merge and I guess we'll see how it works. :wink:

ferdnyc:
SWIG shouldn't be affected, the `path` is a private member so the bindings never see it.

PR #303, opened 2019-08-10
Use std::stoll to convert JSON values

ferdnyc:
The code to convert JSON values from string to numeric types was using the old C `atoll()` function from `stdlib.h` for type-conversion. But C++11 has `std::stoll()` to convert directly from `std::string`, instead of going through C byte strings.

This PR switches from `#include <stdlib.h>` to `#include <cstdlib>`, and rewrites the conversions to use `std::stoll()`.

### DIFF
--- a/include/CacheBase.h
+++ b/include/CacheBase.h
@@ -32,6 +32,7 @@
 #define OPENSHOT_CACHE_BASE_H
 
 #include <memory>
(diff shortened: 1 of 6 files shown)


ferdnyc:
Merging this, it's minor and pretty safe.

PR #305, opened 2019-08-11
QtTextReader and QtHtmlReader

jeffski:
I've been looking in to making some improvements to working with text in in libopenshot. One main feature in particular is to have text of different colours in a sentence. You see it quite a lot on news and social media videos where some of the words are highlited. Unfortunately I couldn't see a way to do this with Magick++.

The best solution I could come up with was to use Qt's HTML features to do this and created QtHtmlReader - it supports all the HTML features here: https://doc.qt.io/qt-5/richtext-html-subset.html. Along the way we created a QtTextReader as well.

An added bonus is that both of these are much faster than ImageMagick, nearly 3 times faster in my tests, even the HTML is faster.

We've tried to follow the existing libopenshot code as best as possible. If you would like to merge these readers please review and suggest any changes/quaility/improvements. @ferdnyc I have removed the std namespace following your current pull request (#296).

Special thanks to @jediserg who figured out the Qt stuff and wrote most of the code for this.


ferdnyc:
@jeffski Wow, this looks great thanks! I'll make some time to offer up my review either today or tomorrow, but a few off-the-top-of-my-head thoughts:

- [ ] Adding some unit tests in `tests/QtHtmlReader_Tests.cpp` and `tests/QtTextReader_Tests.cpp` would be great, if possible. We really _should_ make it a requirement that tests accompany any major new addition of code, certainly for entire new _classes_. (But we're not at that point yet, so I wouldn't say they're a hard requirement. _Somebody_ should definitely write them, but it doesn't have to be you guys.)
- [ ] By the same token, this will need integration into the Python and Ruby bindings, but I'll be happy to take care of that.
- [ ] One of the things unit tests would give us is a way to _see_ the code in use. The classes look to be well-documented, so that provides some example code already, but a quick demo of each new Reader's capabilities — whether it's in the form of tests, or just a quick sample program — would be great in helping us to mentally connect the code to its concrete output.

And... this last one's more of a high-level thing, which is why I'm a bit reluctant to even bring it up this late in the game. (So, please take it as a discussion-starter only, _not_ a request that major changes be made to the code just to accommodate a half-baked idea...):
- [ ] I notice that the readers have parameters for things like background color, text positioning, etc: Particularly because one of them is an _HTML_ text renderer, I'm just curious whether you explored the possibility of just accepting a CSS stylesheet to control all of that (and plenty more) — either instead of, or in addition to, the individual parameters? (Assuming that's even possible with the QTextDocument class?)

Like I said, just off the top of my head, some of the above may be addressed (or I may rethink) when I get to looking at the code. So, feel free to ignore anything I wrote, if I'm going to realize later that it was dumb anyway. :wink:

jeffski:
Thanks for the early enthusiasm and feedback.

Some quick responses:
- I could try and tackle the tests, at least set up a very basic test to start with, but how to run them? I tried `make test` and got an error: `make: *** No rule to make target 'test'.  Stop.`

- It would be great if someone with more knowledge could tackle the bindings.

- I can add an example to src/examples, those should be straight forward.

Regarding the stylesheet - that looks like a great idea, you could probably pass in a stylesheet string as a parameter and then QTextDocument has a `setDefaultStyleSheet` method to apply it. I think the other parameters are valid/useful - just a background color (I guess you could set that using body css), gravity might also be useful to set where you want the html to start rendering - otherwise it always starts from the top left. I guess you could set body width and height to the viewport size and then vertically/horizontally align but that is always a pain with HTML. I'll experiment a bit more.

One final thing, I ran `make doxygen` and only the new files threw warnings: `multiple use of section label 'LICENSE' while adding section`. Not sure if that's an issue or not?

ferdnyc:
I actually just pushed a commit (3187167a7d980e0642bb57797d3c7d1be78cde17) to this PR's branch in my fork, which creates a `src/examples/ExampleHtml.cpp`, and adds a CMake target `openshot-html-test` to build it.

ferdnyc:
(Specifically, in your build directory with that commit applied, you can `make openshot-html-test` (but it'll also build by default with just `make`), and you can run it with `src/openshot-html-test` **from the build (not src) dir**.)

The reason I published that is, I'm currently getting segfaults whenever I try to run it, from within Qt itself. It looks like it's failing in its Font-lookup methods, even if I add font styling to the test html. Here's a gdb traceback, though I don't have all of the debug symbols installed for it to be completely useful:

```
#0  0x00007ffff6f663f8 in  () at /lib64/libQt5Gui.so.5
#1  0x00007ffff6f6773a in QGuiApplication::font() () at /lib64/libQt5Gui.so.5
#2  0x00007ffff7039f28 in QFontDatabase::load(QFontPrivate const*, int) ()
    at /lib64/libQt5Gui.so.5
#3  0x00007ffff700f5cb in QFontPrivate::engineForScript(int) const ()
    at /lib64/libQt5Gui.so.5
#4  0x00007ffff7042919 in QTextEngine::fontEngine(QScriptItem const&, QFixed*, QFixed*, QFixed*) const () at /lib64/libQt5Gui.so.5
#5  0x00007ffff704451e in QTextEngine::shapeText(int) const ()
    at /lib64/libQt5Gui.so.5
#6  0x00007ffff704537e in QTextEngine::shape(int) const ()
    at /lib64/libQt5Gui.so.5
#7  0x00007ffff7055476 in QTextLine::layout_helper(int) ()
    at /lib64/libQt5Gui.so.5
#8  0x00007ffff7056493 in QTextLine::setLineWidth(double) ()
    at /lib64/libQt5Gui.so.5
#9  0x00007ffff7097dfd in  () at /lib64/libQt5Gui.so.5
#10 0x00007ffff7099739 in  () at /lib64/libQt5Gui.so.5
#11 0x00007ffff709f38b in  () at /lib64/libQt5Gui.so.5
#12 0x00007ffff709faa0 in  () at /lib64/libQt5Gui.so.5
#13 0x00007ffff70a0360 in QTextDocumentLayout::doLayout(int, int, int) ()
    at /lib64/libQt5Gui.so.5
#14 0x00007ffff70a0428 in  () at /lib64/libQt5Gui.so.5
#15 0x00007ffff70a0b65 in  () at /lib64/libQt5Gui.so.5
#16 0x00007ffff70a180f in QTextDocumentLayout::documentChanged(int, int, int)
    () at /lib64/libQt5Gui.so.5
#17 0x00007ffff7067cb6 in QTextDocument::documentLayout() const ()
    at /lib64/libQt5Gui.so.5
#18 0x00007ffff7f385dc in openshot::QtHtmlReader::Open() (this=0x7fffffffcc20)
    at /home/ferd/rpmbuild/REPOS/libopenshot/src/QtHtmlReader.cpp:76
#19 0x00007ffff7f3831e in openshot::QtHtmlReader::QtHtmlReader(int, int, int, int, openshot::GravityType, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)
    (this=0x7fffffffcc20, width=720, height=480, x_offset=5, y_offset=5, gravity=openshot::GRAVITY_CENTER, html="<span style=\"font-family:sans-serif; color:#fff;\"><b>Check out</b> this Text!</span>", background_color="#000000")
    at /home/ferd/rpmbuild/REPOS/libopenshot/src/QtHtmlReader.cpp:48
#20 0x00000000004094b6 in main(int, char**) (argc=1, argv=0x7fffffffd138)
    at /home/ferd/rpmbuild/REPOS/libopenshot/src/examples/ExampleHtml.cpp:49
```

ferdnyc:
I wonder if it's maybe just necessary to set an application default font — I guess using `QGuiApplication::setFont()`, looks like — before using the QtHtmlReader? And if so, I wonder if the Reader could do that itself as a fallback, in case the user doesn't. (The user should still be able to override it, of course.)

ferdnyc:
> Thanks for the early enthusiasm and feedback.
> 
> Some quick responses:
> 
> * I could try and tackle the tests, at least set up a very basic test to start with, but how to run them? I tried `make test` and got an error: `make: *** No rule to make target 'test'.  Stop.`

Yeah, because `test` is a reserved word in CMake now, we had to name our test target `os_test`. We should probably print a hint to that effect, at the end of the compile.

Then again, I'm pretty sure just bare `make` both builds and _runs_ the tests, after building everything else.

> * It would be great if someone with more knowledge could tackle the bindings.

\*nod\* I've already got that done, in fact I first hit the segfault issue attempting to use the HtmlReader from Python. And then I had to back up a step and write a C++ test program, to see whether it was the bindings that were crashing. :laughing: 

> One final thing, I ran `make doxygen` and only the new files threw warnings: `multiple use of section label 'LICENSE' while adding section`. Not sure if that's an issue or not?

It's only an issue for the documentation (and not a critical one there, either), but I'll fix that up. The code you based yours on is a bit out of date compared to the repo, I noticed that when I was looking over the various CMakeLists.txt files. I had originally merged `develop` into the branch to catch it up, but then I backed that out too just in case it was the source of the segfaulting. (But it wasn't, so I'll probably reapply it.) But, the license-comment formatting is one of the things we'd subsequently fixed in the rest of the files.

Oh, and to build the docs, it's better to use `make doc`. That'll set up the output path and call the internal `doxygen` target for you, if it was found during the CMake setup process.

ferdnyc:
> Regarding the stylesheet - that looks like a great idea, you could probably pass in a stylesheet string as a parameter and then QTextDocument has a `setDefaultStyleSheet` method to apply it. I think the other parameters are valid/useful - just a background color (I guess you could set that using body css), gravity might also be useful to set where you want the html to start rendering - otherwise it always starts from the top left. I guess you could set body width and height to the viewport size and then vertically/horizontally align but that is always a pain with HTML. I'll experiment a bit more.

Yeah, regarding gravity... I've been thinking about that a lot more. I think it _works_ to use gravity as a proxy for text-alignment, but I think it might feel limiting or awkward down the road. 

A user may very well **want** to have left-aligned text placed in the lower-right corner of the screen. In theory, they could do that by passing in HTML like this:
```html
<div style="width: 50%; text-align:left; position:absolute; right: 0px; bottom:0px;">
<p>Paragraph text, paragraph text, paragraph text.</p>
<p>More paragraph text</p>
</div>
```

...It might prove more flexible if gravity controlled the alignment of the _rendered output_ within the full frame. (Though, if the output is always full-frame, that could also be dispensed with entirely, since any sort of gravity features can be more flexibly emulated with HTML positioning like I showed above.)

ferdnyc:
OK, I've merged our `develop` branch into yours, so you're up to date with the repo source, and fixed the license references in those file comments to make Doxygen happy.

ferdnyc:
And, actually, just for fun I pushed my commit 3187167a7d980e0642bb57797d3c7d1be78cde17 (`src/examples/ExampleHtml.cpp` and the CMake target to build it) onto your branch as well.

ferdnyc:
> The reason I published that is, I'm currently getting segfaults whenever I try to run it, from within Qt itself. It looks like it's failing in its Font-lookup methods, even if I add font styling to the test html. Here's a gdb traceback, though I don't have all of the debug symbols installed for it to be completely useful:

Ooh, I may know why that's happening! The test code isn't being compiled with the Qt options, so there's no `-fPIC` compiler flag, which Qt requires. So, **#MyBad**.

ferdnyc:
> > The reason I published that is, I'm currently getting segfaults whenever I try to run it, from within Qt itself. It looks like it's failing in its Font-lookup methods, even if I add font styling to the test html. Here's a gdb traceback, though I don't have all of the debug symbols installed for it to be completely useful:
> 
> Ooh, I may know why that's happening! The test code isn't being compiled with the Qt options, so there's no `-fPIC` compiler flag, which Qt requires. So, **#MyBad**.

Well, that wasn't it either. But I think I see now.

Using `QTextDocument` as the renderer actually forces any application using the class to have instantiated a full `QGuiApplication` or `QCoreApplication`, looks like. (I think we can get away with `QCoreApplication`, and not have to involve the Gui. _Possibly_.) I'll rewrite my `ExampleHtml.cpp` code accordingly.

src/QtHtmlReader.cpp
```diff
@@ -0,0 +1,250 @@
...
+			return;
+		}
+
+		//set background
+		painter.setBackground(QBrush(background_color.c_str()));
+
+		//draw text
+		QTextDocument text_document;
+		text_document.setTextWidth(width);
+		text_document.setHtml(html.c_str());
```
ferdnyc: Oh yeah, I knew there was something I forgot. [Qt's docs say](https://doc.qt.io/qt-5/qtextdocument.html#undoRedoEnabled-prop) that QTextDocument has an undo/redo stack, and it's _enabled_ by default? 

I assume there must be some overhead associated with that. (Though it doesn't seem to be very well-documented just _how much_ it might be — I'm also not clear on whether actions are automatically added to the stack, or only when explicitly added by the code.) Since the caller has no access to the QTextDocument, and **we** don't use the stack, a `text_document.setUndoRedoEnabled(false);` is probably worth doing here.

jeffski:
> Using `QTextDocument` as the renderer actually forces any application using the class to have instantiated a full `QGuiApplication` or `QCoreApplication`, looks like. (I think we can get away with `QCoreApplication`, and not have to involve the Gui. _Possibly_.) I'll rewrite my `ExampleHtml.cpp` code accordingly

Sorry, I forgot to mention that anywhere - I've been using `QApplication` at the start of my main function. I just tested `QGuiApplication` which worked fine but with `QCoreApplication` I get an error trying to load fonts.

jeffski:
I'll keep looking in to the gravity issue. One problem with the QtTextDocument is it only supports a subset of HTML and CSS (https://doc.qt.io/qt-5/richtext-html-subset.html), it's quite restrictive compared to what you might be used to with a modern web browser. For example padding only seems to work on table cells. It also doesn't look like `position` is supported so absolute positioning probably isn't possible. I think users would need to get creative with how they use it.

jeffski:
I pulled the latest changes and got the following errors when building:

`make[2]: *** No rule to make target '../include/Version.h', needed by 'src/bindings/ruby/openshotRUBY_wrap.cxx'.  Stop.
CMakeFiles/Makefile2:399: recipe for target 'src/bindings/ruby/CMakeFiles/rbopenshot.dir/all' failed
make[1]: *** [src/bindings/ruby/CMakeFiles/rbopenshot.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[2]: *** No rule to make target '../include/Version.h', needed by 'src/bindings/python/openshotPYTHON_wrap.cxx'.  Stop.
CMakeFiles/Makefile2:344: recipe for target 'src/bindings/python/CMakeFiles/_pyopenshot.dir/all' failed
make[1]: *** [src/bindings/python/CMakeFiles/_pyopenshot.dir/all] Error 2`

I got round it using `-DENABLE_RUBY=0`

Just pointing this out in case there is an issue or maybe I need to update or install a dependency?

include/QtHtmlReader.h
```diff
@@ -0,0 +1,137 @@
+/**
+ * @file
+ * @brief Header file for QtHtmlReader class
+ * @author Jonathan Thomas <jonathan@openshot.org>
+ *
```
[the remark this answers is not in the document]
jeffski: Done. Thanks

include/QtTextReader.h
```diff
@@ -0,0 +1,148 @@
...
+	class QtTextReader : public ReaderBase
+	{
+	private:
+		int width;
+		int height;
+		int x_offset;
+		int y_offset;
+		std::string text;
+		std::string font;
+		double size;
```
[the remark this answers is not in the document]
jeffski: Have changed to font_size instead of just size.

src/QtHtmlReader.cpp
jeffski: Done

jeffski:
Some quirks I thought I'd just mention:

- Compared to ImageMagick the font size is not the same, Qt seems to be a bit bigger
- In Qt you can use the actual font name - i.e. Work Sans Light where as with IM you need to use the IM name - i.e. Work-Sans-Light
- Qt seems to handle transparncy the wrong way round from standard HTML - so its #argb instead of #rgba - see here: https://doc.qt.io/qt-5/qcolor.html#setNamedColor - so if you want to set a transparent background you need to set it as #00000000 - I think TextReader/ImageMagick allow "none".

ferdnyc:
> I pulled the latest changes and got the following errors when building:

That can be cleared up by either deleting and recreating your build dir entirely, or doing an `rm CMakeCache.txt` in the existing build dir and running CMake again. The build environment changed significantly since the previous version of the code.

I'm running a build now to double-check, but since `src/bindings/ruby/openshotRUBY_wrap.cxx` is a generated file that doesn't exist in the source, an out-of-date builddir is the most reasonable explanation.

ferdnyc:
> Some quirks I thought I'd just mention:
> 
> * Compared to ImageMagick the font size is not the same, Qt seems to be a bit bigger

And it'll likely be more or less different depending on the font used, due to the complex conversion of font pitch, point sizes, and other layout parameters into pixel locations that both tools perform (in different ways). Font layout in general is not consistent between different renderers, unfortunately never will be. 

Heck, here's the word "This" in 15pt Bodoni Std, Helvetica Neue LT Pro, Arial, and Source Sans Pro...

As rendered by the GIMP:

image

And in Google Chrome:

image

Those are both actual size, BTW. I originally did it at 20pt, but had to reduce it because the GIMP version was too large to display full-size in GitHub's comment layout.

> * In Qt you can use the actual font name - i.e. Work Sans Light where as with IM you need to use the IM name - i.e. Work-Sans-Light

\*nod\* I suspect ImageMagick's replacement of spaces with dashes in font names is a concession to it being a command line tool first-and-foremost.

> * Qt seems to handle transparncy the wrong way round from standard HTML - so its #argb instead of #rgba - see here: https://doc.qt.io/qt-5/qcolor.html#setNamedColor - so if you want to set a transparent background you need to set it as #00000000 - I think TextReader/ImageMagick allow "none".

Yeah, that's kind of a mess. (Though, for transparent you can just use the name `transparent`, which also exists in CSS.)

`#RRGGBBAA` wasn't added to CSS until _very_ late in the game. [CSS-COLOR-4](https://drafts.csswg.org/css-color-4/#numeric-rgb) finally added it, but it's not even in CSS3; until just a couple of years ago you had to use `rgba(255, 0, 0, 50%)` or `rgba(255, 0, 0, 0.5)` to get half-transparent red, `#ff000080` and `#80ff0000` were _both_ errors.

QColor's "#AARRGGBB" notaton [was added 6 years ago](https://gitlab.com/pteam/pteam-qtbase/commit/9b021a1fbd0b974a452a6e18a8c0b4b97cbf8edb) and apparently piggybacks off the same notation already being used in QtQuick. Unfortunately for them (and for us!), HTML/CSS broke the other way. (And `QColor::setNamedColor()` doesn't accept any form of CSS `rgba()` notation, just to drive the knife home.)

I suspect next year's Qt 6 will fix a lot of this, and at a minimum standardize on CSS-COLOR-4 which means `#RRGGBBAA` and `rgb(255 0 0 / 0.5)`. (Yep, they decided to lose the commas and make the alpha a divisor, since it modifies the previous three. You'll notice GitHub doesn't even recognize that notation yet, same as `#RRGGBBAA`. Chrome understands both, though!)

For the moment, dealing with alpha-blended color still sucks, and it might actually be better to accept transparency level as a separate argument from color value. You can use `QColor("#rrggbb").setAlphaF(trans);` to get the color `#rrggbb` with the alpha defined by `trans` as a float in the range `0.0` – `1.0`, or `.setAlpha(trans)` as an integer in the range `0` – `255`.

ferdnyc:
> I just tested `QGuiApplication` which worked fine but with `QCoreApplication` I get an error trying to load fonts.

Yeah, I later noticed that QTextDocument was a member of the QtGui module, which I guess makes sense. So QCoreApplication wouldn't cut it.

ferdnyc:
> I'll keep looking in to the gravity issue. One problem with the QtTextDocument is it only supports a subset of HTML and CSS (https://doc.qt.io/qt-5/richtext-html-subset.html), it's quite restrictive compared to what you might be used to with a modern web browser. 

> For example padding only seems to work on table cells. 

Heh, and I bet only because they translate it directly into `cellpadding`.

> It also doesn't look like `position` is supported so absolute positioning probably isn't possible. I think users would need to get creative with how they use it.

Mmm, so I see, and no `auto` values for `margin`, either. Even worse, "Note that the `float` property is only supported for tables and images." So it looks like it's back to 1995 with layout done in `table`s instead of `div`s.

ferdnyc:
^ But, actually, now that I think about it that makes the case even better for separating gravity from text alignment.

Text-alignment _is_ a property supported by the restrictive Qt Rich Text, so the user already has control over that without gravity. But placement comes into play when the text is rendered in an area _smaller_ than the output frame size. (If I'm interpreting the docs correctly, it will be by default — assuming the text fits in the frame — since a fixed output size isn't a requirement for the class. In fact it has `QTextDocument::adjustSize()` to fit the output size to the contents.) 

You're already checking `documentSize().height()` to offset the painter's y-location by the height, for bottom-gravity positioning. It should work to look up the width with `documentSize().width()` and offset the x-location the same way, for right-hand gravity. With an `adjustSize()` possibly needed first, to compute the correct width for the current contents.

And given that my `<div style="position:absolute;">` trick won't work for lower-right-corner placement, setting `GRAVITY_BOTTOM_RIGHT` would be the easiest way to render a block of smaller-than-full-frame, left-justified text in that corner of the frame.

(The other reason to not get text-alignment mixed up with gravity, maybe, is for RTL languages.)

jonoomph:
I'll try and take a look at this today! @jeffski this looks great though on the surface! I've also thought about implementing an HTML reader, but you beat me to it! No complains from me! haha. On a side note, I also think it would be great to support CSS and SVG animations in the future, powered by an HTML renderer (although I'm not sure Qt's engine would work for that). Anyway, I'll take a look at this very soon.

ferdnyc:
I just committed an updated `src/examples/ExampleHtml.cpp` with a QGuiApplication shell, similar to our other examples it creates an MP4 video using a QtHtmlReader to supply the frames.

include/QtHtmlReader.h
```diff
@@ -0,0 +1,144 @@
...
+
+		/// @brief Constructor for QtHtmlReader with all parameters.
+		/// @param width The width of the requested openshot::Frame (not the size of the text)
+		/// @param height The height of the requested openshot::Frame (not the size of the text)
+		/// @param x_offset The number of pixels to offset the text on the X axis (horizontal)
+		/// @param y_offset The number of pixels to offset the text on the Y axis (vertical)
+		/// @param gravity The alignment / gravity of the text
+		/// @param html The html you want to render / display
+		/// @param background_color The background color of the text (also supports Transparent)
+		QtHtmlReader(int width, int height, int x_offset, int y_offset, GravityType gravity, std::string html, std::string css, std::string background_color);
```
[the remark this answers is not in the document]
jeffski: Done

include/QtHtmlReader.h
```diff
@@ -0,0 +1,144 @@
...
+		QtHtmlReader();
+
+		/// @brief Constructor for QtHtmlReader with all parameters.
+		/// @param width The width of the requested openshot::Frame (not the size of the text)
+		/// @param height The height of the requested openshot::Frame (not the size of the text)
+		/// @param x_offset The number of pixels to offset the text on the X axis (horizontal)
+		/// @param y_offset The number of pixels to offset the text on the Y axis (vertical)
+		/// @param gravity The alignment / gravity of the text
+		/// @param html The html you want to render / display
+		/// @param background_color The background color of the text (also supports Transparent)
```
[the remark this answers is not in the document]
jeffski: Done

src/QtHtmlReader.cpp
```diff
@@ -0,0 +1,259 @@
...
+ 		// Draw image
+		text_document.drawContents(&painter);
+
+		painter.end();
+
+		// Update image properties
+		info.has_audio = false;
+		info.has_video = true;
+		info.file_size = 0;
+		info.vcodec = "Constant image uniform color";
```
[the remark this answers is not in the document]
jeffski: Changed to QImage

src/QtHtmlReader.cpp
```diff
@@ -0,0 +1,259 @@
...
+// Close reader
+void QtHtmlReader::Close()
+{
+	// Close all objects, if reader is 'open'
+	if (is_open)
+	{
+		// Mark as "closed"
+		is_open = false;
+	}
+}
```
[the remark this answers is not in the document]
jeffski: Done - copied from QImageReader, please double check.

src/QtTextReader.cpp
```diff
@@ -0,0 +1,271 @@
...
+// Close reader
+void QtTextReader::Close()
+{
+	// Close all objects, if reader is 'open'
+	if (is_open)
+	{
+		// Mark as "closed"
+		is_open = false;
+	}
+}
```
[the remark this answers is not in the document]
jeffski: As above

ferdnyc:
Python and Ruby bindings added for the new readers, and a `src/examples/ExampleHtml.py` added that does the same thing in Python that `src/examples/ExampleHtml.cpp` does in C++.

ferdnyc:
The Python example can be run against an uninstalled build of libopenshot like so, after building the library and bindings in `/path/to/builddir`:
```sh
cd /path/to/cloned-repo/src/examples
PYTHONPATH=/path/to/builddir/src/bindings/python ./ExampleHtml.py
# A relative path like PYTHONPATH=../../build/src/bindings/python/ will also work.
```

(Once again, if necessary change the `/var/tmp/` path for the output file before running.)

ferdnyc:
Pushed one more commit to correct my `SetJson()` memory leak (see #273), which snuck into the new readers.

jeffski:
I've spent a bit of time playing about with the HTML to try and align elements to the 9 positions that gravity would support.

CSS positioning isn't supported and it seems that with tables you can't set the height of rows or cells so you can't set the height to be 100% (or us pixels) and then vertical align to bottom or center.

I was able to use a spacer image to force the table cell height by setting it's width to 0 and height in pixels.

Horizontal alignment works fine so gravity in those directions isn't needed.

The advantage I see with being able to set the gravity vertically (instead of setting the height and offset of the actual QImage) is that I can set it to bottom and then the area will scale upwards without having to set a height, for example if I change font size or if text wraps on to a new line.

Perhaps instead of calling it gravity it should be `vertical_alignment` with a top, center|middle and bottom option?

I'd appreciate everyone's thoughts or if you can think of a better way to align/position elements to all 9 gravity points.

ferdnyc:
@jeffski Apologies, I know I owe you feedback on your last comment, I still haven't had time to get to it. Hopefully in the next day or so.

ferdnyc:
> I've spent a bit of time playing about with the HTML to try and align elements to the 9 positions that gravity would support.

Yeah, in the HTML/CSS it doesn't work at all.

> The advantage I see with being able to set the gravity vertically (instead of setting the height and offset of the actual QImage) is that I can set it to bottom and then the area will scale upwards without having to set a height, for example if I change font size or if text wraps on to a new line.

Agreed, and IMHO that's the same advantage being able to set horizontal gravity: If your text is narrower than the screen, and you want it placed at the right side of the screen, you can do that. Even if it's multi-line. Without having to have the text right-justified. (Again, think of multi-line text.)

> Horizontal alignment works fine so gravity in those directions isn't needed.

I'd still disagree with that, personally. Horizontal alignment works fine, so there's no reason to _override_ it with gravity. That way, you can use horizontal alignment in the HTML/CSS for its real purpose (_aligning_ text), and use gravity to align the _block_ of text. I believe I've cracked that.

Here are the bits that matter from a modified `ExampleHtml.cpp` I used:
```cpp
std::string html_code = R"html(
    <p><span id="red">Check out</span>, for your viewing pleasure:<ul>
    <li>Paragraphs!</li>
    <li>Spans!</li>
    <li>Lists!</li>
    </ul></p>
    <p id="second">(Aligned as we please!)</p>
    <p id="third">It's pretty great.</p>)html";

std::string css_code = R"css(
    * {font-family:sans-serif; font-size:15pt; color:#ffffff; background-color:transparent;}
    #red {color: #ff0000;}
    p#second {text-align:center;}
    p#third {text-align:right;})css";

// Create a reader to generate an openshot::Frame containing text
QtHtmlReader r(720, // width
               480, // height
               15, // x_offset
               15, // y_offset
               GRAVITY_BOTTOM_RIGHT, // gravity
               html_code, // html
               css_code, // css
               "#888888" // background_color
               );
```
(The transparent `background-color` in the CSS is unnecessary; it's the default, and setting it to any other color doesn't really look great anyway. I just accidentally left that in there, sorry.)

Anyway, that code with a modified `QtImageReader.cpp` produced this frame:
image

My `QtHtmlReader.cpp` code is currently a _disaster_ of debug `std::cout`s and the like, but basically the way I did it was as I'd suggested a few comments back: Get the `text_document.idealWidth()` and use it as a `td_width`, along with your existing `td_height` calculation. Then I actually go in and _set_ the document PAGE size to `td_width` × `td_height`, along with setting the document margins to 0 as well. (Otherwise, setting page size to `idealWidth()` will cause the widest line of text to wrap.)

So,
```cpp
//create the HTML/CSS document page
text_document.setDefaultStyleSheet(css.c_str());
text_document.setHtml(html.c_str());

qreal td_height = td_layout->documentSize().height();
qreal td_width = text_document.idealWidth();

text_document.setPageSize(QSizeF{td_width, td_height});
text_document.setDocumentMargin(0);
```

After that, the horizontal translations are basically the same as the vertical: on the right side they're based on `(width - td_width)`, in the center `(width - td_width) / 2`. (In my code I used the offsets to shift the text away from the gravity direction, so a positive `x_offset` with right-hand gravity was _subtracted_ from the horizontal translation. Not sure if that's the norm for libopenshot — and if not, we probably shouldn't break with the norm — it just made sense to me at the time.)

You'll also notice that the list bullets ignore my text color and remain black. That appears to be a bug in the `QTextDocument` CSS handling, even if I set `li { color: #ffffff; }` they'd still be black. So, bulleted (or numbered) lists might not be usable in `QtHtmlReader`.

ferdnyc:
There also _appears_ to be a little bit of an issue with transparency, in that a _fully_ transparent background breaks... well, everything, with the text rendering. It appears to lose all of its antialiasing, not _entirely_ surprisingly.

Like, here's the same code as my previous comment, with the `background_color` set to `"black"`:

normal_black

And with it set to `"transparent"`:
normal_transparent

That can be worked around by setting a `background_color` for the text, even a part-transparent one. Here's the same thing (`background_color = "transparent"`), but with this CSS:
```css
* {font-family:sans-serif; font-size:15pt; color:#ffffff; background-color:#40000000;}
#red {color: #ff0000;}
```
cppHtmlExample

And with this CSS:
```css
* {font-family:sans-serif; font-size:15pt; color:#80ffffff; background-color:#40000000;}
#red {color: #80ff0000;}
```
cppHtmlExample

But we'll have to see how that looks superimposed over an image/video.

src/QtHtmlReader.cpp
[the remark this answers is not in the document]
ferdnyc: Seems to work, in testing.

src/QtHtmlReader.cpp
```diff
@@ -0,0 +1,267 @@
...
+		std::shared_ptr<Frame> image_frame(new Frame(requested_frame, image->size().width(), image->size().height(), background_color, 0, 2));
+
+		// Add Image data to frame
+		image_frame->AddImage(image);
+
+		// return frame object
+		return image_frame;
+	} else {
+		// return empty frame
+		std::shared_ptr<Frame> image_frame(new Frame(1, 640, 480, background_color, 0, 2));
```
ferdnyc: You don't have video (you'll never have video), seems better to just use the `Frame()` constructor without those arguments:
```cpp
std::shared_ptr<Frame> image_frame(new Frame(1, 640, 480, background_color));
```

src/QtHtmlReader.cpp
```diff
@@ -0,0 +1,267 @@
...
+	}
+}
+
+// Get an openshot::Frame object for a specific frame number of this reader.
+std::shared_ptr<Frame> QtHtmlReader::GetFrame(int64_t requested_frame)
+{
+	if (image)
+	{
+		// Create or get frame object
+		std::shared_ptr<Frame> image_frame(new Frame(requested_frame, image->size().width(), image->size().height(), background_color, 0, 2));
```
ferdnyc: Same here, this can be:
```cpp
std::shared_ptr<Frame> image_frame(new Frame(requested_frame, image->size().width(),
                                             image->size().height(), background_color));
```

src/QtHtmlReader.cpp
```diff
@@ -0,0 +1,267 @@
...
+		info.pixel_ratio.den = 1;
+		info.duration = 60 * 60 * 24; // 24 hour duration
+		info.fps.num = 30;
+		info.fps.den = 1;
+		info.video_timebase.num = 1;
+		info.video_timebase.den = 30;
+		info.video_length = round(info.duration * info.fps.ToDouble());
+
+		// Calculate the DAR (display aspect ratio)
+		Fraction size(info.width * info.pixel_ratio.num, info.height * info.pixel_ratio.den);
```
ferdnyc: There isn't _technically_ any reason to multiply `width` and `height` by 1 here (you know it'll always be 1, it's explicitly set above), but I suppose the compiler will optimize it out so there's no real harm.

ferdnyc:
> > The advantage I see with being able to set the gravity vertically (instead of setting the height and offset of the actual QImage) is that I can set it to bottom and then the area will scale upwards without having to set a height, for example if I change font size or if text wraps on to a new line.
> 
> Agreed, and IMHO that's the same advantage being able to set horizontal gravity: If your text is narrower than the screen, and you want it placed at the right side of the screen, you can do that. Even if it's multi-line. Without having to have the text right-justified. (Again, think of multi-line text.)
> 
> > Horizontal alignment works fine so gravity in those directions isn't needed.
> 
> I'd still disagree with that, personally. Horizontal alignment works fine, so there's no reason to _override_ it with gravity. That way, you can use horizontal alignment in the HTML/CSS for its real purpose (_aligning_ text), and use gravity to align the _block_ of text. I believe I've cracked that.

Actually, I'm thinking now that a hybrid approach might be best. For the _center_ gravities (all three of them), don't do the `idealWidth()` reduction, and leave the text set to full-width — that way, the user can use left- and right-aligned text across the full width of the output frame. So, a user who wants their right-justified text to be all the way at the top right corner of the screen can still do that by setting `GRAVITY_TOP` and right-aligning in the HTML. (Otherwise, the text would always be centered no matter how they aligned it, if the `idealWidth()` was narrower than the frame.)

jeffski:
My approach to left and right alignment of blocks of text was to use tables. I'm not at my desk right now but I think I was able to align a table right or a table within a table aligned right. Then the text can still be aligned left. It does require a bit of extra markup and might not be an obvious solution but it was possible.

I'm not sure if this is better than using the gravity solution. But I agree we need to be able to align a block of text right but align the text left.

jeffski:
> There also _appears_ to be a little bit of an issue with transparency

I also noticed this, as I use transparent backgrounds and text quite a bit. I noticed that some colors are worse than others too, red is particularly bad while black and white are OK. I'm not sure if some fonts and font sizes are better than others as well.

I have never noticed it before so I compared it using the ImageMagick TextReader and I am pretty sure the results are the same. I'm not sure if this is the text being rendered as an image or when it is encoded using FFMpeg. Also not sure, but I think the text is generated by the operating systems font rendering engine, FreeType for Linux, so ImageMagick and Qt would both create the same result. It would be great if someone could confirm this.

ferdnyc:
Without `width: auto` I'm not entirely sure how we'd _implement_ it offhand (unless some form of auto-width works in table cells? They seem to have special rules in a lot of cases), but I think it could even be workable to have wrapper HTML that implements horizontal gravity, and just surround the user's content with that. Qt's "rich content" pseudo-HTML is limited enough that anything the user can create would probably exist happily inside a `<td>...</td>` tag without breaking in any significant way..

Btw, this is all just me meddling / brainstorming, ultimately it's your feature and I'd have no objections to merging even if you decide to go with the alignment-based gravity. I'll rarely shut up until told to (character flaw), but I'm happy to be told to when it comes to that. 😉

Butt for the record, I think this is great as-is, and I have zero reservations about merging it whenever you / @jonoomph / anyone else feels it's ready.

src/QtHtmlReader.cpp
ferdnyc: Formerly nonsense comment above this one now corrected to be non-nonsense. ("Sense" might still be overstating it.)

ferdnyc:
Re: transparency, I suspect it has to do very directly with the literal absence of any background color. Font antialiasing, especially in the world of subpixel hinting, works not only with the outline of the font glyphs, but with the background they're placed against. The same font in the same color will be rendered differently against a black background vs. a red background vs. a blue background. The differences are just subtle tweaks around the edges, but they make all the difference. When there's _no_ color there, decisions are made that won't be ideal.

One possibility might be to specify the _color_ of the transparent pixels, instead of leaving them literally transparent. (In most image formats that support transparency, transparent pixels can still have a color associated with them.) I'll play with that some tomorrow.

jeffski:
> Butt for the record, I think this is great as-is, and I have zero reservations about merging it whenever you / @jonoomph / anyone else feels it's ready.

I'm happy with it too, so long as @jonoomph is happy, be great to see it merged.

jeffski:
> Re: transparency

I had a bit of a play around. You are right, setting a background color makes a very subtle difference with a huge impact. Fonts are so much smoother looking that way. I did try the transparent pixels with a color but I couldn't see an improvement.

Not sure what the options are with this, my understanding is that if you want to layer the text over an image or video then the image and text are not aware of each other and created as separate images/frames, the text has no way to know what colors are in the layer below it and its background needs to be transparent.

Anyway, I guess that is a separate issue.

ferdnyc:
> > Re: transparency
> 
> I had a bit of a play around. You are right, setting a background color makes a very subtle difference with a huge impact. Fonts are so much smoother looking that way. I did try the transparent pixels with a color but I couldn't see an improvement.

Yeah, that was a total bust, it has absolutely no effect. My current thinking is that we may be able to turn off subpixel hinting for the QTextDocument's font(s). I've got it on my list to fiddle with that a bit.

The _ideal_ way to handle text overlay would be to delay rendering the text until the background image is available, and then render it hinted against that background rather than rendering into a transparent field and overlaying the rendered image later. But doing it that way would require a significant redesign of the entire reader/compositor structure in libopenshot.

> Not sure what the options are with this, my understanding is that if you want to layer the text over an image or video then the image and text are not aware of each other and created as separate images/frames, the text has no way to know what colors are in the layer below it and its background needs to be transparent.

\*nod\* And that's exactly the issue — at its core, the "problem" here is that we _want_ to render text without knowing its placement against the background, which makes subpixel hinting an impossibility because it can't work that way. (That's OK, because subpixel hinting shouldn't be used anyway. It's only appropriate when the text is being rendered for screen display — and export to a video file is _**not**_ screen display. It's impossible to pre-render (subpixel-hinted) text unless you know what screen the video will ultimately be displayed on. It may have a different pixel arrangement or aspect, it may even be a CRT and have no subpixel grid at all.)

Which means that, for our current use case, subpixel hinting is working _against_ us. Rendering using grayscale hinting may produce better results than just switching antialiasing off _completely_, which is what it appears to do now.

jeffski:
I've just pushed a couple of updates:

1. I was trying to use Roboto Black Italic font and while you can just specify the font name with ImageMagick with Qt you need to specify the font and the style - i.e. Roboto Black and style Italic. I have added parameters that let you set text to bold or  italic to handle these kinds of fonts and styles.

2. The existing TextReader has a feature that lets you set the background colour behind text, it's useful for captions/subtitles. This was missing from the QtTextReader so I added it.

ferdnyc:
> 1. I was trying to use Roboto Black Italic font and while you can just specify the font name with ImageMagick with Qt you need to specify the font and the style - i.e. Roboto Black and style Italic. I have added parameters that let you set text to bold or  italic to handle these kinds of fonts and styles.

A thought: Would it be easier to just take a QFont as a single parameter? You're going to end up setting it up into one anyway, correct?

jeffski:
> A thought: Would it be easier to just take a QFont as a single parameter? You're going to end up setting it up into one anyway, correct?

It would make sense as there are a lot more font options available that could be used. I had a quick go at setting it up but I'm not sure how `QtTextReader::JsonValue()` and `QtTextReader::SetJsonValue(Json::Value root)` would work.

I am getting errors when compiling if this helps:
```
/path/libopenshot/src/QtTextReader.cpp: In member function ‘virtual Json::Value openshot::QtTextReader::JsonValue()’:
/path/libopenshot/src/QtTextReader.cpp:217:17: error: no match for ‘operator=’ (operand types are ‘Json::Value’ and ‘QFont’)
  root["font"] = font;
                 ^~~~
In file included from /usr/local/include/json/reader.h:11:0,
...

/usr/local/include/json/value.h:352:10: note: candidate: Json::Value& Json::Value::operator=(Json::Value)
   Value& operator=(Value other);
          ^~~~~~~~
/usr/local/include/json/value.h:352:10: note:   no known conversion for argument 1 from ‘QFont’ to ‘Json::Value’
/path/libopenshot/src/QtTextReader.cpp: In member function ‘virtual void openshot::QtTextReader::SetJsonValue(Json::Value)’:
/path/libopenshot/src/QtTextReader.cpp:275:21: error: no match for ‘operator=’ (operand types are ‘QFont’ and ‘Json::Value’)
   font = root["font"];
                     ^
In file included from /usr/include/x86_64-linux-gnu/qt5/QtGui/qfontinfo.h:44:0,
....

/usr/include/x86_64-linux-gnu/qt5/QtGui/qfont.h:241:12: note: candidate: QFont& QFont::operator=(const QFont&)
     QFont &operator=(const QFont &);
            ^~~~~~~~
/usr/include/x86_64-linux-gnu/qt5/QtGui/qfont.h:241:12: note:   no known conversion for argument 1 from ‘Json::Value’ to ‘const QFont&’
/usr/include/x86_64-linux-gnu/qt5/QtGui/qfont.h:248:19: note: candidate: QFont& QFont::operator=(QFont&&)
     inline QFont &operator=(QFont &&other) Q_DECL_NOEXCEPT
                   ^~~~~~~~
/usr/include/x86_64-linux-gnu/qt5/QtGui/qfont.h:248:19: note:   no known conversion for argument 1 from ‘Json::Value’ to ‘QFont&&’
```

ferdnyc:
> It would make sense as there are a lot more font options available that could be used. I had a quick go at setting it up but I'm not sure how `QtTextReader::JsonValue()` and `QtTextReader::SetJsonValue(Json::Value root)` would work.

Fnard. I didn't even think about the JSON. Yeah, unless there's some way to serialize a QFont usefully that's probably a no-go.

Oh! But there is. [Qfont::toString](https://doc.qt.io/qt-5/qfont.html#toString)
> Returns a description of the font. The description is a comma-separated list of the attributes, perfectly suited for use in QSettings.
> 
> See also fromString().

ferdnyc:
Since `QFont::toString` returns a `QString`, you'd actually want to use `font.toString().toStdString()` if the serialization can accept `std::string`, or if it really needs a `char *`, then (ugh) `font.toString().toStdString().c_str()`.

ferdnyc:
Reading it back in, you'd want to set it via either:
```c++
font = new QFont(QString::fromStdString(root["font"]));
// or
font = new QFont(QString::fromUtf8(root["font"]));
```
Depending, again, whether `std::string` is a kosher type coercion of `Json::Value`.

ferdnyc:
Wait, nope, sorry. Forgot about `fromString()`. So:
```c++
font.fromString(QString::fromStdString(root["font"]));
// or
font.fromString(QString::fromUtf8(root["font"]));
```
Assuming `font` is already allocated as a QFont.

jeffski:
Great, the first on worked for `QtTextReader::JsonValue()`.

The next next on for `QtTextReader::SetJsonValue(Json::Value root)` I updated the code to:

```
if (!root["font"].isNull())
	font = font.fromString(QString::fromStdString(root["font"]));
```
But I get the following error:

```
/path/libopenshot/src/QtTextReader.cpp: In member function ‘virtual void openshot::QtTextReader::SetJsonValue(Json::Value)’:
/path/libopenshot/src/QtTextReader.cpp:278:61: error: no matching function for call to ‘QString::fromStdString(Json::Value&)’
   font = font.fromString(QString::fromStdString(root["font"]));
                                                             ^
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qcoreapplication.h:44:0,
...
/usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:1321:16: note: candidate: static QString QString::fromStdString(const string&)
 inline QString QString::fromStdString(const std::string &s)
                ^~~~~~~
/usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:1321:16: note:   no known conversion for argument 1 from ‘Json::Value’ to ‘const string& {aka const std::__cxx11::basic_string<char>&}’
```

Any ideas?

ferdnyc:
> Any ideas?

Aha, yep! `Json::Value::asString()`. Also, you wouldn't want to do `font = font...` — if `font` is already allocated as a QFont, then just `font.fromStdString(...)` without the assignment. Otherwise, you'd need to allocate a new QFont first, _then_ call `font.fromString()`:
```c++
font = new QFont();
font.fromString(QString::fromStdString(root["font"].asString()));
```
...But I think it's already allocated.

ferdnyc:
`font.fromStdString()` returns bool, presumably indicating whether it was successful. So I suppose the Proper™ thing to do would be to check the return value, and raise the alarm if it's false.

jeffski:
Thanks for your awesome help. I've got it working using QFont which opens up a whole new set of font possibilities. Pushed an update just now. I wasn't quite sure how to handle the `fromFontString` returning a bool though.

ferdnyc:
Yeah, I have to look at that. I think there's an exception you can raise. I'll dig around in the other classes a bit.

ferdnyc:
Yeah, if `QFont::fromString()` returns `false`, I _suppose_  it wouldn't be completely wrong to throw an `InvalidJSON` exception, even though it's not technically the _JSON_ that's invalid per se:

https://github.com/OpenShot/libopenshot/blob/c685571e6388ad5cc6c40661fd51bd15b436ccac/src/Timeline.cpp#L1135-L1142

The second argument is a file path, and it's required (at least until #315 merges), so you can just pass an empty string same as above.

ferdnyc:
@jeffski So I was gearing up to make a renewed attempt at pressuring @jonoomph into reviewing these contributor PRs (as there are only 3 or 4 in a ready-to-merge state), but then I got to thinking.

Now that QtTextReader accepts a QFont, does it make sense that QtHtmlReader should do the same? I know in theory you have the CSS to style the content, so it's not strictly REQUIRED, but as we've seen Qt HTML is very limited, and as you pointed out having access to QFont opens up a whole host of configuration possibilities.

jeffski:
> Now that QtTextReader accepts a QFont, does it make sense that QtHtmlReader should do the same? I know in theory you have the CSS to style the content, so it's not strictly REQUIRED, but as we've seen Qt HTML is very limited, and as you pointed out having access to QFont opens up a whole host of configuration possibilities.

I could certainly do this - replace the html and css parameters with a [QTextDocument](https://doc.qt.io/qt-5/qtextdocument.html). I still think you would still have the same HTML/CSS limitations however and I'm not sure how useful some of the other QtTextDocument features would be. But I'm just think about my own use case and, others might find them useful.

Let me know and I can update.

ferdnyc:
> > Now that QtTextReader accepts a QFont, does it make sense that QtHtmlReader should do the same? I know in theory you have the CSS to style the content, so it's not strictly REQUIRED, but as we've seen Qt HTML is very limited, and as you pointed out having access to QFont opens up a whole host of configuration possibilities.
> 
> I could certainly do this - replace the html and css parameters with a [QTextDocument](https://doc.qt.io/qt-5/qtextdocument.html). I still think you would still have the same HTML/CSS limitations however and I'm not sure how useful some of the other QtTextDocument features would be. But I'm just think about my own use case and, others might find them useful.

Oh, sorry, no... that's not what I was suggesting at all. (It's an interesting thought in its own right, but I suspect too complex for most users. Callers linking with libopenshot directly, sure, but for incorporating into the OpenShot interface, I think there's a huge advantage to exposing the "document" parameters as easily-serializable types like HTML and CSS (both just text strings).

No, what I was suggesting was an **additional** parameter (or setter), that would accept a `QFont` and pass it to `QTextDocument::setDefaultFont()`, to configure the document's font rendering.

There are certain things you can do with a QFont that you can't do in CSS — things like switching `kerning` on/off, or setting the `hintingPreference` — and some others that are just _easier_ than with CSS, like using a font chooser to select the font graphically instead of having to worry about getting the CSS parameters right. And as we've seen, QFont is easily serializable, so it doesn't complicate the interface much to add it in.

And, of course, anyone who doesn't _want_ to use the QFont can always just **not**, and continue to work exclusively in HTML and CSS if that's what they prefer.

That was my thought, as far as I can remember. (It was a while ago!) :wink:

jeffski:
Right, I think I understand now. I think the way it is now does make it a bit easier to use, just pass in the HTML and CSS. But, yes, the QFont could also be passed in and and looks like it can be set using [setDefaultFont](https://doc.qt.io/qt-5/qtextdocument.html#setDefaultFont). I'll have to give it a quick test and make sure it will work.

jonoomph:
Looks like some conflicts on this one now. @jeffski  please let me know when this one is ready to merge. Thx!

ferdnyc:
I've resolved the conflicts here, which were only in `src/CMakeLists.txt` due to the other build system changes that have been merged.

src/QtHtmlReader.cpp
ferdnyc:
```suggestion
		std::shared_ptr<Frame> image_frame(new Frame(1, 640, 480, background_color));
```

src/QtHtmlReader.cpp
ferdnyc:
```suggestion
		std::shared_ptr<Frame> image_frame(new Frame(requested_frame, image->size().width(), image->size().height(), background_color));
```

src/QtTextReader.cpp
```diff
@@ -0,0 +1,290 @@
...
+		// Update image properties
+		info.has_audio = false;
+		info.has_video = true;
+		info.file_size = 0;
+		info.vcodec = "QImage";
+		info.width = width;
+		info.height = height;
+		info.pixel_ratio.num = 1;
+		info.pixel_ratio.den = 1;
+		info.duration = 60 * 60 * 24; // 24 hour duration
```
ferdnyc:
```suggestion
		info.duration = 60 * 60 * 1;  // 1 hour duration
```

src/QtHtmlReader.cpp
```diff
@@ -0,0 +1,267 @@
...
+		// Update image properties
+		info.has_audio = false;
+		info.has_video = true;
+		info.file_size = 0;
+		info.vcodec = "QImage";
+		info.width = width;
+		info.height = height;
+		info.pixel_ratio.num = 1;
+		info.pixel_ratio.den = 1;
+		info.duration = 60 * 60 * 24; // 24 hour duration
```
ferdnyc:
```suggestion
		info.duration = 60 * 60 * 1;  // 1 hour duration
```

ferdnyc: (To match the band-aid I added to #346.)

src/QtTextReader.cpp
ferdnyc: (To match the band-aid I added to #346.)

src/QtHtmlReader.cpp
jonoomph: :+1:

src/QtTextReader.cpp
jonoomph: :+1:

jonoomph:
@jeffski , once you commit the 2 suggestions above, I'll go ahead and get this merged. Thx again for this PR. Looks great! Hopefully it will work on all 3 OSes. :crossed_fingers:

jeffski:
Interesting Github feature - I was able to commit the suggestions directly from this page. Should be ready to go!!

ferdnyc:
> Interesting Github feature - I was able to commit the suggestions directly from this page.

Indeed. It's handy, modulo the fact that suggestions can currently only be attached to a single line of code. That limits their ability to be used for anything substantial, but for quick-fix edits like the ones I left here it works.

Early next year they plan to finally introduce multi-line suggestions (building on the already-available new multi-line code comments feature), at which point it'll _really_ be useful.

ferdnyc:
Travis is done and @jonoomph indicated intent-to-merge, so I'll press the big green button. Thanks @jeffski !

jonoomph:
@ferdnyc @jeffski Looks like this borked our build servers... oops:

```
[ 55%] Building CXX object src/CMakeFiles/openshot.dir/QtTextReader.cpp.o
cd /home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/build/src && /usr/bin/c++   -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -DOPENSHOT_IMAGEMAGICK_COMPATIBILITY=0 -DQT_CORE_LIB -DQT_GUI_LIB -DQT_MULTIMEDIAWIDGETS_LIB -DQT_MULTIMEDIA_LIB -DQT_NETWORK_LIB -DQT_NO_DEBUG -DQT_WIDGETS_LIB -DUSE_IMAGEMAGICK=1 -DUSE_RESVG=1 -Dopenshot_EXPORTS -I/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/include -I/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/include/effects -I/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/include/Qt -I/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/build/include -I/usr/include/ImageMagick -I/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/build/install-x64/include/libopenshot-audio -isystem /usr/include/qt5 -isystem /usr/include/qt5/QtWidgets -isystem /usr/include/qt5/QtGui -isystem /usr/include/qt5/QtCore -isystem /usr/lib/x86_64-linux-gnu/qt5/mkspecs/linux-g++-64 -isystem /usr/include/qt5/QtMultimedia -isystem /usr/include/qt5/QtNetwork -isystem /usr/include/qt5/QtMultimediaWidgets -I/usr/local/include/resvg -I/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/../thirdparty/jsoncpp -isystem /usr/include/x86_64-linux-gnu  -fPIE -fPIE -fPIE -fPIE -fPIE -O3 -DNDEBUG -fPIC   -fopenmp -std=c++11 -o CMakeFiles/openshot.dir/QtTextReader.cpp.o -c /home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/QtTextReader.cpp
In file included from /home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/QtTextReader.cpp:33:0:
/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/../include/QtTextReader.h:149:16: error: ‘string’ has not been declared
   void SetJson(string value); ///< Load JSON string into this object
                ^
/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/QtTextReader.cpp:55:43: error: variable or field ‘SetTextBackgroundColor’ declared void
 void QtTextReader::SetTextBackgroundColor(string color) {
                                           ^
/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/QtTextReader.cpp:55:43: error: ‘string’ was not declared in this scope
/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/QtTextReader.cpp:55:43: note: suggested alternative:
In file included from /usr/include/c++/4.8/iosfwd:39:0,
                 from /usr/include/c++/4.8/ios:38,
                 from /usr/include/c++/4.8/ostream:38,
                 from /usr/include/c++/4.8/iostream:39,
                 from /home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/../include/ReaderBase.h:34,
                 from /home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/../include/QtTextReader.h:36,
                 from /home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/QtTextReader.cpp:33:
/usr/include/c++/4.8/bits/stringfwd.h:62:33: note:   ‘std::string’
   typedef basic_string<char>    string;   
                                 ^
/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/QtTextReader.cpp:228:6: error: prototype for ‘void openshot::QtTextReader::SetJson(std::string)’ does not match any in class ‘openshot::QtTextReader’
 void QtTextReader::SetJson(std::string value) {
      ^
In file included from /home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/QtTextReader.cpp:33:0:
/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/src/../include/QtTextReader.h:149:8: error: candidate is: void openshot::QtTextReader::SetJson(int)
   void SetJson(string value); ///< Load JSON string into this object
        ^
make[2]: *** [src/CMakeFiles/openshot.dir/QtTextReader.cpp.o] Error 1
make[2]: Leaving directory `/home/gitlab-runner/builds/5cd61c66/0/OpenShot/libopenshot/build'
make[1]: *** [src/CMakeFiles/openshot.dir/all] Error 2
```

jonoomph:
Oh, it looks like this is already fixed in another branch? And... my Mac builder decided to disconnect WiFi, cool. Okay, let me retrigger the builds and see if all is well now. :+1:

ferdnyc:
Yeah, that was the branch for PR #296, and I've already resolved the issues there.

jonoomph:
Thanks! And... Mac builder is now successfully connected with Ethernet!

PR #306, opened 2019-08-11
Bindings: Apply #defines for optional libs

ferdnyc:
The resetting of `CMAKE_SWIG_FLAGS` to the empty string, in the bindings' CMakeLists.txt files, was clearing away the definitions for things like `USE_IMAGEMAGICK` and `USE_BLACKMAGIC`. This caused the bindings to be built _without_ those classes even if they're available in the library.

(And it's been like that for _three and a half **years**_.)


jonoomph:
Some notes:

"Bilinear is fast but rather low quality. Best avoided unless you are looking to reduce bitrate.

Bicubic delivers sharper results than Bilinear. It is best for enlarging. When reducing it can lead to a lot of moire artifacts.

Lanczos is best for reducing. It gives sharp results with less artifacting than Bicubic. For enlarging it about the same as Bicubic."

jonoomph:
LG

PR #220, opened 2019-04-17
Hardware Support Integration (big changes to FFmpegReader and FFmpegWriter)

jonoomph:
We have been running this hardware-support branch in parallel for a while, and based on our internal testing, it works good enough to be merged into `develop`. I would love to find some additional help with maintaining some of this code, so hopefully `develop` will help make this work more visible. Thank you @eisneinechse for the hard work putting this together!

PR #307, opened 2019-08-11
WIP: Remove not used code

SuslikV:
Custom scale_mode is never used during writing.

The option intended to be used in chroma downscale during writing,
and Lanczos has no advantages over Bilinear for this task (size is
reduces twice horizontally and vertically or isn't changes at all).

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -2032,11 +2032,6 @@ void FFmpegWriter::OutputStreamInfo() {
 
 // Init a collection of software rescalers (thread safe)
 void FFmpegWriter::InitScalers(int source_width, int source_height) {
-	int scale_mode = SWS_FAST_BILINEAR;
-	if (openshot::Settings::Instance()->HIGH_QUALITY_SCALING) {
-		scale_mode = SWS_LANCZOS;
-	}
-
 	// Init software rescalers vector (many of them, one for each thread)
 	for (int x = 0; x < num_of_rescalers; x++) {
 		// Init the software scaler from FFMpeg


ferdnyc:
Yeah, some static analysis tool or other that I ran on the code also flagged `scale_mode` as never actually used, in `FFmpegWriter.cpp`. It _is_ used in `FFmpegReader.cpp`, and looks like copy-paste code might be the explanation for its existence.

Still, given that I don't know what plan, if any, there is/was for it, I'm inclined to leave this one for the authors/maintainers of that code to decide what they want to do with it, even if it's just to replace it with a `TODO` comment about future plans.

But +1 on that variable currently serving no purpose, since it's only ever assigned a value that never actually gets read anywhere.

eisneinechse:
I have to think about it. I think it was because of some speed issues with CPUs that aren't top notch. Please check that.

SuslikV:
Want to mention that the `sws_getContext` (from the code) here operates of already transformed and combined image sources. The need of this function - is because RGBA input incompatible with YUV (4:2:0) output. So, usage of scaling only to downscale chroma planes (or later, when true lossless will be implemented, just planes conversion of the same geometry).

ferdnyc:
@eisneinechse 



> I have to think about it. I think it was because of some speed issues with CPUs that aren't top notch. Please check that.

Done. Based on the history of those lines... it's actually much stranger.

`scale_mode` was introduced just a few months ago, in commit 13bd272eade3da6926cf81b5082a0a50451d4e23 (which created the Settings class). At the time, it _was_ integrated into the `sws_getContext()` call in `FFmpegWriter::InitScalers()`:

https://github.com/OpenShot/libopenshot/blob/13bd272eade3da6926cf81b5082a0a50451d4e23/src/FFmpegWriter.cpp#L1751-L1760

But then the subsequent merge of the `hardware-support` branch in #220 appears to have been reconciled in a way that partially reversed that change, unintentionally. (Or, with no explicit expression of intent, since the commit in question (7e3669b620d96af79a98892eb9b7a11e2114853e) was based before the Settings merge and knew nothing about `scale_mode`.)

So, I guess chalk one up to the dangers of merging long-lived topic branches. But I'm even _less_ clear on where that leaves things with `scale_mode`, now.

ferdnyc:
> But I'm even _less_ clear on where that leaves things with `scale_mode`, now.

(Especially since the scaling mode that's _actually_ used in `sws_getContext()`, post-`hardware-support`-merge, is `SWS_BILINEAR` which is **neither** of the `scale_mode` options.) 

@SuslikV says that `SWS_LANCZOS` is pointless for output scaling, so should `scale_mode` select between `SWS_FAST_BILINEAR` and `SWS_BILINEAR`? Or is there no real point in offering a "high-quality" output scaling option at all, when the only difference is between those two?

eisneinechse:
Sorry, I am in a stressful situation for a few weeks and might not have the right answers (right now).
As far as I can remember the high quality should be for exporting but for playback it might be necessary to have a lower quality conversion as the processor power might simply be to low for a reasonable playback. That is almost always true for 60 fps videos in high resolution. I'll try to get more time to see what it might be useful for as it also touches another theme that I would like to look at, HDR. If someone could come up with a reasonable way to include it I would be highly grateful. So far I only could explore that googleing for "ffmpeg  yuv420p10" brings the best search results.
I also see that right now basically every video editor is struggling with high resolutions combined with high frame rates. The amount of data that has to be moved around in memory is just a little high for hardware that is reasonably priced.

SuslikV:
> ...I would like to look at, HDR

Qt used for transform and rendering of the images sets number of limits for the mentioned improvements (depends on Qt version). Color handling (16 to 8 conversion) is questionable thing for such simple editor too. Number of HDR devices sold worldwide is still very limited. I don't think that this question require any attention for the next 4 years from the OpenShot.

SuslikV:
Few screens to compare the quality of the scaling during writing:

SWS_FAST_BILINEAR
<details>
SWSfastbilinearWrite02 screenshot_result

</details>

SWS_BILINEAR
<details>
SWSbilinearWrite02 screenshot_result

</details>

SWS_LANCZOS
<details>
SWSLanczosWrite02 screenshot_result

</details>

Quality: _fast_bilinear_ produces worst quality, _lanczos_ and _bilinear_ here has almost identical quality. As for me, the _fast_bilinear_ produces a bit blurred image (probably that also causes smaller file size when compress with a loss).
Speed: difference during export unnoticeable - about 00:06:30 to complete for all tests (about 0.005% of difference).
Resulting file size: _fast_bilinear_ smaller for 1.4% from both _lanczos_ and _bilinear_ (last two produces similar sizes).

Setup details: FFmpeg 3.4.2, the write during export was forced to use scaling method in custom build of OpenShot, project is 720p@30fps, 00:03:27 length, video only file, libx264, crf=23.
Input file has same resolution 1280x720 as output, _Scale_ in OpenShot was set to _None_ (otherwise it can do nasty filtering by Qt transform). I can't test bigger samples properly because of weak PC and small display.
Incorrect colors on the screenshots caused by default bt.601 output of ffmpeg and bt.709 decoding of my video players. This was mentioned in https://github.com/OpenShot/libopenshot/pull/219 (and this is actually part of it).

The comparison itself is based is on my own estimation and my own view.

eisneinechse:
@SuslikV 
4k TVs are HDR devices. The biggest improvement is HDR. I will use whatever can handle HDR. I would guess that the other question about BT.2020 is in connection with the increase in colours. Just because other don't use a thing certainly want change my decision on using it, that goes for everything including HDR.

ferdnyc:
Coming back around to this in the hopes of getting it resolved...

@SuslikV  writes: 
> Quality: _fast_bilinear_ produces worst quality, _lanczos_ and _bilinear_ here has almost identical quality. 

It's funny you say that, though. Because, at _full_ size, I definitely agree — however, I noticed that when the image is scaled down to fit GitHub's page format, it seems to me that `FAST_BILINEAR` actually looks **better** than the other two. Or at least, it scales down better in Chrome.

I'm looking primarily at this part here, in `FAST_BILINEAR`:

image

vs. in `LANCZOS` (and as you say, `BILINEAR` is nearly identical, even scaled-down):

image

So, if that holds true for video generally, there may be advantages to using `FAST_BILINEAR` on the _preview_ output beyond just a speed increase (if any) — the previews may actually look better, at reduced size. A possibility, anyway?

Anyway, that's all sort of beside the point for **this** change. Here, at least, the scaling mode is currently locked in at `BILINEAR`, and at least based on the screens provided I'm inclined to agree that there's no real advantage to enabling `LANCZOS` scaling — `BILINEAR` appears to be indistinguishable, though that can be revisited in the context of HDR and/or improved 4K support.

However, since there _may_ be some advantage to enabling `FAST_BILINEAR` even today, why don't we restore the ability for `scaling_mode` to be changed via the Settings, but have it select between `FAST_BILINEAR` and `BILINEAR` instead of between `FAST_BILINEAR` and  `LANCZOS`?

So, the code would become:
```c++
 int scale_mode = SWS_FAST_BILINEAR; 
 if (openshot::Settings::Instance()->HIGH_QUALITY_SCALING) { 
 	scale_mode = SWS_BILINEAR; 
 } 
  
 // Init software rescalers vector (many of them, one for each thread) 
 for (int x = 0; x < num_of_rescalers; x++) 
 { 
 	// Init the software scaler from FFMpeg 
 	img_convert_ctx = sws_getContext(source_width, source_height, PIX_FMT_RGBA,
 	    info.width, info.height, AV_GET_CODEC_PIXEL_FORMAT(video_st, video_st->codec),
 	    scale_mode, NULL, NULL, NULL); 
```
Meaning that, if we change the Settings switch for `HIGH_QUALITY_SCALING` so it defaults **`true`**, the scaling would be unchanged from today. But if it's turned off, `FAST_BILINEAR` would be used instead.

....Actually, on that basis I'm sort of inclined to change the setting to `FAST_SCALING`, default `false`. So maybe let's go with:
```c++
 int scale_mode = SWS_BILINEAR; 
 if (openshot::Settings::Instance()->FAST_SCALING) { 
 	scale_mode = SWS_FAST_BILINEAR; 
 } 
  
 // Init software rescalers vector (many of them, one for each thread) 
 for (int x = 0; x < num_of_rescalers; x++) 
 { 
 	// Init the software scaler from FFMpeg 
 	img_convert_ctx = sws_getContext(source_width, source_height, PIX_FMT_RGBA,
 	    info.width, info.height, AV_GET_CODEC_PIXEL_FORMAT(video_st, video_st->codec),
 	    scale_mode, NULL, NULL, NULL); 
```
Along with the appropriate changes to `Settings.cpp`.

Would that work for everyone?

SuslikV:
You shouldn't look at Chrome at all, @ferdnyc . But you can implement any phantom improvements to this code.

jonoomph:
Some notes:

"Bilinear is fast but rather low quality. Best avoided unless you are looking to reduce bitrate.

Bicubic delivers sharper results than Bilinear. It is best for enlarging. When reducing it can lead to a lot of moire artifacts.

Lanczos is best for reducing. It gives sharp results with less artifacting than Bicubic. For enlarging it about the same as Bicubic."

jonoomph:
Yes, the original intention was to allow the Settings class to change the scaling mode, so our previews could be run with the minimum amount of CPU, and our exports could look nice and sharp. However, as my previous comment mentions, there might not be 1 perfect HQ scaling mode. It might depend if you are upscaling or downscaling the source content. In that scenario, I like the idea that this could be adjusted in the Preferences or the Export dialog ("High Quality Scaling Mode" or something), but during preview, fast bilinear would be fixed.

jonoomph:
If everyone agrees, let's implement it this way for now, and see how it goes. Or I can add in the Preferences bit in a new PR if that is easier. But for sure, let's add back in the Settings control of scale mode, and default to FAST_BILINEAR for previews. :+1:

SuslikV:
Don't forget that this scaling mainly scales down color planes (twice), nothing more.

ferdnyc:
@SuslikV I don't think anyone has necessarily committed to Lanczos specifically, @jonoomph even had to go and introduce _Bicubic_ into the mix which previously nobody had even mentioned! :laughing: (But it's also clear, from his notes, we don't want to even touch that for previewing.)

But as @jonoomph said, in terms of the current _Settings_ control (which would still not be user-facing) this is mostly about restoring the ability to switch to `FAST_BILINEAR` internally for the preview renderer, something that was accidentally lost in the `hardware-support` merge.

SuslikV:
@ferdnyc It is big question what you should look at during preview of the clip, but this PR is about the saving/export/writing.

ferdnyc:
That.... is a really excelling point. <Code>FFmpeg<b>Writer</b></code> has nothing to do with previews. @jonoomph?

SuslikV:
To be honest, this PR changes absolutely nothing. There is dead code in the program. It's time to remove it. That's all. It is not new feature. Why did it faced this long conversation? Did you find something that you forget longtime ago?

jonoomph:
This PR is a reminder that we lost our Settings ability to change between fast scaling and slow scaling, needed for high performance previews, and high quality final exports. Let's go ahead and **close** this one, and I'll bring back the Settings code (which is the correct way to fix this).

SuslikV:
But in openshot-qt it works right now, automatically - fast decoding for previews, lanc decoding for export. What are talking about? Preview to export (draft)?

jonoomph:
This is now fixed in a different PR (https://github.com/OpenShot/libopenshot/pull/366), to correctly use the scale_mode variable. I'm closing this one.

PR #308, opened 2019-08-12
Disable debug logger on close

SuslikV:
Fixes https://github.com/OpenShot/openshot-qt/issues/2933

### DIFF
--- a/src/ZmqLogger.cpp
+++ b/src/ZmqLogger.cpp
@@ -160,6 +160,9 @@ void ZmqLogger::Path(string new_path)
 
 void ZmqLogger::Close()
 {
+	// Disable logger as it no longer needed
+	enabled = false;
+
 	// Close file (if already open)
 	if (log_file.is_open())
 		log_file.close();


ferdnyc:
@SuslikV Does this solve OpenShot/openshot-qt#2933 for you? 

It looks fine to me and doesn't seem to cause any problems, but as far as fixing the crash issue I wasn't able to test that since I can't reproduce the issue to begin with. On Linux OpenShot seems to exit cleanly from a debugging+previewing state, even without this change.

(It's still an improvement, though, so so I'll merge it regardless. Thanks!)

SuslikV:
>...On Linux OpenShot seems to exit cleanly from a debugging+previewing state, even without this change.

Is _Preview File_ window running after you closing main application window?

ferdnyc:
> > ...On Linux OpenShot seems to exit cleanly from a debugging+previewing state, even without this change.
> 
> Is _Preview File_ window running after you closing main application window?

Ohh, the Preview _window_ — I'd misunderstood, I thought you meant the embedded preview widget. Now I see it.

Yeah, without this change, I get a backtrace on closing the main window.

**With** this change, the Preview File window... stays on-screen and keeps running even after I close the QMainWindow, yeah. The application doesn't actually exit until I also close the Preview File window.

Not sure how I feel about _that_ behavior, but changing it would be an openshot-qt issue anyway.

SuslikV:
The window has no parent  and playback running in separate thread, so this is by design (stay alive on close). Is this issue at all?

ferdnyc:
> The window has no parent control over it and playback running in separate thread, so this is by design (stay alive on close). 

It's _Qt_ working as designed, but unless there's a comment in the source that says, "This window is intended to stay around after the main application window is closed),  I can practically guarantee that's not the reason _why_ it's set up like that:

* The preview thread is the same one that the main interface's preview widget runs in. It's in a separate thread purely for performance reasons.
* The window that shows the preview is actually the _Cutting_ window, reused without the cutting controls visible. It was never intended to be a standalone video player.

So, not OpenShot design, but rather a side effect.

> Is this issue at all?

It depends. Perhaps not. But my concerns are:

1. It's probably not what the user intends or expects in 99% of cases. It's an accident on their part.
1. It's not particularly _useful_, since the preview window is not a very good video player. It's not like there's some great advantage to being able to "detach" our preview from the NLE interface completely.
1. Obviously the `openshot-qt` process would be prevented from exiting until the window closes — does that mean the user can't relaunch the main application? If they double-click the "OpenShot Video Editor" app icon after closing the main window, but with the preview still running, does the main window open? Or does "nothing happen" because the preview window is still hanging around?
1. How much memory does OpenShot release when the main window is closed, if the preview is still there? Any? **None?** If all of its memory allocations are retained until that window's closed and the process can finally die, that strikes me as a very real problem. If it reduces to a more modest footprint, then it's not as big a deal.

SuslikV:
Behavior can be changed via `get_app().quit()` - https://github.com/OpenShot/openshot-qt/pull/2936#issuecomment-521632642
Yeah, it is more natural to close all that were spawn.

PR #309, opened 2019-08-12
Remove mentions of nonexistent InitFileInfo()

ferdnyc:
Various documentation comments gave instructions about how this or that class _must call_ `InitFileInfo()` — except, no such method exists.

(There is an `InitEffectInfo()`, also changed one comment to correctly reference it by the proper name.)


ferdnyc:
Another pure docs change, so I'm going to commit this.

PR #311, opened 2019-08-18
Fix initial offset calculations logic for the PTS

SuslikV:
**Edit:** meaning of the PR was changed. Some info below is obsolete now.

------

This is kind of work-in-progress or more like question to the developer.

This is attempt to implement some fixes to resolve the: https://github.com/OpenShot/openshot-qt/issues/2828

This is **NOT** ready to be merged.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -1855,8 +1855,8 @@ void FFmpegReader::UpdatePTSOffset(bool is_video) {
 		// VIDEO PACKET
 		if (video_pts_offset == 99999) // Has the offset been set yet?
 		{
-			// Find the difference between PTS and frame number (no more than 10 timebase units allowed)
-			video_pts_offset = 0 - std::max(GetVideoPTS(), (int64_t) info.video_timebase.ToInt() * 10);
+			// The difference of obtained PTS from 0, for the first frame
+			video_pts_offset = 0 - GetVideoPTS();
 
 			// debug output
 			ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::UpdatePTSOffset (Video)", "video_pts_offset", video_pts_offset, "is_video", is_video);
@@ -1865,8 +1865,8 @@ void FFmpegReader::UpdatePTSOffset(bool is_video) {
 		// AUDIO PACKET
 		if (audio_pts_offset == 99999) // Has the offset been set yet?
 		{
-			// Find the difference between PTS and frame number (no more than 10 timebase units allowed)
-			audio_pts_offset = 0 - std::max(packet->pts, (int64_t) info.audio_timebase.ToInt() * 10);
+			// The difference of obtained PTS from 0, for the first frame
+			audio_pts_offset = 0 - packet->pts;
 
 			// debug output
 			ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::UpdatePTSOffset (Audio)", "audio_pts_offset", audio_pts_offset, "is_video", is_video);


SuslikV:
Commit:
* **Fix 1st frame offset calculation**

Question:
@jonoomph Am I understanding right the meaning of the _offset_ by this PR? Despite it will broke the reading of many files that starts not from the 0 pts (audio or video), what was the real meaning of the _offset_ in libopenshot? Because it was implemented incorrectly - timebase vs seconds units comparison inside the _max_ method - it almost always returned 0 for everything and thus the _video_pts_offset_ may be as large as, for example, -29184 that equals to 1.9 sec for video with 15360 timescale (the media file from the linked issue). Comments inside the code confuses me. It calculates right, large negative offsets returns from the function - but comments says that only 10 timebase units allowed. Maybe offsets shouldn't have any restrictions at all?

SuslikV:
@ferdnyc the `.ToInt()` returns seconds as **number**, exact code:
```
(int) round(1.0/15360) => (int) 0.0000651041(6) => 0
```
This is not timebase's denominator (custom method is in use of the `Fraction.cpp`) - you cannot multiply it to 10 just to get ten times bigger denominator of the timebase. `GetVideoPTS()` returns number of timebase units needs to be taken to get some time value, like: you need to take 29184 of the timebases (0.0000651041(6)) to get 1.9 seconds.

Thus either comment is right and the formula is wrong or both comment and formula is wrong and should be rewritten.

ferdnyc:
You're absolutely right, I didn't catch the `ToInt()`, that's *definitely* wrong.

Still, the concept of "minimum allowed offset" doesn't really make sense to me. How it why would that ever not be 0? So, I think the formula should probably just be `0 - min(GetVideoPTS(), 10)` -- as you point out, it doesn't need correction into `time_base` units, it already is. (The conversion to seconds happens a bit farther down in the code.)

SuslikV:
Even files exported by OpenShot not always starts from 0 pts, thus I had thoughts that offset had absolutely different meaning.

SuslikV:
Now, I know that this option is just dead code. And has misleading comment string above it.

SuslikV:
Updated on top of the current _develop_ branch. Changed meaning of the PR. This is logic fix now.

jonoomph:
Sorry for my delay on this one. The offset and limit was intended to handle video files that contain some crazy initial audio timestamps. It's been a long time, but if I recall correctly, some encoders start encoding the audio stream prior to the video, and timestamps don't start at the same place as the video packets, in some extreme cases, the first few audio timestamps were hours away from the video timestamps. 

Most video players will just disregard timestamps that are too old, so it's no big deal for playback. But when I'm trying to "distribute" the audio samples across frames, these large gaps create havoc. Hope that makes sense.

N3WWN:
Could this be the reason that OpenShot rendered videos always start with the 2nd frame in the video?  ( https://github.com/OpenShot/openshot-qt/issues/733 )

SuslikV:
@jonoomph I still don't understand how it was helpful if it never worked (at least mentioned "limit")

@N3WWN Nope. The starting PTS in output is different thing - it placed into the packet by the routines from the `src/FFmpegWriter.cpp`.  While this is only media reader not writer. The PTS of each  video file may be different. And final (combined) frame may consist of a number of blended frames (from the input files), thus all output PTS'es are produced by the algorithm that writes/makes these packets. This will change nothing for you. More over, it will change nothing in OpenShot too.

Again, THIS PR IS JUST FOR clear reading of the existing code (and no matter what it was intended for). Any of this `max`'es weren't working before and now, and never be, unless you are using timebase=10 or similar value (fps = 10, or 10Hz sampled audio or less).

jonoomph:
Addressing this in a new PR: https://github.com/OpenShot/libopenshot/pull/565

PR #313, opened 2019-08-21
support for HEVC (and possibly more) HW VAAPI encoding

k3a:
Ability to use more HW encoders (on top of existing h.264). 
Additional `FFMPEG_LOGGING` define to optionally redirect ffmpeg messages to the debug log.

Tested on AMDGPU for both:
- h.264
- HEVC (h.265)

The pull request for HEVC preset is at OpenShot/openshot-qt#2941.


ferdnyc:
@k3a Thanks, this looks great! I won't be able to _test_ it myself, as my GT 710 doesn't even support HEVC hardware decoding, never mind encoding, but I can still offer code reviews at least. :wink: 

I'll try to take a more detailed look later this week, but I can already see that I'm going to suggest tthe `#define`s and the definition of `ffmpeg_log_callback()` be moved into `include/FFmpegUtilities.h` as macros that can be invoked from `FFmpegReader.cpp` and `FFmpegWriter.cpp`. (You'll see from context how that's typically done.) 

_Especially_ if there are any differences needed between FFmpeg versions. Are the `av_log_*` calls and the `ffmpeg_log_callback()` signature identical for all FFmpeg versions 2.x – 4.2+? (I'll check myself, but can't right now.)

...TBH, I'm sort of leaning towards suggesting that the logging and the HEVC support get split into separate PRs, they're really totally unrelated changes.

This looks really good, though, and I'm excited to see it added to libopenshot! Thanks again.

SuslikV:
I agree, redirection of the FFmpeg logging should be in other PR.

k3a:
Makes sense. I will separate it later today.

Ffmpeg logging is not something really important but it can help in some cases, e.g. when a ffmpeg function fails with some generic error like Invalid arguments. So I think it is something good to have to optionally enable. Probably not worth having it enabled all the time as it is pretty verbose.

src/FFmpegWriter.cpp
```diff
@@ -1361,23 +1352,51 @@ void FFmpegWriter::open_video(AVFormatContext *oc, AVStream *st) {
...
+					"codec_id", video_codec->codec_id);
+				break;
+		}
+
 		// set hw_frames_ctx for encoder's AVCodecContext
 		int err;
 		if ((err = set_hwframe_ctx(video_codec, hw_device_ctx, info.width, info.height)) < 0) {
-				fprintf(stderr, "Failed to set hwframe context.\n");
+				ZmqLogger::Instance()->AppendDebugMethod("FFmpegWriter::open_video ERROR faled to set hwframe context",
+					"width", info.width, "height", info.height);
```
ferdnyc: This could include more context, even — something like
"FFmpegWriter::open_video (set_hwframe_ctx) ERROR faled to set hwframe context"
(Just to make the call searchable within the code. But it's fine as-is, too.)

I was going to suggest adding an `av_err2str(err)` argument to the logging call, but the log redirection (your other PR, hopefully upcoming? :grin: ) should take care of passing along any error message that results from a failed call, so no sense in creating duplicate messages.

src/FFmpegWriter.cpp
```diff
@@ -173,58 +173,49 @@ void FFmpegWriter::SetVideoOptions(bool has_video, string codec, Fraction fps, i
...
 			hw_en_av_device_type = AV_HWDEVICE_TYPE_DXVA2;
-		}
-		else {
-			if ( (strcmp(codec.c_str(),"h264_nvenc") == 0)) {
+		} else if (strstr(codec.c_str(), "_nvenc") != NULL) {
 				new_codec = avcodec_find_encoder_by_name(codec.c_str());
 				hw_en_on = 1;
 				hw_en_supported = 1;
 				hw_en_av_pix_fmt = AV_PIX_FMT_CUDA;
 				hw_en_av_device_type = AV_HWDEVICE_TYPE_CUDA;
```
ferdnyc: Outdent these lines to the same level as the previous block, please.

src/FFmpegWriter.cpp
```diff
@@ -173,58 +173,49 @@ void FFmpegWriter::SetVideoOptions(bool has_video, string codec, Fraction fps, i
...
-			}
-			else {
+		} else {
 				new_codec = avcodec_find_encoder_by_name(codec.c_str());
 				hw_en_on = 0;
 				hw_en_supported = 0;
```
ferdnyc: And these as well.

src/FFmpegWriter.cpp
```diff
@@ -1361,23 +1352,51 @@ void FFmpegWriter::open_video(AVFormatContext *oc, AVStream *st) {
...
+					video_codec->i_quant_offset = 0;
+					video_codec->b_quant_factor = 6.0/5;
+					video_codec->qmin = -1;
+					video_codec->qmax = -1;
+				*/
+				break;
+			default:
+				ZmqLogger::Instance()->AppendDebugMethod("No codec-specific options defined for this codec. HW encoding may fail",
+					"codec_id", video_codec->codec_id);
+				break;
```
ferdnyc: I'm tempted to say this should  just `throw InvalidOptions("No hardware encoding parameters have been defined for this codec.", path);` or something like that, instead. It doesn't feel like there's any reasonable chance that an unknown, unconfigured codec _would_ successfully be encoded by the hardware, right?

src/FFmpegWriter.cpp
```diff
@@ -1361,23 +1352,51 @@ void FFmpegWriter::open_video(AVFormatContext *oc, AVStream *st) {
...
+				video_codec->max_b_frames = 0;        // At least this GPU doesn't support b-frames
+				video_codec->pix_fmt   = hw_en_av_pix_fmt;
+				video_codec->profile = FF_PROFILE_H264_BASELINE | FF_PROFILE_H264_CONSTRAINED;
+				av_opt_set(video_codec->priv_data, "preset", "slow", 0);
+				av_opt_set(video_codec->priv_data, "tune", "zerolatency", 0);
+				av_opt_set(video_codec->priv_data, "vprofile", "baseline", AV_OPT_SEARCH_CHILDREN);
+				break;
+			case AV_CODEC_ID_HEVC:
+				// must be zero (at least on AMDGPU on Linux 8/2019)
+				video_codec->bit_rate = 0;
```
ferdnyc: Hrrrm. This is the only part I'm _really_ unsure about. (Along with the preset in OpenShot/openshot-qt#2941 that defines High, Medium, and Low quality settings with identical 0 bitrates — though that's admittedly somewhat our fault for having such an inflexible preset-definition syntax.) It's odd that bitrate wouldn't work, too, since the documentation [_seems_ to indicate](http://www.ffmpeg.org/ffmpeg-codecs.html#VAAPI-encoders) that it's supported... I guess maybe this is part of VAAPI on AMD [only having "partial" support](https://trac.ffmpeg.org/wiki/HWAccelIntro#PlatformAPIAvailability), perhaps?

Still... I don't know if you saw that OpenShot _also_ has CRF support, as an alternative to raw bitrates — we're [already using it](https://github.com/OpenShot/openshot-qt/blob/develop/src/presets/format_mp4_x265_crf.xml) for software-encoded HEVC in the x265 preset, in fact.

Unfortunately, fat lot of good _that_ does us, either, because [VAAPI doesn't support that, either](https://trac.ffmpeg.org/wiki/Hardware/VAAPI#Mappingoptionsfromlibx264).
> No CRF-like mode is currently supported. The only constant-quality mode is CQP (constant quantisation parameter), which has no adaptivity to scene content. It does, however, allow different quality settings for different frame types, to improve compression by spending fewer bits on unreferenced B-frames - see the `(i|b)_q(factor|offset)` options.

However, here again it _seems_ to imply that setting a bitrate is supported and will use VBR by default — though it does say that it can't be combined with a `qp` setting... Just out of curiosity, when you tested setting `bit_rate`, were you _**also**_ setting `qp`? Sounds like that may be why it wouldn't work (emphasis mine):
> **CQP mode cannot be combined with a maximum bitrate or buffer size.**
>
> CBR and VBR modes are supported, though the output of them varies significantly by driver and device (default is VBR, set `-maxrate` equal to `-b:v` for CBR). HRD buffering options (rc_max_rate, rc_buffer_size) are functional, and the encoder will generate buffering_period and pic_timing SEI when appropriate.

They also give some examples:
> Encode an input with H.264 at 5Mbps VBR:
> ```
> ffmpeg -vaapi_device /dev/dri/renderD128 -i input.mp4 -vf 'format=nv12,hwupload' -c:v h264_vaapi -b:v 5M output.mp4
> ```

> Encode with H.264 at good constant quality:
> ```
> ffmpeg -vaapi_device /dev/dri/renderD128 -i input.mp4 -vf 'format=nv12,hwupload' -c:v h264_vaapi -qp 18 output.mp4
> ```
> Encode with 10-bit H.265 at 15Mbps VBR (recent hardware required - Kaby Lake or later Intel):
> ```
> ffmpeg -vaapi_device /dev/dri/renderD128 -i input.mp4 -vf 'format=p010,hwupload' -c:v hevc_vaapi -b:v 15M -profile 2 output.mp4
> ```

If setting bitrate can't be made to work, I think we need to provide _some_ way of configuring the quality targets so we have real High, Medium, and Low options. Maybe it's time to implement a `cqp` bitrate mode that sets `qp`, to go along with the `crf` support here:
https://github.com/OpenShot/libopenshot/blob/857297abdf3b3c4d8067b5f32559d10269b2d7e0/src/FFmpegWriter.cpp#L399-L463

src/FFmpegWriter.cpp
k3a: it is actually a good idea - these errors are "higher-level" enum-like, useful to give some generic error overview. Full ffmpeg is verbose and I am not sure if it should be enabled every time a OpenShot debug log is enabled. Let's talk about it in the separate error PR but I will include av_err2str here just in case.

src/FFmpegWriter.cpp
k3a: actually there is some possibility it may work with some default options (default quality). For example HEVC would work without any config as long as the bitrate were 0. I will do some tests with other codecs on my AMD card out of curiosity.

src/FFmpegWriter.cpp
k3a: This is a bit problematic! I started with those example ffmpeg commands (with -b:v 5M) but it produced garbage. I tried that on two Linux machines (gentoo-based and arch-based) with open-source AMDGPU driver and two different AMD cards. Setting bitrate to 0 and modifying qp did the trick. Do you think bitrate has to work? Maybe there is a driver or encoder bug?

Original: https://k3a.me/u/big_buck_bunny_720p_1mb.mp4
Encoded: https://k3a.me/u/bunny_bug.mp4

Encoding done with
`fmpeg -vaapi_device /dev/dri/renderD128 -i big_buck_bunny_720p_1mb.mp4 -vf 'format=nv12,hwupload' -c:v hevc_vaapi -qp:v 25 -b:v 5M -y bunny_bug.mp4`

Not setting `-qp:v 25` (but keeping -b:v 5M) resulted in another garbage. :( 
Here: https://k3a.me/u/bunny_bug2.mp4

GPU Radeon RX 550 (POLARIS11), ffmpeg n4.1.3, libva-mesa-driver-19.1.2-1

From the [ffmpeg source](https://github.com/FFmpeg/FFmpeg/blob/20fed2f0ab197d60801280dfc844f6b29a397ff2/libavcodec/vaapi_encode.c#L1479) it looks like there are several modes. I will try to look into it and make it select a different mode on Monday/Tuesday and see if there is a way to use bit_rate on my HW and report back.

ferdnyc: > This is a bit problematic!

Yeah, it's really a pain. I'm hoping we can lick it with enough trial-and-error, though it's sad that it always seems to come down to that with ffmpeg. I'll do some stackoverflow-and-etc. scouring over the weekend as well, though.

One other test, if you'd be willing? The same no-QP encode, but also setting `-maxrate` as they'd suggested, for _fixed_ bitrate. I'm curious whether _that_ might at least work:
```
ffmpeg -vaapi_device /dev/dri/renderD128 -i big_buck_bunny_720p_1mb.mp4 \
-vf 'format=nv12,hwupload' -c:v hevc_vaapi -b:v 5M -maxrate 5M -y bunny_bug.mp4
```

k3a: Unfortunately the same - https://k3a.me/u/bunny_bug3.mp4 :(

src/FFmpegWriter.cpp
ferdnyc: > Full ffmpeg is verbose and I am not sure if it should be enabled every time a OpenShot debug log is enabled. 

At the moment, the big problem is that debug libopenshot is even _more_ verbose — and as things stand it's very much all-or-nothing, either you have debugging turned on and you get _flooded_ with messages for every frame encoded, or you don't even get ERROR messages when something goes wrong on the ffmpeg end. We need to strike a better balance, or implement some sort of multi-level message passing (not sure if it should be level-setting or multiple distinct ZMQ streams, but _something_)... not there yet, though.

> Let's talk about it in the separate error PR but I will include av_err2str here just in case.

Sounds good, thanks!

src/FFmpegWriter.cpp
ferdnyc: Poop. Oh well. But setting `-qp:v` works fine? Weird. Do you need to _explicitly_ set bitrate to 0, or can you just omit it? 

If an explicit bitrate setting can be omitted entirely, then I'm leaning back towards my "cqp mode" fix. That wouldn't be so hard to implement, it's basically a cut-and-paste mod of the existing crf code. So, that's an option, I'll think about putting it together (would be good to have anyway) and bounce the idea off @jonoomph . (...Consider it bounced! **#tag**)

k3a:
setting `-qp:v` works fine. And can be done from the code. Bitrate is probably 0 by default, unless explicitly specified.

src/FFmpegWriter.cpp
k3a: Sorry, I had to focus on something else past week but I am ready to work on this tomorrow.

I did some ffmpeg cmdline testing today after updating the system and it seems that bitrate setting now magically works in the updated `libva-mesa-driver-19.1.5-1`. It works for AMD but doesn't for Intel (well it at least produces a working file but with 76 kb/s bitrate).

ferdnyc: Hm. You lost me there, which hardware produces a file with 76kbps bitrate, and under what conditions?

k3a: Sorry, I will try to explain.

In the meantime, I updated my Arch Linux box and there were new versions of FFMpeg and Mesa. I also rebooted and enabled iGPU (internal GPU in the CPU) as I realized it should support VP9 too. So now I have two VAAPI HW encoders. And with those updated libraries, HW encoding started behaving better - it made usable files using AMD as well as Intel but Intel had 76 kb/s bitrate (despite using the same ffmpeg comamndline options).

It is much easier to look at my test script here - https://gist.github.com/k3a/dcfb3c34745899c50eb1bb0967c0f3be those commented-out lines were also tested. Comments preceding a line belongs to that line.

I've spent almost the whole day today on this and found this out:
1. it is possible to ask ffmpeg which options an encoder support (`ffmpeg -h encoder=h264_vaapi`). The problem is, those seem to be codec-specific but encoding also accepts some (which?!) of the common ones, like bitrate, of course..
2. so `vaapi` encoders all support `rc_mode` which is so called explicit resource control mode
3. ffmpeg's `libavcodec/vaapi_encode.c` decides which rc_mode to apply depending on various combinations and it looks like it can be wrong sometimes and those rules can change with each ffmpeg version bump
4. if I set rc_mode explicitly, it started behaving consistently, except for "VBR" mode, where `maxrate` had to be set as well.
5. I had to enable Ffmpeg log while working on it. It really gives very helpful output if some codec configuration doesn't work so it will be really good to have later (not part of this commit). For example:
```
ffmpeg: Input surface format is nv12.
ffmpeg: Compatible profile VAProfileVP9Profile0 (19) is not supported by driver.
ffmpeg: No usable encoding profile found.
```
6. There needs to be proper HW card detection made in openshot-qt. Currently it allows no acceleration or card 1. I have two cards now, though. On linux, it is quite easy to do, just look for dev files `ls -1 /dev/dri/renderD12*`:
```
/dev/dri/renderD128 # that is my AMD
/dev/dri/renderD129 # that is my Intel
```
maybe there is some better way to detect it or even get GPU name, I don't know yet. But I know you can do `vainfo --display drm --device /dev/dri/renderD128 | head -n3` to get:
```
vainfo: VA-API version: 1.5 (libva 2.5.0)
vainfo: Driver version: Mesa Gallium driver 19.1.5 for Radeon RX 560 Series (POLARIS11, DRM 3.32.0, 5.2.9-arch1-1-ARCH, LLVM 8.0.1)
vainfo: Supported profile and entrypoints
```
7. original property setting for h.264 HW encoding may be useless. Unfortunately it is hard to find proper docs for ffmpeg but here is what I think about those:
```c
// probably nvenc only (Fmpeg/libavcodec/nvenc_h264.c)
av_opt_set(video_codec->priv_data, "preset", "slow", 0);
// seems like non-sense. It is top level bool parameter (Fmpeg/libavcodec/nvenc_h264.c)
// should probably be av_opt_set(video_codec->priv_data, "zerolatency", "1", 0);
av_opt_set(video_codec->priv_data, "tune", "zerolatency", 0);
// probably nonsense. Couldn't grep-find vprofile in the whole ffmpeg source; I think only generic video_codec->profile is used here
av_opt_set(video_codec->priv_data, "vprofile", "baseline", AV_OPT_SEARCH_CHILDREN);
```
I kept it as it was there though; in order to prevent breaking previously-working code.

I've updated the code and also included your cqp commit. But I haven't tested cqp yet. 
I tested that H264, HEVC and VP9 produce usable results. 
For 15 Mb/s default OpeNshot export settings it produced:

**AMD tests**:
H.264 9062 kb/s
HEVC 15002 kb/s
VP9 unsupported

**Intel test**:
H.264 9202 kb/s
HEVC 15252 kb/s
VP9 15778 kb/s

Any comments or problems with the pull requests?

jonoomph:
@k3a Can I please get an update on this PR? I'm super excited about it, but I'm just playing catch up right now, and would like to know if this is ready for merging? close to ready? Anything I can do to help? Thx!

src/examples/Example.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# OpenShot Library (libopenshot) is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with OpenShot Library. If not, see <http://www.gnu.org/licenses/>.
+
+import sys
```
jonoomph: Remove this import, and you should pass the Codacy/PR Quality Review

k3a: From my point of view, it was ready Sep 4 (https://github.com/OpenShot/libopenshot/pull/313/#discussion_r320969868). It produced correct results on my AMD and Intel HW encoders. The remaining work would be adding encoding presets to openshot-qt and consider improving HW encoder detection and selection (some people like me have both CPU and GPU encoders available).

jonoomph:
LG! Merging this as is. Hopefully it will work with all 3 OSes, if not... I might revert. But let's see how the merge and build servers do. :crossed_fingers:

PR #314, opened 2019-08-24
Fix return type mismatch

SuslikV:
`video_position` of the class `PlayerPrivate` declared as `int64_t` inside the _include/Qt/PlayerPrivate.h_

Note: The `ClipBase` class has its own `Position()` method that returns `float` (used in the Timeline operations, the latter has number of internal limits that is not affected by this PR).

### DIFF
--- a/include/PlayerBase.h
+++ b/include/PlayerBase.h
@@ -81,7 +81,7 @@ namespace openshot
 		virtual void Pause() = 0;
 
 		/// Get the current frame number being played
-		virtual int Position() = 0;
+		virtual int64_t Position() = 0;
 
 		/// Seek to a specific frame in the player
 		virtual void Seek(int64_t new_frame) = 0;
--- a/include/QtPlayer.h
+++ b/include/QtPlayer.h
@@ -81,7 +81,7 @@ namespace openshot
 	void Pause();
 	
 	/// Get the current frame number being played
-	int Position();
+	int64_t Position();
 	
 	/// Seek to a specific frame in the player
 	void Seek(int64_t new_frame);
--- a/src/QtPlayer.cpp
+++ b/src/QtPlayer.cpp
@@ -127,7 +127,7 @@ void QtPlayer::Pause()
     Speed(0);
 }
 
-int QtPlayer::Position()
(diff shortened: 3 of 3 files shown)


SuslikV:
Just forgot to add another entry of the method, so updated it. It seems that the return value only used in _src\Qt\PlayerDemo.cpp_ and not changes a lot.

ferdnyc:
> Note: The `ClipBase` class has its own `Position()` method that returns `float` (used in the Timeline operations, the latter has number of internal limits that is not affected by this PR).

Yeah, `Clip`'s (and `ClipBase`'s) `position` _is_ a float, like `start` and `end` it's a time in fractional seconds rather than a frame number.
https://github.com/OpenShot/libopenshot/blob/857297abdf3b3c4d8067b5f32559d10269b2d7e0/include/ClipBase.h#L56-L63
https://github.com/OpenShot/libopenshot/blob/857297abdf3b3c4d8067b5f32559d10269b2d7e0/include/ClipBase.h#L82-L88

This looks good to me; there's actually more that could be done along these lines, if you're interested in broader cleanup. Offhand:

* `AudioBufferSource` and `AudioReaderSource` use the JUCE type `int64` (actually `juce::int64`) rather than `int64_t`. Which is fine for internals, but they even use it it in their public APIs. If `int64_t` is compatible, seems like it'd be preferable to expose the standard `int64_t` type instead? But that's a "maybe", it'd probably require chasing down the type definition/documentation and determining whether it's safe to convert.

* `Timeline` and `KeyFrame` still use `long` frequently, when storing time-based position values converted into frame numbers. Commit c95db460d replaced all of the `long int` types with `int64_t` across the codebase, and that includes all of the _external_ API for both classes. But it seems because these internal references were written as `long` instead of `long int`, they got missed.

SuslikV:
>... `Timeline` and `KeyFrame` still use `long` frequently, when storing time-based position values converted into frame numbers.

Maybe I will look at them later (not in this PR). But _this_ PR it seems to be harmless by any means.

ferdnyc:
Makes sense. Merging.

PR #315, opened 2019-08-27
Exceptions: Make (most) file_path args optional

ferdnyc:
It's official: "Don't make programmers write awkward code" is now a trilogy! (The first two thrilling installments: #266 , OpenShot/openshot-qt#2853.) That's right, it now has _seven_ Executive Producer credits, four of whom are just the overpaid stars of the franchise. We'll probably kill one of them off this time, only to bring them back two PRs later.

This sequel concerns the libopenshot exceptions, with a fix very analogous to our first installment (but less impactful or widespread).

### Don't make programmers write awkward code

Many of libopenshot's exception classes take a `file_path` argument, to denote the input/output file in use when the exception occurred. Which is fine, if your exception involves a file path. But, not every exception does, as many are encountered by classes that don't operate exclusively on disk files. So you get code like this:

https://github.com/OpenShot/libopenshot/blob/5e1e8ce601952db3160000998e54454cc2da77cd/src/effects/Blur.cpp#L289-L291

https://github.com/OpenShot/libopenshot/blob/5e1e8ce601952db3160000998e54454cc2da77cd/src/Clip.cpp#L392-L397

#### Once again, a job for the compiler

Just like in #266, filling in empty strings for optional arguments is a job better left to the compiler than the programmer. So, I added a default arg of `""` to every exception that takes a `file_path` argument, _except_ when it was something like the `InvalidFile` exception. For that one, a `file_path` really _should_ be required!

Then I replaced every instance of `throw _____("...", "");` in the codebase with `throw _____("...");` and committed the change. (The only exceptions involved were the two above, as it turns out. Only `InvalidJSON` and `ReaderClosed` are ever _actually_ called with an empty `file_path` string, as it turns out. But, now the rest _can_ be if it's ever needed.)

### DIFF
--- a/include/Exceptions.h
+++ b/include/Exceptions.h
@@ -111,7 +111,7 @@ namespace openshot {
 	{
 	public:
 		std::string file_path;
-		InvalidChannels(std::string message, std::string file_path)
+		InvalidChannels(std::string message, std::string file_path="")
 			: BaseException(message), file_path(file_path) { }
 		virtual ~InvalidChannels() noexcept {}
 	};
@@ -121,7 +121,7 @@ namespace openshot {
 	{
 	public:
 		std::string file_path;
-		InvalidCodec(std::string message, std::string file_path)
+		InvalidCodec(std::string message, std::string file_path="")
 			: BaseException(message), file_path(file_path) { }
 		virtual ~InvalidCodec() noexcept {}
 	};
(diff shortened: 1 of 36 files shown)


ferdnyc:
Merging, since it's purely a cosmetic change to make things more readable for developers, with no real impact on the compiled output.

PR #318, opened 2019-08-30
CMake: Rename LIBOPENSHOT_AUDIO to OpenShotAudio

ferdnyc:
The `freeze.py` script in openshot-qt still contains a lot of hardcoded paths for things like library `.so` files, that only apply on the project build servers (if even there).

To generalize those better, I had the thought that we could leverage CMake's `--find-package` commandline mode even from the non-CMake `openshot-qt` project. For example:
```
$ cd /path/to/repo/libopenshot
$ cmake -DCMAKE_MODULE_PATH=cmake/Modules --find-package -DNAME=FFmpeg \
-DCOMPILER_ID=GNU -DLANGUAGE=CXX -DMODE=LINK 

/lib64/libavcodec.so /lib64/libavformat.so /lib64/libavutil.so
```
However, that mode _requires_ that the variable names used to store the results of the Find module match its name. When I rewrote `FindOpenShotAudio.cmake`, I retained the `LIBOPENSHOT_AUDIO_` variable names for convenience. But, that won't work for `--find-package`. 

So, this PR renames `LIBOPENSHOT_AUDIO_INCLUDE_DIRS`, `LIBOPENSHOT_AUDIO_LIBRARY`, etc. to `OpenShotAudio_INCLUDE_DIRS`, `OpenShotAudio_LIBRARY`, and so on.

With this change, we can use `--find-package` to locate `libopenshot-audio.so`, and if we were to write a similar find module for `libopenshot.so`, that would work the same way:
```
$ cd /path/to/repo/libopenshot
$ cmake -DCMAKE_MODULE_PATH=cmake/Modules --find-package -DNAME=OpenShotAudio \
-DCOMPILER_ID=GNU -DLANGUAGE=CXX -DMODE=LINK

/usr/lib64/libopenshot-audio.so 
```

For backwards-compatibility / convenience, the `LIBOPENSHOT_AUDIO_DIR` environment variable and CMake variable are both still respected, along with a new CMake variable `OpenShotAudio_DIR` which can (should) be used instead.

### DIFF
--- a/cmake/Modules/FindOpenShotAudio.cmake
+++ b/cmake/Modules/FindOpenShotAudio.cmake
@@ -1,9 +1,9 @@
 # - Try to find JUCE-based OpenShot Audio Library
 # libopenshot-audio;
 # Once done this will define
-#  LIBOPENSHOT_AUDIO_FOUND - System has libjuce.so
-#  LIBOPENSHOT_AUDIO_INCLUDE_DIRS - The juce.h include directories
-#  LIBOPENSHOT_AUDIO_LIBRARIES - The libraries needed to use juce
+#  OpenShotAudio_FOUND - System has libjuce.so
+#  OpenShotAudio_INCLUDE_DIRS - The juce.h include directories
+#  OpenShotAudio_LIBRARIES - The libraries needed to use juce
 
 if("$ENV{LIBOPENSHOT_AUDIO_DIR}" AND NOT "${OpenShotAudio_FIND_QUIETLY}")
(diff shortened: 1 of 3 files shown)


ferdnyc:
Withdrawing and closing this, since I have a WIP implementation of exported targets underway for libopenshot-audio, and I'd rather go that way instead. Using it would allow us to do away with `FindOpenShotAudio.cmake` entirely, at least in the libopenshot build.

(Instead, we'd use):
```cmake
find_package(OpenShotAudio REQUIRED)
target_link_libraries(openshot PUBLIC OpenShot::Audio)
```

PR #321, opened 2019-09-07
WIP: Towards a more Pythonic openshot.Fraction

ferdnyc:
This extends openshot::Fraction in the Python wrapper with some additional methods implementing Python builtins, to permit:
```py
f = openshot.Fraction(16, 9)
float(f) # returns 1.777777777
int(f) # returns 2
print(f) # outputs "16:9"
dict(f.GetMap()) # returns f in Python dict form
```
No more calling `ToInt()`, `ToDouble()`, etc. from the Python code.

The last example, the `dict` accessor, was mostly the focus of this, as it would allow the replacement of this OpenShot code:
```python
self._data["fps"] = {"num" : profile.info.fps.num, "den" : profile.info.fps.den}
```
with simply:
```python
self._data["fps"] = profile.info.fps.GetMap()
```
I feel like it _should_ be possible to wrap openshot::Fraction so that `dict(f)` will work directly, and the previous could be simply:
```python
self._data["fps"] = dict(profile.info.fps)
```
...but I haven't quite found it yet, SWIG-wise. (At one point I had a `__getitem__()` implemented, so you could use `f["num"]` and `f["den"]` as well, but it was kind of ugly and still didn't allow `dict(f)`.)

I've marked this WIP for now, not because I don't think it's complete/ready, but because I think more could be done. It seems silly to _only_ improve `openshot.Fraction` (though it is by far the most-used-in-non-Pythonic-ways of all the `openshot::` classes, and the one that will benefit the Python side of things the most with these improvements).

Still, consider this a call for suggestions, if anyone can think of _other_ classes that could be made to fit more comfortably in with the Python code with the addition of a few wrapper methods.


jonoomph:
Conflicts on this one too

ferdnyc:
I mean, I'll fix 'em anyway... :wink:

jonoomph:
LG!

PR #325, opened 2019-09-12
Example.py: Remove useless `import sys`

ferdnyc:
Codacy pointed out that there's absolutely no reason to be
importing the sys module in this code.


ferdnyc:
Merging as almost impossibly trivial.

PR #326, opened 2019-09-13
Some minor fixes made while testing hwaccel

ferdnyc:
(And mostly unrelated to hardware acceleration.)

- Clean up some misleading indentation in FFmpegWriter
- Replace this FFmpegReader message (printed on stdout):
   ```
   
   
   Decodiing Device Nr: 2
   ```
  With this:
   ```
   Hardware decoding device number: 2
   ```
- Eliminate the hardcoded `/home/jonathan/` source-file path in `Example.cpp` by adding a definition of `TEST_MEDIA_PATH` to `src/CMakeListst.txt`, same as `tests/CMakeLists.txt`, and using that as the path to the input file.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -372,6 +372,14 @@ target_link_libraries(openshot ${REQUIRED_LIBRARIES})
 # Create test executable
 add_executable(openshot-example examples/Example.cpp)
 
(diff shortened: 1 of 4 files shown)


jonoomph:
LGTM

PR #328, opened 2019-09-15
Travis updates (Bionic, Qt 5.12, config)

ferdnyc:
This PR modernizes our Travis build configs (no more manual `sudo`, packages managed using Travis' supported `apt` addon, etc.), as well as makes some updates:
* The FFmpeg 3 and 4 builds are upgraded to their Ubuntu 18.04 (Bionic) image. FFmpeg 2 remains on the 16.04 (Xenial) image
* The Bionic builds are also upgraded from Qt 5.10 to 5.12.3
* The FFmpeg 3 PPA is dropped, because Bionic includes FFmpeg 3 by default

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,53 +1,88 @@
-dist: xenial
-sudo: required
+language: cpp
+
+# This section uses a rather esoteric (and tricky!) feature of YAML,
+# &aliases and *anchors, to build package lists out of sublists without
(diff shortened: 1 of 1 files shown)


ferdnyc:
I'm going to merge this, since if it breaks anything it will only be the CI, which can't affect users. And then I can fix it.

PR #329, opened 2019-09-15
CMakeLists: Also detect cppzmq headers

ferdnyc:
We don't _just_ use the ZeroMQ library, we also use the optional (and external) cppzmq C++ bindings.

Previously we weren't detecting for those at all, mostly because Ubuntu just arbitrarily bundles them with its `libzmq3-dev` package. Most distros don't, and when packaged properly cppzmq includes appropriate CMake configuration to make itself discoverable.

This PR adds a check for those configs, and just adds its include dir to the build (as it's a header-only library).


jonoomph:
:+1:

PR #330, opened 2019-09-16
Re-enable 'make test' when possible (requires CMake 3.11+)

ferdnyc:
CMake's decision to reserve certain target names (test, package) was amended in CMake 3.11, so that the names are only reserved when the feature in question (CTest, CPack) is actually used in the project.

So, with newer CMake versions, it's again possible to provide a `make test` command to build and run the unit tests. This PR enables that functionality if the CMake version is sufficiently new. The old `make os_test` target is preserved for backwards-compatibility, and should still be preferred for build/CI purposes.

This change also adds a message to the CMake generation output detailing what targets are available for unit tests, and updates the documentation target with a similar message (if Doxygen i available).

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -202,7 +202,7 @@ endif(USE_SYSTEM_JSONCPP)
 
 IF (NOT DISABLE_TESTS)
 	###############  SET TEST SOURCE FILES  #################
-	SET ( OPENSHOT_TEST_FILES tests.cpp
+	SET ( OPENSHOT_TEST_FILES
 		   Cache_Tests.cpp
 		   Clip_Tests.cpp
 		   Color_Tests.cpp
@@ -220,6 +220,7 @@ IF (NOT DISABLE_TESTS)
 
 	################ TESTER EXECUTABLE #################
 	# Create unit test executable (openshot-test)
+	message (STATUS "Tests enabled, test executable will be built as tests/openshot-test")
 	add_executable(openshot-test
 				   tests.cpp
 				   ${OPENSHOT_TEST_FILES} )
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -220,14 +220,32 @@ IF (NOT DISABLE_TESTS)
...
+		cmake_policy(SET CMP0037 NEW)
+	endif()
+	if (CMAKE_VERSION VERSION_GREATER 3.11)
+		message(STATUS "Cmake 3.11+ detected, enabling 'test' target")
+		add_custom_target(test COMMAND openshot-test)
+		list(APPEND OS_TEST_CMDS " or " "'make test'")
+	endif()
+
+	string(CONCAT t ${OS_TEST_CMDS})
+	message("\nTo run unit tests, use: ${t}")
```
ferdnyc: This is a little dumb. I'd originally written it with OS_TEST_CMDS as a list of just the commands, which would then be formatted for output:
```cmake
list(JOIN OS_TEST_CMDS ", " t)
```
...Then I discovered that `list(JOIN)` wasn't added to CMake until 3.12. So, my carefully-constructed logic was ruined, and I kind of hacked around it with a machete. Really `OS_TEST_CMDS` shouldn't even be a list, since there's no point in doing it that way without `list(JOIN)`. Oh well.

tests/CMakeLists.txt
```diff
@@ -220,14 +220,32 @@ IF (NOT DISABLE_TESTS)
...
+	# This requires CMake 3.11+, where the CMP0037 policy
+	# configured to 'NEW' mode will not reserve target names
+	# unless the corresponding feature is actually used
+	if (POLICY CMP0037)
+		cmake_policy(SET CMP0037 NEW)
+	endif()
+	if (CMAKE_VERSION VERSION_GREATER 3.11)
+		message(STATUS "Cmake 3.11+ detected, enabling 'test' target")
+		add_custom_target(test COMMAND openshot-test)
+		list(APPEND OS_TEST_CMDS " or " "'make test'")
```
ferdnyc: This is a little dumb. I'd originally written it with OS_TEST_CMDS as a list of just the commands, which would then be formatted for output:
```cmake
list(JOIN OS_TEST_CMDS ", " t)
```
...Then I discovered that `list(JOIN)` wasn't added to CMake until 3.12. So, my carefully-constructed logic was ruined, and I kind of hacked around it with a machete. Really `OS_TEST_CMDS` shouldn't even be a list, since there's no point in doing it that way without `list(JOIN)`. Oh well.

jonoomph:
This LG (assuming it falls back for older versions of Cmake... which I think it does if I understand it correctly). Mergin'

PR #331, opened 2019-09-19
Delete FindPythonLibs.cmake

ferdnyc:
Our copy of the module is outdated, and fails detecting Python 3.8. (Besides, we shouldn't be mirroring standard CMake modules.)


ferdnyc:
(BTW: This caused me two days of _hell_ trying to figure out why my Fedora Rawhide package builds kept failing, when they succeeded on all the earlier releases. I could see that the standard module had been updated to support Python 3.8 (the default for Rawhide / future-F32), and my simple test `CMakeFiles.txt` with `find_package(PythonLibs)` was finding the Python 3.8 libs just fine.)

jonoomph:
Dang. good catch! LG

PR #333, opened 2019-09-21
Version information: Separate implementation, add pythonic touches

ferdnyc:
With warnings enabled, `g++` was complaining about `GetVersion()` being unused every time the `OpenShotVersion.h` header file was included. So I separated out the implementation into `src/OpenShotVersion.cpp`, which silenced the warnings.

Then I added some conveniences, and some Pythonic flourishes:
1. I moved the initialization of the struct members to the struct itself, instead of in `GetVersion()`, and I made them `const`
1. I changed the const member names to uppercase first letter, since apparently Ruby has some rule about constants starting with a capital letter, and it complained otherwise (and renamed them anyway!)
1. I added a static `openshot::Version`, of type `openshot::OpenShotVersion`
1. I gave `openshot::OpenShotVersion` a string representation in the Python interface file

End result? The warnings are all gone, old code all works the same, but you don't _have_ to do things the old, clunky way anymore:
```python
Python 3.7.4 (default, Jul  9 2019, 16:32:37) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import openshot
>>> openshot.GetVersion().ToString()
'0.2.3'
>>> print(openshot.Version)
0.2.3
>>> str(openshot.Version)
'0.2.3'
>>> str(openshot.Version.Major)
'0'
>>> str(openshot.Version.Minor)
'2'
>>> str(openshot.Version.So)
'17'
>>> print(openshot.OpenShotVersion)
<class 'openshot.OpenShotVersion'>
>>> print(openshot.OpenShotVersion())
0.2.3
```

And then I changed things a bit more, and decided to use `OPENSHOT_VERSION_FULL` as the string representation instead:
```python
>>> import openshot
>>> str(openshot.Version)
'0.2.3-dev1'
>>> print(openshot.OpenShotVersion())
0.2.3-dev1
```

### DIFF
--- a/include/OpenShotVersion.h.in
+++ b/include/OpenShotVersion.h.in
@@ -43,36 +43,28 @@
 #define OPENSHOT_VERSION_SO @PROJECT_SO_VERSION@         /// Shared object version number. This increments any time the API and ABI changes (so old apps will no longer link)
 
 #include <sstream>
-using namespace std;
 
 namespace openshot
 {
 	/// This struct holds version number information. Use the GetVersion() method to access the current version of libopenshot.
 	struct OpenShotVersion {
-		int major; /// Major version number
-		int minor; /// Minor version number
-		int build; /// Build number
-		int so; /// Shared Object Number (incremented when API or ABI changes)
+		static const int Major = OPENSHOT_VERSION_MAJOR; /// Major version number
+		static const int Minor = OPENSHOT_VERSION_MINOR; /// Minor version number
+		static const int Build = OPENSHOT_VERSION_BUILD; /// Build number
+		static const int So = OPENSHOT_VERSION_SO; /// Shared Object Number (incremented when API or ABI changes)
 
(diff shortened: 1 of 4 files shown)


ferdnyc:
Also possible, in this brave new world:
```python
>>> import openshot
>>> str(openshot.Version)
'0.2.3-dev1'
>>> str(openshot.Version) < '0.2.4'
True
>>> str(openshot.Version) < '0.2.3'
False
>>> str(openshot.Version) > '0.2.3'
True
>>> str(openshot.Version) > '0.2.3-dev2'
False
```

ferdnyc:
Merging this. Should be harmless since it's backwards-compatible with older code. But I really want to get a dependent PR submitted to OpenShot, finally cutting it over to using the _new_ interfaces instead.

PR #338, opened 2019-10-02
FFmpegReader::CheckMissingFrame -- checked_frames std::map tweaks

ferdnyc:
`operator[]` for `std::map` [is documented](https://en.cppreference.com/w/cpp/container/map/operator_at) as auto-inserting and initializing members if they're not present, so this change just replaces as few lines of conditional "check-then-initialize-else-increment" logic with a simple `++checked_frames[frame]`, [a common pattern](https://en.cppreference.com/w/cpp/container/map/operator_at#Example).


jonoomph:
LG!

PR #340, opened 2019-10-02
FFmpeg: Create and use targets in CMake build

ferdnyc:
This PR brings two changes to the CMake build environment.

### 1. `cmake/Modules/FindFFmpeg.cmake` will now create IMPORTED targets

The first commit is a major upgrade to the `FindFFmpeg.cmake` module used to discover the FFmpeg library paths on the build system. In a **completely backward-compatible way**, it now not only passes out information via result variables, but also creates for each <var>component</var> library an <code>FFmpeg::<var>component</var></code> target encapsulating that library's details.

Targets are the preferred method of managing build process information in CMake because their information is self-contained, persistent, and transient. If a target is added as a `PUBLIC` or `INTERFACE` dependency of another library target (like our own `openshot` target, for building the shared library), then any executables or other targets linked with that library will automatically (given a new enough CMake, or with workarounds) pick up the same include directories, compile options, etc.

### 2. Link our library and executable targets with the `FFmpeg::` targets

The other two commits replace all of our CMakeLists' result-variable-based management of FFmpeg component dependencies with consumption of the `FFmpeg::` targets. This greatly simplifies the build process. 

1. **ALL** (redundant) FFmpeg discovery is eliminated from `tests/CMakeLists.txt`. The test executable is linked with the `openshot` target, and that in turn is linked with the `FFmpeg::` targets, so nothing more is required.

1. The `src/bindings/{python,ruby}/CMakeLists.txt` files do have to be enhanced with a workaround for include-directory passing. It wasn't until CMake 3.13 (IOW, very recently) that SWIG targets were able to transiently pick up include dirs when generating output files, and even now it has to be explicitly told to do so. We set the appropriate property to enable that feature, if CMake is new enough, and if not the include directories are manually copied from the `openshot` target using a generator expression.

### This is [Modern CMake](https://cliutils.gitlab.io/modern-cmake/)
(Well, it's a good start anyway.) This change brings us much more in line with how CMake is _meant_ to work and to be used. In time, and one step at a time, I hope to get all facets of our build upgraded in similar fashion.


ferdnyc:
In making this change I'd inadvertently made _all_ FFmpeg components `REQUIRED`, which broke the Windows builders as there's no `avresample` there (surprisingly enough). The most recent commit changes the `find_package` call for FFmpeg to use `REQUIRED COMPONENTS avcodec avdevice avformat avutil swscale`, which should fix that. Just waiting on the GitLab builders to confirm...

jonoomph:
LGTM!

PR #341, opened 2019-10-03
Build: prefer system-installed JsonCpp by default, with bundled fallback

ferdnyc:
This change reverses the JsonCpp discovery logic, defaulting `USE_SYSTEM_JSONCPP` **`ON`** instead of `OFF`. Meaning:
1. A system-installed JsonCpp will automatically be probed for and used, if found
2. If not found, the build will automatically fall back to the bundled copy
3. Unless fallback is disabled by setting new option `DISABLE_BUNDLED_JSONCPP` (default `OFF`) to `ON`
4. If system-discovery and bundled-fallback both fail and/or are disabled, the configuration process will abort

This PR also makes use of my commit 491de1dfae2fef9888ab20f69b0630d9129fd74e from #340, to pick up include directories from the library target during SWIG wrapper generation.


jonoomph:
@ferdnyc Conflicts on this one as well

ferdnyc:
Resolved!

ferdnyc:
Resolved again!

jonoomph:
LGTM!

PR #344, opened 2019-10-08
CMake: REQUIRE OpenMP, use (or create) targets for dependency propagation

ferdnyc:
OpenMP is now `REQUIRED` as the build will fail if it's not available. This PR updates the `find_package`, and creates the necessary targets for older CMake versions (3.9+ will do it automatically).

The targets are then made `PUBLIC` dependencies of the `openshot` target, which causes the OpenMP configs to automatically propagate. The redundant OpenMP detection is removed from `tests/CMakeLists.txt`.

The detection code comes verbatim from [this post](https://iscinumpy.gitlab.io/post/omp-on-high-sierra/), which also credits Modern CMake.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -6,8 +6,8 @@
 #
 # Copyright (c) 2008-2019 OpenShot Studios, LLC
 # <http://www.openshotstudios.com/>. This file is part of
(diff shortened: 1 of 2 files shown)


jonoomph:
Nice find! LG

PR #310, opened 2019-08-14
WIP: Use CTest to run tests

ferdnyc:
This PR moves to running our tests with the CTest component of CMake, integrating them better into the CMake environment.

To help facilitate this, the 14 different test source files are compiled into 14 different executables, instead of just a single `openshot-test` program which runs all of them. They can then be run as a unit using `make test`, or selectively with `ctest -R $include_regexp` or `ctest -E $exclude_regexp`

```console
$ make test
Running tests...
Test project /home/ferd/rpmbuild/REPOS/libopenshot/build_ctest
      Start  1: Cache
 1/14 Test  #1: Cache ............................   Passed    0.36 sec
      Start  2: Clip
 2/14 Test  #2: Clip .............................   Passed    0.68 sec
      Start  3: Color
 3/14 Test  #3: Color ............................   Passed    0.08 sec
      Start  4: Coordinate
 4/14 Test  #4: Coordinate .......................   Passed    0.07 sec
      Start  5: ReaderBase
 5/14 Test  #5: ReaderBase .......................   Passed    0.07 sec
      Start  6: ImageWriter
 6/14 Test  #6: ImageWriter ......................   Passed    2.91 sec
      Start  7: FFmpegReader
 7/14 Test  #7: FFmpegReader .....................   Passed    2.08 sec
      Start  8: FFmpegWriter
 8/14 Test  #8: FFmpegWriter .....................   Passed    1.33 sec
      Start  9: Fraction
 9/14 Test  #9: Fraction .........................   Passed    0.07 sec
      Start 10: FrameMapper
10/14 Test #10: FrameMapper ......................   Passed    0.65 sec
      Start 11: KeyFrame
11/14 Test #11: KeyFrame .........................   Passed    0.39 sec
      Start 12: Point
12/14 Test #12: Point ............................   Passed    0.07 sec
      Start 13: Settings
13/14 Test #13: Settings .........................   Passed    0.07 sec
      Start 14: Timeline
14/14 Test #14: Timeline .........................   Passed    0.84 sec

100% tests passed, 0 tests failed out of 14

Total Test time (real) =   9.67 sec

$ ctest -R Writer
Test project /home/ferd/rpmbuild/REPOS/libopenshot/build_ctest
    Start 6: ImageWriter
1/2 Test #6: ImageWriter ......................   Passed    2.90 sec
    Start 8: FFmpegWriter
2/2 Test #8: FFmpegWriter .....................   Passed    1.33 sec

100% tests passed, 0 tests failed out of 2

[...]

PR #345, opened 2019-10-09
DynamicEffects: Add code for Linux with tests

kirillmakhonin:
This PR adds possibility to develop and integrate effects for libopenshot as a shared library.
Right now it works only for Linux, on other platform this method does not do anything (returns NULL).

As an example of "plugin" I've copied Blur effect with name SuperBlur.

From user-perspective, this feature is needed to have an ability create granular "plugins" w/o entire libopenshot compilation.

Later, this system needs to be improved to:

1. Make an ability to contain more then one effect in library
2. Add version check during plugin loading (to be sure that plugin was developed for correct version of libopenshot)
3. Add after-loading of plugins if path to them is provided in env. variable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,5 @@
 .project
 .cproject
 /.metadata/
+/plugins*
 *~
--- a/external-effects/super-blur/CMakeLists.txt
+++ b/external-effects/super-blur/CMakeLists.txt
@@ -27,28 +27,6 @@
 # Pick up our include directories from the parent context
 include_directories(${OPENSHOT_INCLUDE_DIRS})
 
-SET(TEST_MEDIA_PATH "${PROJECT_SOURCE_DIR}/src/examples/")
-SET(TEST_PLUGINS_PATH "${PROJECT_SOURCE_DIR}/plugins/")
-add_definitions( -DTEST_PLUGINS_PATH="${TEST_PLUGINS_PATH}" )
-
-################ WINDOWS ##################
-# Set some compiler options for Windows
-# required for libopenshot-audio headers
-IF (WIN32)
-	STRING(REPLACE "/" "\\\\" TEST_MEDIA_PATH TEST_MEDIA_PATH)
-	add_definitions( -DIGNORE_JUCE_HYPOT=1 )
-	SET(CMAKE_CXX_FLAGS " ${CMAKE_CXX_FLAGS} -include cmath")
-ENDIF(WIN32)
-
-add_definitions( -DTEST_MEDIA_PATH="${TEST_MEDIA_PATH}" )
-
-################### UNITTEST++ #####################
-# Find UnitTest++ libraries (used for unit testing)
-FIND_PACKAGE(UnitTest++ REQUIRED)
-
-# Include UnitTest++ headers (needed for compile)
-include_directories(${UNITTEST++_INCLUDE_DIR})
-
 ################ IMAGE MAGICK ##################
 # Set the Quantum Depth that ImageMagick was built with (default to 16 bits)
 IF (MAGICKCORE_QUANTUM_DEPTH)
@@ -159,38 +137,25 @@ if (USE_SYSTEM_JSONCPP)
 	find_package(JsonCpp REQUIRED)
 	include_directories(${JSONCPP_INCLUDE_DIRS})
 else()
-	include_directories("../thirdparty/jsoncpp")
+	include_directories("../../thirdparty/jsoncpp")
 endif(USE_SYSTEM_JSONCPP)
 
-IF (NOT DISABLE_TESTS)
-	###############  SET TEST SOURCE FILES  #################
-	SET ( OPENSHOT_TEST_FILES tests.cpp
-		   Cache_Tests.cpp
-		   Clip_Tests.cpp
-		   Color_Tests.cpp
-		   Coordinate_Tests.cpp
-           DynamicEffects_Tests.cpp
-		   ReaderBase_Tests.cpp
-		   ImageWriter_Tests.cpp
-		   FFmpegReader_Tests.cpp
-		   FFmpegWriter_Tests.cpp
-		   Fraction_Tests.cpp
-		   FrameMapper_Tests.cpp
-		   KeyFrame_Tests.cpp
-		   Point_Tests.cpp
-		   Settings_Tests.cpp
-		   Timeline_Tests.cpp )
+SET ( EFFECT_SOURCE_FILES
+		SuperBlur.cpp)
 
-	################ TESTER EXECUTABLE #################
-	# Create unit test executable (openshot-test)
-	add_executable(openshot-test
-				   tests.cpp
-				   ${OPENSHOT_TEST_FILES} )
+IF (NOT DISABLE_EXTERNAL_EFFECTS)
+	# Create shared openshot library
+	add_library(libeffectsuperblur SHARED
+			${EFFECT_SOURCE_FILES})
 
-	# Link libraries to the new executable
-	target_link_libraries(openshot-test openshot ${UNITTEST++_LIBRARY} libeffectsuperblur)
+	set(LIB_INSTALL_DIR libeffect${LIB_SUFFIX}) # determine correct lib folder
 
-	#################### MAKE TEST ######################
-	# Hook up the 'make os_test' target to the 'openshot-test' executable
-	ADD_CUSTOM_TARGET(os_test COMMAND openshot-test)
-ENDIF (NOT DISABLE_TESTS)
+	add_custom_command(
+			TARGET libeffectsuperblur
+			POST_BUILD
+			COMMAND ${CMAKE_COMMAND} -E copy
+			$<TARGET_FILE:libeffectsuperblur>
+			${PROJECT_SOURCE_DIR}/plugins/$<TARGET_FILE_NAME:libeffectsuperblur>
+	)
+
+ENDIF (NOT DISABLE_EXTERNAL_EFFECTS)
--- a/external-effects/super-blur/SuperBlur.h
+++ b/external-effects/super-blur/SuperBlur.h
@@ -1,0 +1,130 @@
+/**
+ * @file
+ * @brief Header file for SuperBlur effect class
+ * @author Jonathan Thomas <jonathan@openshot.org>
+ *
+ * @ref License
+ */
+
+/* LICENSE
+ *
+ * Copyright (c) 2008-2019 OpenShot Studios, LLC
+ * <http://www.openshotstudios.com/>. This file is part of
+ * OpenShot Library (libopenshot), an open-source project dedicated to
+ * delivering high quality video editing and animation solutions to the
+ * world. For more information visit <http://www.openshot.org/>.
+ *
+ * OpenShot Library (libopenshot) is free software: you can redistribute it
+ * and/or modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * OpenShot Library (libopenshot) is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with OpenShot Library. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef OPENSHOT_EXTERNAL_BLUR_EFFECT_H
+#define OPENSHOT_EXTERNAL_BLUR_EFFECT_H
+
+#include "../../include/EffectBase.h"
+
+#include <cmath>
+#include <ctime>
+#include <iostream>
+#include <omp.h>
+#include <stdio.h>
+#include <memory>
+#include <QRect>
+#include "../../include/Color.h"
+#include "../../include/Exceptions.h"
+#include "../../include/Json.h"
+#include "../../include/KeyFrame.h"
+#include "../../include/ReaderBase.h"
+#include "../../include/FFmpegReader.h"
+#include "../../include/QtImageReader.h"
+#include "../../include/ChunkReader.h"
+
+using namespace std;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void * factory();
+
+#ifdef __cplusplus
+}
+#endif
+
+
+namespace openshot
+{
+
+	/**
+	 * @brief This class adjusts the blur of an image, and can be animated
+	 * with openshot::Keyframe curves over time.
+	 *
+	 * Adjusting the blur of an image over time can create many different powerful effects. To achieve a
+	 * box blur effect, use identical horizontal and vertical blur values. To achieve a Gaussian blur,
+	 * use 3 iterations, a sigma of 3.0, and a radius between 3 and X (depending on how much blur you want).
+	 */
+	class SuperBlur : public EffectBase
+	{
+	private:
+		/// Init effect settings
+		void init_effect_details();
+
+		/// Internal blur methods (inspired and credited to http://blog.ivank.net/fastest-gaussian-blur.html)
+		int* initBoxes(float sigma, int n);
+		void boxBlurH(unsigned char *scl, unsigned char *tcl, int w, int h, int r);
+		void boxBlurT(unsigned char *scl, unsigned char *tcl, int w, int h, int r);
+
+
+	public:
+		Keyframe horizontal_radius;	///< Horizontal blur radius keyframe. The size of the horizontal blur operation in pixels.
+		Keyframe vertical_radius;	///< Vertical blur radius keyframe. The size of the vertical blur operation in pixels.
+		Keyframe sigma;				///< Sigma keyframe. The amount of spread in the blur operation. Should be larger than radius.
+		Keyframe iterations;		///< Iterations keyframe. The # of blur iterations per pixel. 3 iterations = Gaussian.
+
+		/// Blank constructor, useful when using Json to load the effect properties
+        SuperBlur();
+
+		/// Default constructor, which takes 1 curve. The curve adjusts the blur radius
+		/// of a frame's image.
+		///
+		/// @param new_horizontal_radius The curve to adjust the horizontal blur radius (between 0 and 100, rounded to int)
+		/// @param new_vertical_radius The curve to adjust the vertical blur radius (between 0 and 100, rounded to int)
+		/// @param new_sigma The curve to adjust the sigma amount (the size of the blur brush (between 0 and 100), float values accepted)
+		/// @param new_iterations The curve to adjust the # of iterations (between 1 and 100)
+        SuperBlur(Keyframe new_horizontal_radius, Keyframe new_vertical_radius, Keyframe new_sigma, Keyframe new_iterations);
+
+		/// @brief This method is required for all derived classes of EffectBase, and returns a
+		/// modified openshot::Frame object
+		///
+		/// The frame object is passed into this method, and a frame_number is passed in which
+		/// tells the effect which settings to use from its keyframes (starting at 1).
+		///
+		/// @returns The modified openshot::Frame object
+		/// @param frame The frame object that needs the effect applied to it
+		/// @param frame_number The frame number (starting at 1) of the effect on the timeline.
+		std::shared_ptr<Frame> GetFrame(std::shared_ptr<Frame> frame, int64_t frame_number);
+
+		/// Get and Set JSON methods
+		string Json(); ///< Generate JSON string of this object
+		void SetJson(string value); ///< Load JSON string into this object
+		Json::Value JsonValue(); ///< Generate Json::JsonValue for this object
+		void SetJsonValue(Json::Value root); ///< Load Json::JsonValue into this object
+
+		/// Get all properties for a specific frame (perfect for a UI to display the current state
+		/// of all properties at any time)
+		string PropertiesJSON(int64_t requested_frame);
+	};
+
+}
+
+#endif
--- a/include/EffectInfo.h
+++ b/include/EffectInfo.h
@@ -32,6 +32,14 @@
 #define OPENSHOT_EFFECT_INFO_H
 
 #include "Effects.h"
+#include <vector>
+#include <map>
+#include <pthread.h>
+
+
+#if defined(__linux__)
+#include <dlfcn.h>
+#endif
 
 
 using namespace std;
@@ -51,9 +59,18 @@ namespace openshot
 		// Create an instance of an effect (factory style)
 		EffectBase* CreateEffect(string effect_type);
 
+		// Load effect built in shared library (factory style)
+		EffectBase* LoadEffect(string location);
+
+		// Unload all effect loaded dynamically
+		void UnloadDynamicEffects();
+
 		/// JSON methods
 		static string Json(); ///< Generate JSON string of this object
 		static Json::Value JsonValue(); ///< Generate Json::JsonValue for this object
+
+    private:
+        pthread_mutex_t		m_mutex;
 
 	};
 
--- a/src/EffectInfo.cpp
+++ b/src/EffectInfo.cpp
@@ -30,9 +30,9 @@
 
 #include "../include/EffectInfo.h"
 
-
 using namespace openshot;
-
+static map<string, EffectBase*(*)()> m_loadedDynamicEffects;
+static vector<void*> m_loadedDynamicHandles;
 
 // Generate JSON string of this object
 string EffectInfo::Json() {
@@ -43,6 +43,11 @@ string EffectInfo::Json() {
 
 // Create a new effect instance
 EffectBase* EffectInfo::CreateEffect(string effect_type) {
+    // Try to find dynamically loaded effect
+    auto effect_factory = m_loadedDynamicEffects.find(effect_type);
+    if (effect_factory != m_loadedDynamicEffects.end())
+        return effect_factory->second();
+
 	// Init the matching effect object
 	if (effect_type == "Bars")
 		return new Bars();
@@ -88,6 +93,52 @@ EffectBase* EffectInfo::CreateEffect(string effect_type) {
 	return NULL;
 }
 
+#if defined(__linux__)
+EffectBase* EffectInfo::LoadEffect(string location){
+    pthread_mutex_lock(&m_mutex);
+
+    void * file = dlopen(location.c_str(), RTLD_NOW);
+    if (file == nullptr){
+        pthread_mutex_unlock(&m_mutex);
+        return nullptr;
+    }
+    void * factory_ptr = dlsym(file, "factory");
+    if (factory_ptr == nullptr){
+        pthread_mutex_unlock(&m_mutex);
+        return nullptr;
+    }
+
+    EffectBase * (*factory)();
+    factory = (EffectBase* (*)()) factory_ptr;
+
+    EffectBase* instance = factory();
+    m_loadedDynamicEffects.insert(make_pair(instance->info.name, factory));
+    m_loadedDynamicHandles.insert(m_loadedDynamicHandles.end(), file);
+
+    pthread_mutex_unlock(&m_mutex);
+    return instance;
+}
+
+void EffectInfo::UnloadDynamicEffects(){
+    pthread_mutex_lock(&m_mutex);
+    for (auto & handle : m_loadedDynamicHandles){
+        dlclose(handle);
+    }
+
+    m_loadedDynamicEffects.clear();
+    m_loadedDynamicHandles.clear();
+    pthread_mutex_unlock(&m_mutex);
+}
+#else
+EffectBase* EffectInfo::LoadEffect(string location){
+    return NULL;
+}
+
+void EffectInfo::UnloadDynamicEffects(){
+
+}
+#endif
+
 // Generate Json::JsonValue for this object
 Json::Value EffectInfo::JsonValue() {
 
@@ -109,6 +160,12 @@ Json::Value EffectInfo::JsonValue() {
 	root.append(Saturation().JsonInfo());
 	root.append(Shift().JsonInfo());
 	root.append(Wave().JsonInfo());
+
+	for (auto & effect : m_loadedDynamicEffects){
+	    auto instance = effect.second();
+	    root.append(instance->JsonInfo());
+	    delete instance;
+	}
 
 	// return JsonValue
 	return root;
--- a/tests/DynamicEffects_Tests.cpp
+++ b/tests/DynamicEffects_Tests.cpp
@@ -1,0 +1,93 @@
+/**
+ * @file
+ * @brief Unit tests for openshot::Timeline
+ * @author Jonathan Thomas <jonathan@openshot.org>
+ *
+ * @ref License
+ */
+
+/* LICENSE
+ *
+ * Copyright (c) 2008-2019 OpenShot Studios, LLC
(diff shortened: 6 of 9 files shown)


ferdnyc:
@kirillmakhonin Wow! This looks _really_ interesting, I'm excited to check it out. I love the idea of making libopenshot more modular, period, and breaking out the effects seems like one of the more natural ways to do that. I can even imagine porting _all_ of our built-in effects to build as runtime-loaded modules, perhaps, and doing away with the "built-in vs. external" divide entirely.

But for now, this sounds like a perfect first step. I'll give it a spin and post more feedback once I've had a chance to digest it more fully, and then hopefully a review some time in the next couple of days.

Some thoughts on your improvements list, for starters:
> Later, this system needs to be improved to:
> 
> 1. Make an ability to contain more then one effect in library

Eh... I mean, maybe? If there are effects  that might benefit from sharing code, I can see there being an advantage to that. (But it's not like _any_ of our EXISTING effects share even a single line of code, as things stand today.) Unless there's some practical advantage to building multi-effect libraries, though, my knee-jerk inclination is to keep each one in its own module. Linux's libdl, at least, is pretty efficient about loading multiple objects. But maybe I'll reconsider that or be talked out of it later, currently it's just a snap judgement.

> 2. Add version check during plugin loading (to be sure that plugin was developed for correct version of libopenshot)

Yeah, versioning will definitely be important here — potentially of _both_ the source (versioned for API changes) _and_ the objects. (libopenshot should probably require that the objects export some build-ident signature that matches its own copy, so it only loads modules if they were compiled for that _exact_ `libopenshot.so`. No copying effect modules from a different install or anything, at least not in the beginning.)

> 3. Add after-loading of plugins if path to them is provided in env. variable

Best thing might be for us to define a Settings variable that holds the plugin path, that'll provide callers with an already-established API for configuring it. And supporting an envvar for runtime overrides (or additions?) makes sense as well.

Some other possible items to consider:

4. I think we'll need a way to generate at least Python bindings for these, as well, if they're to be usable within OpenShot itself (which would be ideal).

    I've been toying with the idea of moving the effects into their own `openshot::effect` namespace, and breaking them out into a separate Python submodule. (Simply because the SWIG-generated C++ source for the Python bindings is a single `.cpp` file that currently clocks in at 2.3 MEGABYTES, which is just insane. The effects seemed the most obvious and natural place to start breaking that apart a bit.) This might just be one more reason to do that, because if the effects are `openshot.effect.Blur` / `openshot.effect.Crop`, etc., then it's easy enough to create SWIG interface files for more modules. We'd have to figure out the logistics of loading those, as well.

    Or, the radical alternative would be to _not_ do that, and instead eliminate the public APIs of the individual effects entirely. Instead we define a full API for accessing effects via libopenshot, so calls are made in a structured way, and the same calls provide access to every currently-loaded effect. ...But that's a lot more effort and a lot more disruption.

5. And, yeah, we'll probably have to figure out MacOS and Windows as well, to make this ready for prime-time. My understanding is, people still use them, for whatever reason.


Sorry for the brain-dump, I'll probably have more focused feedback once I've tried this out and gotten more of a hands-on feel for it. But thanks again for the contribution, this really seems like it'll be an awesome evolutionary next-step for libopenshot!

.gitignore
```diff
@@ -6,4 +6,5 @@
 .project
 .cproject
 /.metadata/
+/plugins*
```
ferdnyc: Drop this from the PR, please — you can add local, non-shared exclusions to the `.git/info/exclude` file in your clone, that way they're separate from the shared repo files. (I only just recently discovered that in the `gitignore` manpage myself.)

external-effects/super-blur/CMakeLists.txt
```diff
@@ -0,0 +1,161 @@
...
+else()
+	include_directories("../../thirdparty/jsoncpp")
+endif(USE_SYSTEM_JSONCPP)
+
+SET ( EFFECT_SOURCE_FILES
+		SuperBlur.cpp)
+
+IF (NOT DISABLE_EXTERNAL_EFFECTS)
+	# Create shared openshot library
+	add_library(libeffectsuperblur SHARED
```
ferdnyc: If you name this target `effectsuperblur`, it will compile as `libeffectsuperblur.so` and you won't have to refer to it as `liblib...` elsewhere. CMake automatically adds the `lib` prefix to library targets.

(Also, maybe a hyphen or underscore in the name? Something like `effect-superblur` => `libeffect-superblur.so`? Just for readability.)

external-effects/super-blur/CMakeLists.txt
```diff
@@ -0,0 +1,161 @@
...
+	find_package(JsonCpp REQUIRED)
+	include_directories(${JSONCPP_INCLUDE_DIRS})
+else()
+	include_directories("../../thirdparty/jsoncpp")
+endif(USE_SYSTEM_JSONCPP)
+
+SET ( EFFECT_SOURCE_FILES
+		SuperBlur.cpp)
+
+IF (NOT DISABLE_EXTERNAL_EFFECTS)
```
ferdnyc: This check should go into the `external-effects/CMakeLists.txt` file instead, so the `add_subdirectory(super-blur)` is never even executed when external effects are disabled.

external-effects/super-blur/CMakeLists.txt
```diff
@@ -0,0 +1,161 @@
...
+
+	set(LIB_INSTALL_DIR libeffect${LIB_SUFFIX}) # determine correct lib folder
+
+	add_custom_command(
+			TARGET libeffectsuperblur
+			POST_BUILD
+			COMMAND ${CMAKE_COMMAND} -E copy
+			$<TARGET_FILE:libeffectsuperblur>
+			${PROJECT_SOURCE_DIR}/plugins/$<TARGET_FILE_NAME:libeffectsuperblur>
+	)
```
ferdnyc: Is this... copying the _shared library_ back into the source dir?!? Why? Is it just so the tests can load the effect plugin `.so`, later?

There's no need to pollute the source tree for that, the `CMakeLists.txt` for the unit tests can provide a compile-time definition for the _output_ location for the external plugin, in the build tree. (`${PROJECT_BINARY_DIR}/external-effects/super-blur/` in this case.) The tests can load it right from there, instead.

tests/CMakeLists.txt
```diff
@@ -185,7 +188,7 @@ IF (NOT DISABLE_TESTS)
 				   ${OPENSHOT_TEST_FILES} )
 
 	# Link libraries to the new executable
-	target_link_libraries(openshot-test openshot ${UNITTEST++_LIBRARY})
+	target_link_libraries(openshot-test openshot ${UNITTEST++_LIBRARY} libeffectsuperblur)
```
ferdnyc: Hmm. If your unit tests are going to be testing the dynamic loading capability for effect plugins (and they probably should), I don't think you want to link the test target with the plugin target.

You may even want to break these up into two separate tests: an unlinked one that just tests libopenshot's dynamic loading features, then another that's linked directly with the plugin itself, and tests its internal APIs.

external-effects/super-blur/SuperBlur.h
```diff
@@ -0,0 +1,130 @@
...
+#include "../../include/Color.h"
+#include "../../include/Exceptions.h"
+#include "../../include/Json.h"
+#include "../../include/KeyFrame.h"
+#include "../../include/ReaderBase.h"
+#include "../../include/FFmpegReader.h"
+#include "../../include/QtImageReader.h"
+#include "../../include/ChunkReader.h"
+
+using namespace std;
```
ferdnyc: Remove this, please, and prefix all namespaced symbols in header files. See #296 .

external-effects/super-blur/SuperBlur.h
```diff
@@ -0,0 +1,130 @@
...
+		/// The frame object is passed into this method, and a frame_number is passed in which
+		/// tells the effect which settings to use from its keyframes (starting at 1).
+		///
+		/// @returns The modified openshot::Frame object
+		/// @param frame The frame object that needs the effect applied to it
+		/// @param frame_number The frame number (starting at 1) of the effect on the timeline.
+		std::shared_ptr<Frame> GetFrame(std::shared_ptr<Frame> frame, int64_t frame_number);
+
+		/// Get and Set JSON methods
+		string Json(); ///< Generate JSON string of this object
```
ferdnyc: This should be fully specified as `std::string` (and throughout).

include/EffectInfo.h
```diff
@@ -51,10 +59,19 @@ namespace openshot
 		// Create an instance of an effect (factory style)
 		EffectBase* CreateEffect(string effect_type);
 
+		// Load effect built in shared library (factory style)
+		EffectBase* LoadEffect(string location);
```
ferdnyc: Since the `using namespace std;` will be going away in #296, this should also be `std::string` (and throughout).

src/EffectInfo.cpp
```diff
@@ -88,6 +93,52 @@ EffectBase* EffectInfo::CreateEffect(string effect_type) {
...
+
+    void * file = dlopen(location.c_str(), RTLD_NOW);
+    if (file == nullptr){
+        pthread_mutex_unlock(&m_mutex);
+        return nullptr;
+    }
+    void * factory_ptr = dlsym(file, "factory");
+    if (factory_ptr == nullptr){
+        pthread_mutex_unlock(&m_mutex);
+        return nullptr;
```
ferdnyc: Shouldn't `file` be closed (er, `dlclose()`-d) before returning, here?

tests/CMakeLists.txt
ferdnyc: (Note that the above would require multiple separate test executables, which is something I've experimented with moving to anyway for various reasons. See #310 for the WIP stab at that.)

external-effects/super-blur/CMakeLists.txt
ferdnyc: (In fact, it could also go into the root CMakeLists.txt, and prevent the `add_subdirectory(external-effects)` from even happening. I have no strong preference on that one. But it should definitely be wrapped around an `add_subdirectory()` call somewhere, so that CMake never even descends into that dir if there's ultimately nothing for it to do there.

src/EffectInfo.cpp
kirillmakhonin: I guess not, due to [documentation](https://linux.die.net/man/3/dlclose) if we call `dlclose`, this function (`factory`) will be unloaded.
In our case `dlclose` is being called only when you call `EffectInfo.UnloadDynamicEffects()`.

.gitignore
kirillmakhonin: Done

external-effects/super-blur/CMakeLists.txt
kirillmakhonin: Done

external-effects/super-blur/CMakeLists.txt
kirillmakhonin: Done

external-effects/super-blur/CMakeLists.txt
kirillmakhonin: Removed, fixed

tests/CMakeLists.txt
kirillmakhonin: Did it switchable. What do you think?

external-effects/super-blur/SuperBlur.h
kirillmakhonin: Done

external-effects/super-blur/SuperBlur.h
kirillmakhonin: Done

include/EffectInfo.h
kirillmakhonin: Done

kirillmakhonin:
@ferdnyc, I did some updates, could you please review them?
I did not add support for mac and windows right now, but, compilation and tests should not fail on these platforms.
Also I tried to add some versioning, but I'm sure that my approach is right, maybe `OPENSHOT_VERSION_SO` should be used.

src/EffectInfo.cpp
ferdnyc: > In our case `dlclose` is being called only when you call `EffectInfo.UnloadDynamicEffects()`.

Right, but that only happens if the reference returned by `dlopen()` is placed on the `m_loadedDynamicHandles` list — which **won't happen here**, because you return before reaching line 115.

That means you've got a dangling library handle hanging around, and the shared library won't be unloaded even when you _**do**_ later call `dlclose()`, because the handles are reference-counted and...
> a dynamic library is not deallocated until `dlclose()` has been called on it as many times as `dlopen()` has succeeded on it.

The code in the man page Example section doesn't `dlclose()` in its failure case because it's not calling `return()` — it's calling `exit()`. There's no point in cleaning up open references if you're exiting. As [Raymond Chen wrote](https://devblogs.microsoft.com/oldnewthing/?p=8683), on program exit "the building is being demolished. Don’t bother sweeping the floor and emptying the trash cans and erasing the whiteboards."

But if the code is expected to keep running past that point, then you don't want to be leaking `dlopen()` handles, it makes it impossible to clean up properly later on.

kirillmakhonin: @ferdnyc, yeah, I missed it, fully agree with you. I've added `dlclose()` in places where algorithm decides that this library is incorrect, but lib already has been loaded.

tests/CMakeLists.txt
```diff
@@ -184,8 +185,15 @@ IF (NOT DISABLE_TESTS)
...
-	target_link_libraries(openshot-test openshot ${UNITTEST++_LIBRARY})
+	IF (NOT DISABLE_EXTERNAL_EFFECTS)
+		# Link libraries to the new executable
+		target_link_libraries(openshot-test openshot ${UNITTEST++_LIBRARY} effect-superblur)
+		add_definitions(-DTEST_PLUGIN="${PROJECT_BINARY_DIR}/external-effects/super-blur/libeffect-superblur.so"
+						-DTEST_PLUGIN_NAME="SuperBlur")
+	ELSE()
+		# Link libraries to the new executable
+		target_link_libraries(openshot-test openshot ${UNITTEST++_LIBRARY})
+	ENDIF(NOT DISABLE_EXTERNAL_EFFECTS)
```
ferdnyc: Couple issues here:

1. `DynamicEffects_Tests.cpp` will be in the test target sources list even when `DISABLE_EXTERNAL_EFFECTS` is set. That's _okaaaay_ technically, because the whole file's wrapped in `#ifdef TEST_PLUGIN`, but if nothing else it's an extra source file being compiled for no reason.

2. Bigger issue, if you want to test the `dlopen()` functionality of `EffectInfo::LoadEffect`, I suspect you can't _link_ the executable with the library you're trying to `dlopen()` or it will invalidate the test.

   Because `effect-superblur` is added to the `target_link_libraries()` of `openshot-test`, it'll be dynamically linked with the executable:
   ```console
   $ cd build
   $ make openshot-test
   $ ldd tests/openshot-test |grep -i effect
	libeffect-superblur.so => /home/ferd/rpmbuild/REPOS/libopenshot/build/external-effects/super-blur/libeffect-superblur.so (0x00007f41707c6000)
   ```
   The dynamic loader is sure to be smart enough that, when you call `dlopen()` on that file path, it's not going to load anything at all — it's just going to return a handle to the library that was **already** loaded at the start of program execution.

   To test dynamic-loading from the library file on disk, you _don't_ want to link the executable to it. It should have no knowledge of that library prior to `dlopen()` being given the pathname to load it from.

I guess my preference would be something more like this:
```suggestion
# Link libraries to the new executable
target_link_libraries(openshot-test openshot ${UNITTEST++_LIBRARY})
if (NOT DISABLE_EXTERNAL_EFFECTS)
    target_sources(openshot-test PRIVATE DynamicEffects_Tests.cpp)
    target_compile_definitions(openshot-test PRIVATE
        TEST_PLUGIN="${PROJECT_BINARY_DIR}/external-effects/super-blur/libeffect-superblur.so"
        TEST_PLUGIN_NAME="SuperBlur")
endif ()
```

That'll add in the source file and the definitions iff enabled, and leave the library unlinked so it'll _have_ to be loaded from the path in the `TEST_PLUGIN` definition.

kirillmakhonin: I made compilation of file `DynamicEffects_Tests.cpp` optional, also I remove static linking with effect library.
@ferdnyc could you, please, have a look?

ferdnyc: Wow, you are _OLD_-school CMake, huh? :laughing: Well, that's cool, whatever works — and this does, thanks. One day I'll drag this project kicking and screaming into the [Modern CMake](https://cliutils.gitlab.io/modern-cmake/) world, but that day is not today.

I'm actually a little surprised `add_definitions()` is still effective even when it comes after the creation of the executable target (as opposed to `target_compile_definitions()` operating on the target), but it appears that's the case.

> I made compilation of file `DynamicEffects_Tests.cpp` optional, 

And the `add_dependency()` is a good idea, I missed that.

> also I remove static linking with effect library.

*dynamic linking. (A bit pedantic, I admit, but the distinction is critical. Static linking compiles the library right into the executable and creates no external dependencies. With a shared library the runtime dependency is still dynamic linking — as opposed to dynamic _loading_ via `dlopen()`, which is what you're doing instead, and requires no prior linkage between the executable and the shared object file.)

src/EffectInfo.cpp
```diff
@@ -88,6 +93,62 @@ EffectBase* EffectInfo::CreateEffect(string effect_type) {
...
+#else
+EffectBase* EffectInfo::LoadEffect(std::string location){
+    return NULL;
+}
+
+void EffectInfo::UnloadDynamicEffects(){
+
+}
+#endif
+
```
ferdnyc: This might be cleaner/easier to write as...

```cpp
EffectBase* EffectInfo::LoadEffect(std::string location){
#if !defined(__linux__)
    return NULL;
#endif
    pthread_mutex_lock(&m_mutex);
    ...
}

void EffectInfo::UnloadDynamicEffects(){
#if defined(__linux__)
    pthread_mutex_lock(&m_mutex);

    for (auto & handle : m_loadedDynamicHandles){
        dlclose(handle);
    }

    m_loadedDynamicEffects.clear();
    m_loadedDynamicHandles.clear();
    pthread_mutex_unlock(&m_mutex);
#endif
}
```

Primary advantage: If the signature of `EffectInfo::LoadEffect()` or `EffectInfo::UnloadDynamicEffects()` changes, we don't have to worry about updating it in two places and making sure they're kept in sync.

include/EffectBase.h
```diff
@@ -31,6 +31,8 @@
 #ifndef OPENSHOT_EFFECT_BASE_H
 #define OPENSHOT_EFFECT_BASE_H
 
+#define OPENSHOT_PLUGIN_API_VERSION 0x0001
```
ferdnyc: Yeah... as you mentioned in the main PR comments, I'd just use the SO version instead. (It's `#define`d as `OPENSHOT_VERSION_SO` in `OpenShotVersion.h`.)

The main reason being, we really don't _have_ an **effects** API currently (and that's on us). The abstractions for effects use aren't really great. Callers generally need to have knowledge of the API for the specific effects they use, and to an extent that will also include the dynamic ones. (The JSON interfaces do _somewhat_ alleviate that, but more on that later.)

As a result, the plugin compatibility will most likely be tied to the library API **even if** the plugin-_loading_ API stays the same, and a plugin API versioning scheme just adds another numbering to keep track of.

But if this **IS** retained, it should be placed in `include/OpenShotVersion.h.in` instead of here. (Which is the input for a CMake-generated header file that you can include via `#include "OpenShotVersion.h"`, to access any of the versioning macros.)

tests/DynamicEffects_Tests.cpp
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with OpenShot Library. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "UnitTest++.h"
+#include "../include/OpenShot.h"
+
+#ifdef TEST_PLUGIN
+#if defined(__linux__)
```
ferdnyc: I was going to suggest combining these as
```cpp
#if defined(__linux__) && defined(TEST_PLUGIN)
```
But since in theory we might have plugin-loading implementations for other OSes, leaving them separate may be better. The only thing I'd request is commenting the `#endif`s as:
```cpp
#endif // defined(__linux__)
#endif // TEST_PLUGIN
```
since they're way too far away to match up visually.

tests/DynamicEffects_Tests.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    effect = effect_info.CreateEffect(TEST_PLUGIN_NAME);
+
+    CHECK(effect != nullptr);
+
+    delete effect;
+}
+
+#endif
+#endif
```
ferdnyc: Minor nit: File doesn't end in a newline.

src/EffectInfo.cpp
ferdnyc: (Sorry, accidentally had an earlier version of this comment included in the review, along with the final one. I've deleted the redundant comment, but the email notification version may have been a bit confusing.)

ferdnyc:
@kirillmakhonin 

Surprisingly enough, it _may_ be possible to use the same code even on Windows: 

https://github.com/dlfcn-win32/dlfcn-win32

(`dlopen()` already works natively on MacOS, the only wrinkle is that shared libraries are compiled with the `.dylib` extension, instead of `.so`.)

jonoomph:
@kirillmakhonin This looks very promising! Thanks again for your help implementing it. I'm feeling like we should keep iterating on this idea (in this PR), and before we merge, I would suggest we get this working on all 3 OSes. If I need to install additional things on our Windows build server, please let me know. We can also move this into an official branch, so our build servers can play with it and create test builds.

Also, most effect metadata comes from `openshot.EffectInfo.Json()`, where it is returned as JSON for the OpenShot UI. I see no reason why this method couldn't include metadata from `loaded effects` as well as built-in effects. 

Another thought... how would we handle translations for externally loaded effects. I do love the idea of dropping an effect library into a folder, and launching OpenShot, and when detected, those effect libraries would be loaded, translations loaded, icons loaded, etc... and would be completely seamless to the user. :+1:

ferdnyc:
> @kirillmakhonin This looks very promising! Thanks again for your help implementing it. I'm feeling like we should keep iterating on this idea (in this PR)

Agreed. It's really promising, so it’s worth taking the time to make it a first-class feature that works well for all platforms AND for all consumers. (@jeffski and/or @DylanC may have some thoughts on this, perhaps?)

> Also, most effect metadata comes from `openshot.EffectInfo.Json()`, where it is returned as JSON for the OpenShot UI. I see no reason why this method couldn't include metadata from `loaded effects` as well as built-in effects.

\*nod\* And that's a help, but still feels incomplete to me. The JSON interface to properties works for OpenShot, but it's less convenient for direct `libopenshot` consumers.

One of the things I was going to suggest as far as this PR goes, for the unit testing, was to reimplement the Blur tests from `Timeline_Tests.cpp` in SuperBlur:

https://github.com/OpenShot/libopenshot/blob/76725d9d2e355b9e5e12c54a47eb929ba06e2ada/tests/Timeline_Tests.cpp#L416-L445

If you try to actually do that, you immediately run into the snag that, because dynamic effects use the factory pattern to instantiate the effect objects, you lose access to the effect's custom constructors. This isn't how you'd create a SuperBlur object:
```cpp
Keyframe horizontal_radius(5.0);
Keyframe vertical_radius(5.0);
Keyframe sigma(3.0);
Keyframe iterations(3.0);
Blur blur(horizontal_radius, vertical_radius, sigma, iterations);
```

Now, sure, you could use JSON to serialize and apply the appropriate properties, and that works fine for OpenShot, but for libopenshot consumers I think that probably would feel overly complicated and tedious.

What might be better is if we come up with a standard setter/getter interface for effect parameters, and we implement that interface in the `EffectBase` class. It'd be nice to provide an interface for C++ callers that doesn't hardcode the parameters _themselves_ the way the custom constructors do. Then, all of the dynamic effects (or even all of the _effects_, since there's no reason the static effects can't benefit from this as well) can just implement those setters/getters for their properties, instead of being configured through constructors that require callers to have the specific effect's individual unique interface fully defined at compile time.

That'd also make the bindings for effects far less complex. To the point where it _might_ actually be possible to completely drop the SWIG-generated interface for each effect in favor of `EffectBase` setters/getters being the supported interface for direct callers, even with the built-in static effects. (OpenShot already doesn't use the `openshot.effect.{Foo,Bar,Baz}` Python classes for anything; it works exclusively through the JSON interface.)

> 
> Another thought... how would we handle translations for externally loaded effects. I do love the idea of dropping an effect library into a folder, and launching OpenShot, and when detected, those effect libraries would be loaded, translations loaded, icons loaded, etc... and would be completely seamless to the user.

Mmm, I hadn't even thought about translations. ...How do we handle translations for built-in effects? We **don't**, in libopenshot. It's not translated at all. Because of the JSON interface, all parameters already have string names, and those names are used by OpenShot to assemble the translations.

OpenShot queries the list of effects in the Effects model using `openshot.EffectInfo.Json()`, then applies translations while it's populating the model. To get the strings translated in the first place, `src/language/generate_translations.py` actually does essentially the same thing, loading each effect's properties and metadata and adding the strings to the translation master:

```python
# Append on properties from libopenshot
objects = [openshot.Clip(), openshot.Bars(), openshot.Blur(), openshot.Brightness(),
           openshot.ChromaKey(), openshot.ColorShift(), openshot.Crop(), openshot.Deinterlace(), openshot.Hue(), openshot.Mask(),
           openshot.Negate(), openshot.Pixelate(), openshot.Saturation(), openshot.Shift(), openshot.Wave()]

# Loop through each libopenshot object
for object in objects:
    props = json.loads(object.PropertiesJSON(1))

    # Loop through props
    for key in props.keys():
        object = props[key]
        if "name" in object.keys():
            effects_text[object["name"]] = "libopenshot (Clip Properties)"
        if "choices" in object.keys():
            for choice in object["choices"]:
                effects_text[choice["name"]] = "libopenshot (Clip Properties)"

# Append Effect Meta Data
e = openshot.EffectInfo()
props = json.loads(e.Json())

# Loop through props
for effect in props:
    if "name" in effect:
        effects_text[effect["name"]] = "libopenshot (Effect Metadata)"
    if "description" in effect:
        effects_text[effect["description"]] = "libopenshot (Effect Metadata)"
```

In one sense you could say that this means nothing's changed, so there's no need to worry about it... _in libopenshot_. But in another sense, it really complicates things. Translations being managed completely external to `libopenshot` means that this **CAN'T** be solved entirely from _within_ libopenshot. There's no code that can be added to the dynamic translations and exposed through the effects interface that can solve this, the way things are designed today. OpenShot has to have prior knowledge of every effect loaded.

Which does mean that, if we were to have "built-in" dynamic effects (ones bundled with libopenshot, that OpenShot therefore _does_ have prior knowledge of), translations should still work the same as always. This PR already updates `src/EffectInfo.cpp` so that `openshot.EffectInfo.JSON()` will include the dynamic effects in its listing. As long as they're loaded when `generate_translations.py` makes its query, the strings will be included in the translation master.

For "third-party" effects, things become significantly trickier. I have no doubt that Qt's QTranslator system is modular enough to handle this situation, but we have to figure out the best way to make use of it.

If we're lucky, _most_ of the names of parameters and things should be the same across all effects, so we may be able to reuse the strings already defined. To facilitate that, we may want to define a common [translation context](https://doc.qt.io/qt-5/i18n-source-translation.html#defining-a-translation-context) for all of the property names, possibly associated with either `EffectBase` or `EffectInfo`. (Since we're using `gettext` rather than Qt's `lupdate`, as the tool that collects and manages translation data, we'll have to figure out how — or if — that would work.)

Even still there are still at least some text strings — the name of the effect itself, its description, etc. — that won't fall into that shared-reuse category. Those would need to be translated individually for each effect. (Users do tend to be somewhat more accepting of strings like that being _un_-translated, though, if it comes down to it. As long as the actual user interface can be understood.)

There's a possibility we could just look for a `.qm` file (Qt compiled translation dictionary) alongside each effect plugin, and load that file into the translation system if it's present, without really concerning ourselves with exactly how the contents are created in the first place.

Or, we could have each plugin incorporate its translation data as a binary blob, and use the <code>QTranslator::load(const uchar *<var>data</var>, int <var>len</var>, const QString &<var>directory</var>)</code> signature to load that data. That makes things _significantly_ more complicated from the POV of effect plugin creation, but has the advantage that everything remains self-contained in a single shared library file. (The effect's icon could even be handled the same way.)

Or... something else? `¯\_(ツ)_/¯`

ferdnyc:
(I resolved the merge conflicts in `tests/CMakeLists.txt`, BTW. That's what that commit was earlier.)

ferdnyc:
> If we're lucky, _most_ of the names of parameters and things should be the same across all effects, so we may be able to reuse the strings already defined. To facilitate that, we may want to define a common [translation context](https://doc.qt.io/qt-5/i18n-source-translation.html#defining-a-translation-context) for all of the property names, possibly associated with either `EffectBase` or `EffectInfo`. (Since we're using `gettext` rather than Qt's `lupdate`, as the tool that collects and manages translation data, we'll have to figure out how — or if — that would work.)

Actually, scratch that. Because of the way they're collected now, all properties already have a common translation context. From `src/language/OpenShot/OpenShot.pot`:

```pot
#: /home/jonathan/apps/openshot-qt-git/src/windows/views/timeline_webview.py:2408
#: libopenshot (Clip Properties)
msgid "Brightness"
msgstr ""

#: /home/jonathan/apps/openshot-qt-git/src/windows/views/timeline_webview.py:2410
#: libopenshot (Clip Properties)
msgid "Contrast"
msgstr ""
```

So, dynamically-loaded effects would automatically be handled using that same context, and all of the translated strings from existing properties would be available for use with the dynamic effects' properties as well.

ferdnyc:
Actually, an even better example is:
```pot
#: libopenshot (Effect Metadata)
msgid "Brightness & Contrast"
msgstr ""

#: libopenshot (Clip Properties)
msgid "Sigma"
msgstr ""

#: libopenshot (Clip Properties)
msgid "Duration"
msgstr ""

#: libopenshot (Effect Metadata)
msgid "Color Saturation"
msgstr ""
```

No source-code contexts at all. (But it seems like strings will be reused for other contexts even if they **are** defined with a source context. For instance, the `"Brightness"` in my previous comment is the only instance of that string in the `.pot` file, even though at least some effects' metadata will _also_ use the string "Brightness".)

ferdnyc:
Note to self, regarding translations generally: https://pypi.org/project/distutils_ui/

(Helpers that can be used in a `distutils`-built and -installed Python project to manage translations as `.ts` files created by PyQt5's `pylupdate`, instead of using gettext.)

jonoomph:
@ferdnyc I love the idea of bundling a .qm file alongside an effect library. Although, it would also be great if that could be included in the library (not sure how though), like a resource file compiled into the library... hmmm. 

Like you said, the shared context of OpenShot would still be used to translate strings (if the words have already been translated), but maybe an optional *.qm file would be loaded (if found), and that could be used to fill in the remaining translation gaps in an external effect.

If we really implement this correctly, this could end up being a huge deal for OpenShot... and other open-source projects which use libopenshot. The ability to quickly and easily create an effect seems sooo important.  We could also make an `Effects Download` website, which could incorporate our build servers (for cross-platform building), and simplify the `installation`... although this is an entirely out-of-scope idea right now, and we should probably not get sidetracked thinking about that. :stuck_out_tongue_winking_eye:

musteresel:
I've seen this uses `dlopen()`; regarding portability there's [libltdl](https://www.gnu.org/software/libtool/manual/libtool.html#Using-libltdl) which is basically a portable dlopen library. Might be overkill, though. And I'm not sure about whether the licence would be an issue..

PhysSong:
Since `libopenshot` already uses Qt for image processing, it's also possible to use `QLibrary` and `QMutex` which come from Qt.

Issue #327, opened 2019-09-14
Memory increasing in TextReader


jeffski:
When I run the code below memory usage just keeps increasing, it will keep increasing until the system crashes. On a 8GB machine it hits 45% with just 15 clips. By comparison I can create videos with 40-50 QtImageReader clips, if not more, without a problem.

```
Timeline timeline(1280, 720, Fraction(25,1), 44100, 2, LAYOUT_STEREO);
timeline.color = Color("#000000");

Clip clip1(new TextReader(1280, 720, 0, 0, GRAVITY_CENTER, "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor", "Ubuntu", 24, "#ffffff", "#000000"));
clip1.Position(0);
clip1.Layer(1);
clip1.Start(0);
clip1.End(3);

Clip clip2(new TextReader(1280, 720, 0, 0, GRAVITY_CENTER, "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor", "Ubuntu", 24, "#ffffff", "#000000"));
clip2.Position(4);
clip2.Layer(1);
clip2.Start(0);
clip2.End(3);

Clip clip3(new TextReader(1280, 720, 0, 0, GRAVITY_CENTER, "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor", "Ubuntu", 24, "#ffffff", "#000000"));
clip3.Position(8);
clip3.Layer(1);
clip3.Start(0);
clip3.End(3);

Clip clip4(new TextReader(1280, 720, 0, 0, GRAVITY_CENTER, "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor", "Ubuntu", 24, "#ffffff", "#000000"));
clip4.Position(12);
clip4.Layer(1);
clip4.Start(0);
clip4.End(3);

Clip clip5(new TextReader(1280, 720, 0, 0, GRAVITY_CENTER, "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor", "Ubuntu", 24, "#ffffff", "#000000"));
clip5.Position(12);
clip5.Layer(1);
clip5.Start(0);
clip5.End(3);

[...]

PR #346, opened 2019-10-15
Stop 24 hours worth for frames being loaded in to memory (#327)

jeffski:
The 24 hours duration seems to be creating millions of additional frames (24 hours worth) in the FrameMapper for text readers and I presume image readers to as they are just single image frames. This is causing massive memory spikes in the text reader everytime a new TextReader is added.

Not sure why the duration is set to 24 hours, as far as I can tell everything works as expected with a duration of 1 and the memory issue is gone.

Fixes issue #327

### DIFF
--- a/src/ImageReader.cpp
+++ b/src/ImageReader.cpp
@@ -82,7 +82,7 @@ void ImageReader::Open()
 		info.height = image->size().height();
 		info.pixel_ratio.num = 1;
 		info.pixel_ratio.den = 1;
-		info.duration = 1;
+		info.duration = 60 * 60 * 1;  // 1 hour duration
 		info.fps.num = 30;
 		info.fps.den = 1;
 		info.video_timebase.num = 1;
--- a/src/ImageReader.cpp
+++ b/src/ImageReader.cpp
@@ -82,7 +82,7 @@ void ImageReader::Open()
 		info.height = image->size().height();
 		info.pixel_ratio.num = 1;
 		info.pixel_ratio.den = 1;
-		info.duration = 60 * 60 * 24; // 24 hour duration
+		info.duration = 1;
 		info.fps.num = 30;
 		info.fps.den = 1;
 		info.video_timebase.num = 1;
--- a/src/QtImageReader.cpp
+++ b/src/QtImageReader.cpp
@@ -123,7 +123,7 @@ void QtImageReader::Open()
 		info.height = image->height();
 		info.pixel_ratio.num = 1;
 		info.pixel_ratio.den = 1;
-		info.duration = 60 * 60 * 24; // 24 hour duration
+		info.duration = 1;
 		info.fps.num = 30;
 		info.fps.den = 1;
 		info.video_timebase.num = 1;
--- a/src/QtImageReader.cpp
+++ b/src/QtImageReader.cpp
@@ -123,7 +123,7 @@ void QtImageReader::Open()
 		info.height = image->height();
 		info.pixel_ratio.num = 1;
 		info.pixel_ratio.den = 1;
-		info.duration = 1;
+		info.duration = 60 * 60 * 1;  // 1 hour duration
 		info.fps.num = 30;
 		info.fps.den = 1;
 		info.video_timebase.num = 1;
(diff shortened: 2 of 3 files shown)


ferdnyc:
> The 24 hours duration seems to be creating millions of additional frames (24 hours worth) in the FrameMapper for text readers and I presume image readers to as they are just single image frames. 

Well, that's not good!

> This is causing massive memory spikes in the text reader everytime a new TextReader is added.

I would imagine. Sounds like a FrameMapper bug, as I'm pretty sure it's not supposed to do that.
> 
> Not sure why the duration is set to 24 hours, as far as I can tell everything works as expected with a duration of 1 and the memory issue is gone.

Not everything, unfortunately. I just built and tested with OpenShot, and now when you import an image, everything is fine at first when you place it on the timeline... but then if you try to resize the clip at all (change the duration), it immediately _snaps_ to `1.00` and refuses to be made any longer. :cry: 

Looks like we'll either have to find a reasonable middle ground, or track down the actual FrameMapper(?) bug that's failing to reuse the image frame.

ferdnyc:
@jeffski  Would you be able to check whether this happens with libopenshot-0.2.3 release (from OpenShot 2.4.4?) If _not_, I can bisect the code and try to figure out what commit caused this.

ferdnyc:
(Actually, maybe libopenshot-0.2.2 would be a better starting point, since 0.2.3 had the new export-at-non-project-framerate code in it, and that's a pretty good candidate for the source of this bug.)

ferdnyc:
(If you have a piece of test code that triggers the bug with minimal fuss — it doesn't even have to _detect_ the bug, I can run `top` myself, just needs to clearly exhibit the excessive memory allocation — that'd also be helpful to bisect with. I could probably write it myself, but... **#laaaazy**.)

jeffski:
Hey @ferdnyc, I will try and have a look this evening (my time) but if you want to trigger the bug there is a script on the original issue page here: https://github.com/OpenShot/libopenshot/issues/327

ferdnyc:
@jeffski Oh, right, that one. Gotcha. I'll see what I can see later tonight. (Though it may be enough later that it qualifies as tomorrow morning — not just technically, but actually.)

ferdnyc:
**UGH**. The worst part about trying to bisect this is going to be the fact that the code from even just 2 releases ago is a nightmare to compile.

To build libopenshot-0.2.2 I've had to first build the matching out-of-date libopenshot-audio, which involved switching the build to gcc 5.0 because it won't build in my current GCC 9. And now the libopenshot build fell over _again_ because it tried to build against ImageMagick 7, which it's not compatible with... \*mumble, drone...\*

ferdnyc:
ARRRRGH, and I _have_ to compile it with ImageMagick, or I can't use the TextReader! Guess I'll be downgrading to IM-6. :angry:

ferdnyc:
OK, after utterly failing to ever get libopenshot-audio-0.1.7 /  libopenshot-0.2.2 successfully built in Fedora 30 even with older GCC versions, I've switched to the Ubuntu 18.10 VM I've been using for CMake testing, and am hopefully back on track to _start_ testing this.

ferdnyc:
In light of what I wrote in https://github.com/OpenShot/libopenshot/issues/327#issuecomment-546444094 I think what we should do for this, right now, is compromise and set the duration to 1 hour instead of 24. 

That'll still cap the max duration of a still-frame clip (title or image) in OpenShot, but it'll reduce the memory consumption by a factor of 24, which is a start. And multiple copies of the same clip can always be added end-to-end, to extend the duration beyond 3600 seconds.

It's not a _fix_, and the **CORRECT** fix is to figure out why output images are being duplicated for every frame in the generative Reader's duration (when it shouldn't matter, since its output is always the same image).

ferdnyc:
I've made that change to the PR branch, and I'm going to commit that once Travis finishes with it. That'll at least give us some breathing room on this.

jonoomph:
This is fine for now. A while back, I believe we did default this to 1 hour, and people were complaining that they could not add an overlay/watermark over a 1 to 2 hour video (if I recall correctly). But as mentioned, just add another clip next to it, for those very long overlays seems reasonable for a 24 times improvement.

jeffski:
Thanks for picking this up and merging. It will certainly help as is but it would be great to work out the root cause.

PR #348, opened 2019-10-26
ZeroMQ: Use IMPORTED targets

ferdnyc:
This PR adopts `FindZeroMQ.cmake` from the `cppzmq` distribution (slightly modified to still respect `$ENV{ZMQDIR}`, and to do a scan for the include dir if the library isn't found with PackageKit), then uses its `libzmq` `IMPORTED` target instead of the result variables. The separate `cppzmq` `IMPORTED` target is also used, if present.

This is significant, in part, because letting `cppzmq` do its own scan with its `FindZeroMQ.cmake` (in addition to our own `FindZMQ.cmake`) was creating two separate sets of package import data for `libzmq` in the CMake project. Now that they've been synced up, the second attempt is properly skipped as redundant.

Tested with and without a `libzmq.pc` file present in `$PKG_CONFIG_PATH`, on CMake 3.1 and 3.14.


jonoomph:
Once conflicts are resolved, let's merge this! :pray:

ferdnyc:
Done and done! (Pending Travis' permission...)

ferdnyc:
Merging. Fingers crossed on this one, for Windows/MacOS...

ferdnyc:
Hah! Guess what, this actually broke the _Linux_ builder, because there's an ancient version of JsonCPP installed on the system — which it's now finding and trying to use. :laughing: 

I'll update the `.gitlab.yml` file with `-DUSE_SYSTEM_JSONCPP=0`, and that'll fix it.

ferdnyc:
Hmm, also broke Windows. I'll track down why, should only take me a few minutes. Looks like it's pulling in the _static_ library version of ZeroMQ, which has a different target. I didn't account for that, and may have to adjust the `CMakeLists.txt` a bit to compensate.

ferdnyc:
Stil breaking Windows. I've got one more idea for a fix, if that fails then i'll revert these changes until I can figure it out.

Issue #17, opened 2016-03-08
Make tests and unittest++ optional


sbraz:
Hi,
Would it be possible to make the dependency on unittest++ optional? I am pretty sure it is only needed by developers who run tests.

RussNelson:
It should only be required by openshot_test, which oughtn't to be included in the distribution.

sbraz:
On a related note, `DISABLE_TESTS` does not entirely disable tests. I think the whole `tests` directory should be skipped if it's enabled.

PR #349, opened 2019-10-26
DISABLE_TESTS covers entire tests/ dir

ferdnyc:
I found myself sympathetic to the argument (made in #17) that `DISABLE_TESTS` should completely prevent CMake from even entering the `tests/` dir. So, with this PR, that's what it does.

Fixes: #17


ferdnyc:
Merging as trivial (and to close what's currently our third-oldest open issue!)

PR #350, opened 2019-10-26
Bump CMake min-version to 3.2, display during build

ferdnyc:
This bumps the minimum CMake version just _slightly_ to `3.2` (from `3.1`), as 3.2 was when the propagation of target properties to linked targets was fully realized in the source. While our current `CMakeLists.txt` files _seem_ to work with CMake 3.1, at least on Linux, I'm concerned there may be subtle breakages down the line. CMake 3.2 should be far safer, and is still well within the realm of what we can reasonably expect to have on even the oldest still-supported systems.

Output of the `${CMAKE_VERSION}` variable is also added to the config stage:
<pre><code>
$ cmake32 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_VERBOSE_MAKEFILE=1 -DDISABLE_TESTS=1 ..
-----------------------------------------------------------------
          Welcome to the OpenShot Build System!

CMake will now check libopenshot's build dependencies and inform
you of any missing files or other issues.

For more information, please visit <http://www.openshot.org/>.
-----------------------------------------------------------------

Generating build files for OpenShot <b>with CMake 3.2.3</b>
  Building libopenshot (version 0.2.3)
  SO/API/ABI Version: 17

</code></pre>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@
 # along with OpenShot Library. If not, see <http://www.gnu.org/licenses/>.
 ################################################################################
 
-cmake_minimum_required(VERSION 3.1...3.14 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.2...3.14 FATAL_ERROR)
 
 message("\
 -----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging as a minor build quirk (and because I want to check the CMake version on our various builders).

ferdnyc:
So, here's the situation:

| Build host | CMake version |
|--------------|---------------------|
|Travis FFmpeg2 (Ubuntu 16.04 Xenial) | 3.12.4 |
|Travis FFmpeg3/4 (Ubuntu 18.04 Bionic) | 3.12.4 |
|OpenShot Linux builder | 3.7.2 |
|OpenShot Windows builder | 3.15.0 :exclamation:  |
|OpenShot MacOS builder | 3.2.3 :disappointed: |

So, I was right that 3.2 is OK, but we can't go any higher while the MacOS builder remains in its current config.

ferdnyc:
@jonoomph 
The Mac builder may _have_ to have its CMake upgraded, though, because it's currently failing to link with ImageMagick despite `7.0.8-56` being installed. I've seen that in my own testing on Linux as well, and it's down to the `FindImageMagick.cmake` from CMake 3.2 simply being too old to understand IM7.

PR #351, opened 2019-10-26
Threads: Use IMPORTED target

ferdnyc:
We were already doing this since #344 merged, but this explicitly imports `Threads::Threads` _before_ looking for OpenMP, and uses the same (CMake-recommended) options that we're now using in `libopenshot-audio`.


jonoomph:
@ferdnyc Once conflicts are resolved, let's merge this one as well! :pray:

PR #352, opened 2019-10-27
Add fixes for Clang/IWYU building, correct juce:: namespacing

ferdnyc:
So, I spent part of the weekend running libopenshot through [include-what-you-use (IWYU)](https://github.com/include-what-you-use/include-what-you-use/), a `#include` scanner that attempts to help you... well, it [does exactly what it says on the tin](https://en.wikipedia.org/wiki/Does_exactly_what_it_says_on_the_tin). I may have some changes to submit along those lines down the road, but this PR isn't actually about that.

#### include-what-you-use

IWYU is based on, and runs best with, Clang, so in order to run the code through IWYU I first had to get it buildable under Clang. This PR is the work necessary to make that happen.

It was actually fairly easy; the main library source _initially_ didn't require any changes at all.

#### The `juce::` namespace and header pollution

What _did_ ultimately force me to make changes, and where things snagged up a bit at first, was when I got to the unit tests. The unit tests are all based on UnitTest++, and as such they `#include` the `UnitTest++.h` header, which supplies a set of test macros. Those macros all call various methods in the `UnitTest::Test` namespace.

However, JUCE _also_ has a unit test framework, which lives at `juce::UnitTest`. Because the `OpenShot.h` header indirectly includes the JUCE headers from libopenshot-audio, and because those headers (by default) apply a `using namespace juce;` when they're included, _technically_ the name `UnitTest` is ambiguous — it could mean either the `UnitTest` namespace from UnitTest++, _or_ the `juce::UnitTest` class from libopenshot-audio. GCC was able to ignore this ambiguity and figure things out, but it turns out Clang is pickier. The unit tests **would not** compile with Clang as long as that `using namespace juce;` was being added by the headers.

#### Clearing the <strike>air</strike> global namespace

The obvious solution was to add a `#define DONT_SET_USING_JUCE_NAMESPACE 1` to the unit tests before they did their `#include "../include/OpenShot.h"`, and that's exactly what I did.

But doing that means our _headers_ have to be COMPATIBLE when included with _or_ without the `juce` symbols imported into the global namespace, meaning they  have to fully-qualify all of their JUCE symbols as `juce::WhateverClass`. This PR is the work necessary to do that, as well as the modifications to the unit test files to disable the automatic `using namespace juce;`.

This is exactly the kind of issue PR #296 is all about, and a perfect example of why `using namespace foo;` is _such_ bad practice in header files. It's dangerous, it causes problems, and it makes it much, much harder to use the headers in question, if they're going to pollute the global namespace when included, _and_ (worse!) if they're written in a way that **REQUIRES** the global namespace to be polluted with foreign symbols, in order for them to build properly.

#### ENABLE_IWYU

I also tossed in changes to `src/CMakeLists.txt` to facilitate use of IWYU:
* A new `option(ENABLE_IWYU)` is defined, default `OFF` but you can switch it on with `cmake -DENABLE_IWYU=1` at configure time. That'll turn on IWYU as long as CMake finds an executable `iwyu` program in the PATH, and all of the source files will be run through it.
* Another variable, `IWYU_OPTS`, can be set to pass options to `iwyu`. So, for example, you might run:
   ```sh
   cmake -DENABLE_IWYU=1 -DIWYU_OPTS="--max_line_length=120 --verbose=1"
   ```
   to configure the build.

Note that, as I said, IWYU is best paired with Clang. However, the CMake environment doesn't _enforce_ that, and you're welcome to try and run it with any compiler you like. It might even work. But for optimal and best-supported results, you really want to set `CC=clang CXX=clang++` in the environment before running `cmake`. (Or add `-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++` to the `cmake` command line.)

### DIFF
--- a/include/AudioBufferSource.h
+++ b/include/AudioBufferSource.h
@@ -54,25 +54,25 @@ namespace openshot
 	 * The <a href="http://www.juce.com/">JUCE</a> library cannot play audio directly from an AudioSampleBuffer, so this class exposes
 	 * an AudioSampleBuffer as a AudioSource, so that JUCE can play the audio.
 	 */
-	class AudioBufferSource : public PositionableAudioSource
+	class AudioBufferSource : public juce::PositionableAudioSource
 	{
 	private:
 		int position;
 		int start;
 		bool repeat;
-		AudioSampleBuffer *buffer;
+		juce::AudioSampleBuffer *buffer;
 
 	public:
 		/// @brief Default constructor
 		/// @param audio_buffer This buffer contains the samples you want to play through JUCE.
-		AudioBufferSource(AudioSampleBuffer *audio_buffer);
+		AudioBufferSource(juce::AudioSampleBuffer *audio_buffer);
 
 		/// Destructor
 		~AudioBufferSource();
 
 		/// @brief Get the next block of audio samples
 		/// @param info This struct informs us of which samples are needed next.
-		void getNextAudioBlock (const AudioSourceChannelInfo& info);
+		void getNextAudioBlock (const juce::AudioSourceChannelInfo& info);
 
 		/// Prepare to play this audio source
 		void prepareToPlay(int, double);
@@ -82,13 +82,13 @@ namespace openshot
 
 		/// @brief Set the next read position of this source
 		/// @param newPosition The sample # to start reading from
-		void setNextReadPosition (int64 newPosition);
+		void setNextReadPosition (juce::int64 newPosition);
 
 		/// Get the next read position of this source
-		int64 getNextReadPosition() const;
+		juce::int64 getNextReadPosition() const;
 
 		/// Get the total length (in samples) of this audio source
-		int64 getTotalLength() const;
+		juce::int64 getTotalLength() const;
 
 		/// Determines if this audio source should repeat when it reaches the end
 		bool isLooping() const;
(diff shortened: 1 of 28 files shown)


jonoomph:
Very interesting! And although this might create a bunch more conflicts, let's try and merge this.

ferdnyc:
> And although this might create a bunch more conflicts,

Yeah, it did in #296 at least (actually, they were only a small chunk in each of three files), but they were easily-enough resolved and #296 is back to conflict-free status.

ferdnyc:
(Aside: I really wish GitHub **showed** which PRs have merge conflicts from the PR list view, the same way it shows CI status with either a ✔️ or ❌ icon. Or, failing that, if it at least provided a way to filter on conflict status. AFAICT, currently there's no way to know which PRs have conflicts except by visiting each one and scrolling all the way to the bottom of the page. **#LAME**)

ferdnyc:
LOL! I just found this (it's from **2012**) in the Frame header:

https://github.com/OpenShot/libopenshot/blob/9d391fab3acb2cc953ac8699d30fb8c942c438f6/include/Frame.h#L34-L37

So, I guess this _has_ been an issue in the past, even with GCC. (Should be safe to remove now, though.)

cc: @jonoomph

ferdnyc:
(It's also in all of these files, which makes it all the more impressive that I didn't notice it when working on this:)
* `include/AudioBufferSource.h`
* `include/AudioReaderSource.h`
* `include/AudioResampler.h`
* `include/ClipBase.h`
* `include/Clip.h`

PR #354, opened 2019-10-30
CMake: Fix IWYU invocation with unset IWYU_OPTS

ferdnyc:
I messed up the `ENABLE_IWYU` handling in #352, so that it wouldn't be run if you _didn't_ set `IWYU_OPTS`. This fixes that.

It also adds a `CMAKE_VERSION` check to disable IWYU checking in CMake 3.2, since support for it wasn't introduced until CMake 3.3. Under CMake 3.2 it won't be run even if it's installed and present. (A note to that effect is placed in the `OPTION` doc string.)


ferdnyc:
Merging as it's just a small tweak to an existing (and wholly optional) build feature.

PR #355, opened 2019-11-01
Use bundled JsonCPP on linux builder

ferdnyc:
Fix for the official package builder, in the wake of merging #348.


ferdnyc:
Yep, that worked. (Correction, it was #341 that caused the Linux breakage.)

PR #356, opened 2019-11-01
Fix target creation/importing for ZeroMQ on Windows

ferdnyc:
Windows finds only the _static_ version of the library (`libzmq.dll.a`, but not `libzmq.dll`), which the `FindZeroMQ.cmake` from cppzmq places in a different target (`libzmq-static`).

It was also creating _both_ targets, with the `libzmq` `IMPORTED LOCATION` set to `libzmq-NOTFOUND`.

That's not particularly helpful, so it's been updated to only export `FOUND` targets, and `src/CMakeLists.txt` will link only the existing one — preferring `libzmq` (`SHARED`), falling back to `libzmq-static` if necessary.


ferdnyc:
I'm only 90% sure `PRIVATE` is the right scope for the `libzmq-static` target, it may need to be `PUBLIC` to pass along the include dirs. But let's try it and see...

PR #357, opened 2019-11-01
Fix Windows ZeroMQ for real

ferdnyc:
Not 100% sure _why_, honestly, but this version of the Find module works where the previous one didn't.

I believe it comes down to this note in the CMake docs for `add_library`:
<blockquote>An UNKNOWN library type is typically only used in the implementation of Find Modules. It allows the path to an imported library (often found using the find_library() command) to be used without having to know what type of library it is. <b>This is especially useful on Windows where a static library and a DLL’s import library both have the same file extension.</b></blockquote>

This also does away with the `ZMQ_DIR` envvar entirely — `ZeroMQ_DIR` can be set as a CMake variable (`cmake -DZeroMQ_DIR:PATH="C:\msys64\mingw64\"`, for example) to provide the location of ZeroMQ.

But that actually shouldn't be necessary, if it was installed using `pacman` on Windows it should be autodetected just like any other platform. I've removed any setting of the ZeroMQ path from the `.gitlab-ci.yml` Windows configs, as a test of that theory.

It _miiiiiiight_ break Win32, in which case we just add a `-DZeroMQ_DIR:PATH="C:\msys32\usr\"` there (only). But I hope it'll Just Work™, because `pkg-config` is in play on MSYS.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -70,13 +70,12 @@ windows-builder-x64:
     - Expand-Archive -Path artifacts.zip -DestinationPath .
     - $env:LIBOPENSHOT_AUDIO_DIR = "$CI_PROJECT_DIR\build\install-x64"
     - $env:UNITTEST_DIR = "C:\msys64\usr"
-    - $env:ZMQDIR = "C:\msys64\usr"
     - $env:RESVGDIR = "C:\msys64\usr"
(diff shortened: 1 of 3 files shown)


ferdnyc:
Hah! That commit history is an example of the crazy that happens when you're trying to edit the same files on Linux and in a Windows VM at the same time. I'll squash it when committing.

ferdnyc:
Both Windows builders auto-detected the correct path, I think we're good. I'll watch the builds until they finish, though.

ferdnyc:
:+1: Complete success! :100:  Finally. :stuck_out_tongue_closed_eyes: 

Now I just have to report their brokenness to the `cppzmq` folks.

ferdnyc:
> Now I just have to report their brokenness to the `cppzmq` folks.

Done, as zeromq/cppzmq#361.

PR #358, opened 2019-11-02
Clarify some comments

SuslikV:
I was need to look at animation keys for some new feature and found that few comments are wrong.

### DIFF
--- a/include/KeyFrame.h
+++ b/include/KeyFrame.h
@@ -98,7 +98,7 @@ namespace openshot {
 		/// Default constructor for the Keyframe class
 		Keyframe();
 
-		/// Constructor which sets the default point & coordinate at X=0
+		/// Constructor which sets the default point & coordinate at X=1
 		Keyframe(double value);
 
 		/// Add a new point on the key-frame.  Each point has a primary coordinate, a left handle, and a right handle.
--- a/src/KeyFrame.cpp
+++ b/src/KeyFrame.cpp
@@ -55,7 +55,7 @@ void Keyframe::ReorderPoints() {
 	}
 }
 
-// Constructor which sets the default point & coordinate at X=0
+// Constructor which sets the default point & coordinate at X=1
 Keyframe::Keyframe(double value) : needs_update(true) {
 	// Init the factorial table, needed by bezier curves
 	CreateFactorialTable();
--- a/src/Point.cpp
+++ b/src/Point.cpp
@@ -33,7 +33,7 @@
 using namespace std;
 using namespace openshot;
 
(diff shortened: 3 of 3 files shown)


ferdnyc:
Yeah, there are a bunch of comments that weren't updated to match the code around them, when it changed.

...For exactly that reason, personally I prefer that comments _not_ include details like that, which then have to be kept in sync with the code. When it's obvious enough just from looking at the code what the values are, better to leave them out of the comments. Otherwise, they're just going to be wrong again the next time the code changes.

src/Point.cpp
```diff
@@ -43,7 +43,7 @@ Point::Point() : interpolation(BEZIER), handle_type(AUTO)
 	Initialize_Handles();
 }
 
-// Constructor which creates a single coordinate at X=0
+// Constructor which creates a single coordinate at X=1
```
ferdnyc: Like, for this one, you only need to look down 3 lines to see that it creates a `Coordinate(1, y)`. There's no reason for the value to be in the comment at all, IMHO.

src/Point.cpp
```diff
@@ -33,7 +33,7 @@
 using namespace std;
 using namespace openshot;
 
-// Default constructor (defaults to 0,0)
+// Default constructor (defaults to 1,0)
```
ferdnyc: Same thing here. Look down 3 lines, and the fact that it defaults to `(1,0)` is completely apparent from the `Coordinate(1, 0)` in the code itself.

include/KeyFrame.h
```diff
@@ -98,7 +98,7 @@ namespace openshot {
 		/// Default constructor for the Keyframe class
 		Keyframe();
 
-		/// Constructor which sets the default point & coordinate at X=0
+		/// Constructor which sets the default point & coordinate at X=1
```
ferdnyc: This one is probably worth keeping, since it's a documentation comment.

src/KeyFrame.cpp
```diff
@@ -55,7 +55,7 @@ void Keyframe::ReorderPoints() {
 	}
 }
 
-// Constructor which sets the default point & coordinate at X=0
+// Constructor which sets the default point & coordinate at X=1
```
ferdnyc: This one... I'm _really_ ambivalent about. It's _not_ as immediately obvious from reading the code, so I suppose it's arguably useful. (It's certainly better that it at least be _**correct**_, though, so thanks for taking the time to clean these up!)

ferdnyc:
No update, so merging as-is. The correction if definitely an improvement. Thanks again!

PR #360, opened 2019-11-04
CMake: Use targets for QT, AUTOMOC sources, eliminate source globbing

ferdnyc:
This is kind of a major one. Not only does it move to using the Qt targets for linking, but it:
* switches over to AUTOMOC instead of manually wrapping files (some of which never needed to be wrapped),
* removes the globbing for `src/effects/` and `src/Qt/` files. (All source files now have to be named explicitly in `src/CMakeLists.txt`, so if any are added the appropriate list needs to be updated.)
* Calls `include_directories()` once from the top level `CMakeLists.txt`, and then not again.
* Eliminates the `REQUIRED_LIBRARIES` variable from CMake, continuing the process begun in previous `IMPORTED`-target PRs where dependencies get linked separately using multiple `target_link_libraries()` calls.
* <strike>Switches to `$<$<BOOL:Foo_FOUND>:${Foo_LIBRARIES}>` generator expressions, for conditional linking.</strike> (OK, this last one didn't work out.)

The PR branch lives in the project repo, to give the builders a chance to chew on the changes. Marked WIP only pending the outcome of that testing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,13 +72,6 @@ configure_file(include/OpenShotVersion.h.in include/OpenShotVersion.h @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/include/OpenShotVersion.h
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/libopenshot)
 
-############### Set up include paths #################
-list(APPEND OPENSHOT_INCLUDE_DIRS
-    ${CMAKE_CURRENT_SOURCE_DIR}/include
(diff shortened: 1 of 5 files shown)


ferdnyc:
Broke on Windows with ZMQ issues, I think I may have created this branch during yesterday's FindZeroMQ fixes. I just merged `develop`, hoping that solves it.

ferdnyc:
Yup, that fixed it. Removing WIP flag.

jonoomph:
LGTM!

PR #361, opened 2019-11-04
Add Clang compilation to Travis build matrix

ferdnyc:
Now that libopenshot is buildable with Clang, this adds it as another job in the Travis CI build matrix.

(It tests FFmpeg 3 simply because that was the shortest, simplest build configuration to duplicate. Be nice if our builds used an actual matrix-based configuration and clang could just be added as another dimension, but alas we're not there yet.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+compiler: gcc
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging this, as it only affects the CI environment.

PR #366, opened 2019-11-17
Change high quality scaling to use SWS_BICUBIC

jonoomph:
Fixing a small implementation detail related to using high quality scaling in FFMpegWriter, and switching from SWS_LANCZOS to SWS_BICUBIC, for better quality when upscaling images (with a small performance loss). But sharper exports are aim, and if it costs us 1 or 2% slower speed, it seems worth it to me.

Real-time previews still use a faster algorithm: SWS_FAST_BILINEAR.

Some comparisons of the FFmpeg scaling methods: 
http://www.programmersought.com/article/4493483389/

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -1339,7 +1339,7 @@ void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
 
 		int scale_mode = SWS_FAST_BILINEAR;
 		if (openshot::Settings::Instance()->HIGH_QUALITY_SCALING) {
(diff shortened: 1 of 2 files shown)


SuslikV:
The mentioned algorithms of FFmpeg has nothing to do with the OpenShot's method of handling the transform. Did you forgot the Qt Transform?

How to see the quality difference you are talking about (on writing)? The `sws_getContext` doing planes conversion only. If your source is small and placed on a larger canvas - the Qt transform should apply first (bilinear) in OpenShot.

PR #367, opened 2019-11-19
Enable coverage reporting for openshot-test

musteresel:
Uses CodeCoverage.cmake from [1] to build libopenshot with gcov
support.  Option ENABLE_COVERAGE enables this and sets a target
test_coverage which builds and runs openshot-tests and collects
coverage information into a html report.

I'm currently using this to make sure everything that I'm going to modify is covered by at least some test.  If this is useful it might even be interesting to add `make test_coverage` to a travis build to gather the resulting report as artifact.

Coverage reports should be made with `CMAKE_BUILD_TYPE` set to Debug, otherwise the report might be a bit strange.  The libopenshot generate with `ENABLE_COVERAGE=ON` should of course *NOT* be used by external projects ...

[1]: https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,9 @@ include(FeatureSummary)
 # Optional build settings for libopenshot
 option(USE_SYSTEM_JSONCPP "Use system installed JsonCpp, if found" ON)
 option(DISABLE_BUNDLED_JSONCPP "Don't fall back to bundled JsonCpp" OFF)
+option(DISABLE_TESTS "Don't build unit tests" OFF)
 option(ENABLE_IWYU "Enable 'Include What You Use' scanner (CMake 3.3+)" OFF)
+option(ENABLE_COVERAGE "Enable coverage reporting" OFF)
 
 ########## Configure Version.h header ##############
 configure_file(include/OpenShotVersion.h.in include/OpenShotVersion.h @ONLY)
@@ -95,13 +97,20 @@ include_directories(
   ${CMAKE_CURRENT_BINARY_DIR}/include)
 
 ############## Code Coverage #########################
-option(ENABLE_COVERAGE "Enable coverage reporting." OFF)
+if (DISABLE_TESTS AND ENABLE_COVERAGE)
+  message(WARNING "ENABLE_COVERAGE requires tests, overriding DISABLE_TESTS")
+  set(DISABLE_TESTS OFF CACHE BOOL "Don't build unit tests" FORCE)
+endif()
 
 if (ENABLE_COVERAGE)
+  if (NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE "Debug")
+    message(STATUS "Coverage enabled, setting build type to Debug")
+  endif()
   include(CodeCoverage)
   APPEND_COVERAGE_COMPILER_FLAGS()
 endif()
-
+add_feature_info("Coverage" ENABLE_COVERAGE "analyze test coverage and generate report")
 
 ############## PROCESS src/ DIRECTORIES ##############
 add_subdirectory(src)
@@ -128,6 +137,15 @@ endif()
 ############# PROCESS tests/ DIRECTORY ##############
 if(NOT DISABLE_TESTS)
   add_subdirectory(tests)
+endif()
+
+############## COVERAGE REPORTING #################
+if (ENABLE_COVERAGE)
+  setup_target_for_coverage_lcov(
+    NAME coverage
+    EXECUTABLE openshot-test
+    DEPENDENCIES openshot-test)
+    message("Generate coverage report with 'make coverage'")
 endif()
 
 ########### PRINT FEATURE SUMMARY ##############
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,6 +143,7 @@ endif()
 if (ENABLE_COVERAGE)
   setup_target_for_coverage_lcov(
     NAME coverage
+    LCOV_ARGS "--no-external"
     EXECUTABLE openshot-test
     DEPENDENCIES openshot-test)
     message("Generate coverage report with 'make coverage'")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,15 @@ include_directories(
   ${CMAKE_CURRENT_SOURCE_DIR}/include
   ${CMAKE_CURRENT_BINARY_DIR}/include)
 
+############## Code Coverage #########################
+option(ENABLE_COVERAGE "Enable coverage reporting." OFF)
+
+if (ENABLE_COVERAGE)
+  include(CodeCoverage)
+  APPEND_COVERAGE_COMPILER_FLAGS()
+endif()
+
+
 ############## PROCESS src/ DIRECTORIES ##############
 add_subdirectory(src)
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -151,11 +151,3 @@ endif()
 
 string(CONCAT t ${OS_TEST_CMDS})
 message("\nTo run unit tests, use: ${t}")
-
-if (ENABLE_COVERAGE)
-  setup_target_for_coverage_lcov(
-    NAME test_coverage
-    EXECUTABLE openshot-test
-    DEPENDENCIES openshot-test)
-endif()
-
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -151,3 +151,11 @@ endif()
 
 string(CONCAT t ${OS_TEST_CMDS})
 message("\nTo run unit tests, use: ${t}")
+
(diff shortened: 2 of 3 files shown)


ferdnyc:
I think this is a great idea, and I absolutely think we should do it. I'll add a few specific questions/suggestions in a review after, and I also have some suggested changes to the `CodeCoverage.cmake` code itself in commit 70bf3df41d3291e6a56afae8311e8c6a18a96378 on my `coverage-build` branch. But first, some general thoughts:

#### External files

The report is generated with a ton of system include files listed, which makes it kind of busy. But I'm having a bear of a time trying to get them off of there, since `COVERAGE_LCOV_EXCLUDES` takes globs, and apparently _not_ recursive ones. So, `set(COVERAGE_LCOV_EXCLUDES "/usr/include")` or `..."/usr/include/*")` has basically no effect, and while `set(COVERAGE_LCOV_EXCLUDES "/usr/include/qt5/*/*")` works fine for all of the Qt5 includes, but it seems like every include directory would have to be listed individually, and even just `"/usr/include/c++"` would require at least four levels of globs to catch everything. 

I also tried configuring `lcov` with `--no-external`, but that failed miserably. I guess due to the way the CMake build is structured, _all_ of the source files register as external.

So I guess I'm wondering:
1. Do you think we _should_ exclude system include files (assuming we can), or is there value to having them listed in the report?
1. Any thoughts on how?

If need be I can probably work up some CMake generator expressions based on the openshot target's `TARGET_INCLUDE_DIRECTORIES`, but I want to make sure I'm not missing something obvious with `lcov` and/or `genhtml` first. (And also that it's not a bad idea in the first place.)

#### Cleanup

My `CodeCoverage.cmake` changes would add the `${Coverage_NAME}.info` file and the `${Coverage_NAME}` report directory itself to the list of items cleaned up by '`make clean`', which makes me feel a little better, but there doesn't seem to be any good way to get CMake to clean up all of the `.gcda` and `.gcno` files, so they just sit around in the build directory for as long as they feel like.

The coverage tool itself doesn't even have a _list_ of files that it could feed  into an `ADDITIONAL_MAKE_CLEAN_FILES` property. Ideally the files would be set as `GENERATED` during the compile step, but because they're created as side-effects of the `gcc` run, even CMake doesn't know they're there, really. I mean, we could _glob_ for them, but that's ugly and file-globbing in CMake configurations is heavily discouraged for very good reasons.

...Not really a question, just acknowledging the issue, but as I said there really doesn't seem to be a good way around it.

> If this is useful it might even be interesting to add `make test_coverage` to a travis build to gather the resulting report as artifact.

Hmm, that's one option, or what do you think about using a CI coverage reporting system? I know [Coveralls](coveralls.io) is popular, though I don't know much about it (or any of them), but there seem to be a bunch popping up since GitHub opened up the Actions system.

ferdnyc:
> I also tried configuring `lcov` with `--no-external`, but that failed miserably. I guess due to the way the CMake build is structured, _all_ of the source files register as external.
> 
> So I guess I'm wondering:
> 
> 1. Do you think we _should_ exclude system include files (assuming we can), or is there value to having them listed in the report?
> 2. Any thoughts on how?

Ooh! Well, if we do decide to go that route, I just solved it. Adding `-b ${PROJECT_SOURCE_DIR}` to all of the `lcov` commands in `CodeCoverage.cmake` roots the report correctly, and allows `--no-external` to work as expected.

ferdnyc:
Commit ab3aef4 is that change; it works fine with or without `--no-external` passed in `LCOV_ARGS`.

tests/CMakeLists.txt
```diff
@@ -151,3 +151,11 @@ endif()
 
 string(CONCAT t ${OS_TEST_CMDS})
 message("\nTo run unit tests, use: ${t}")
+
+if (ENABLE_COVERAGE)
+  setup_target_for_coverage_lcov(
+    NAME test_coverage
```
ferdnyc:
```suggestion
    NAME coverage
```

What do you think about just calling it 'coverage', so you can build the report with '`make coverage`'? It matches the `ENABLE_COVERAGE` option name, and it's slightly more concise.

CMakeLists.txt
```diff
@@ -85,6 +85,15 @@ include_directories(
...
   ${CMAKE_CURRENT_BINARY_DIR}/include)
 
+############## Code Coverage #########################
+option(ENABLE_COVERAGE "Enable coverage reporting." OFF)
+
+if (ENABLE_COVERAGE)
+  include(CodeCoverage)
+  APPEND_COVERAGE_COMPILER_FLAGS()
+endif()
+
```
ferdnyc: So, I've tested, and I think I would actually move this to the root `CMakeLists.txt`, at the bottom after `add_subdirectory(tests)` is processed. The coverage report is generated at the top level anyway, not in the `tests/` subdirectory, and that way it keeps CMake from polluting `tests/` with target-tracking for the coverage target.

My commit b4682ac makes that change, and also adds some additional tooling to, among other things...
1. Turn `DISABLE_TESTS` into a full-fledged CMake `option()`
1. Check if the user has set `DISABLE_TESTS` and `ENABLE_COVERAGE` together, and emit a warning before overriding and force-enabling the tests
1. Check if `CMAKE_BUILD_TYPE` is **unset**, and if so automatically set it to `"Debug"`. (If it's something other than debug, the `WARNING` in `CodeCoverage.cmake` will fire, so no need to handle any of those cases.)

CMakeLists.txt
```diff
@@ -85,6 +85,15 @@ include_directories(
...
+
```
ferdnyc:
```suggestion
add_feature_info("Coverage" ENABLE_COVERAGE "analyze test coverage and generate report")
```

Inserting this line (along with a merge of #383) will add a Coverage entry to the FeatureSummary output at the end of the configuration run.

CMakeLists.txt
ferdnyc: Yeah, the previous comment was attached to the wrong block. Oops.

tests/CMakeLists.txt
```diff
@@ -151,3 +151,11 @@ endif()
...
+    EXECUTABLE openshot-test
+    DEPENDENCIES openshot-test)
+endif()
+
```
ferdnyc: So, I've tested, and I think I would actually move this to the root `CMakeLists.txt`, at the bottom after `add_subdirectory(tests)` is processed. The coverage report is generated at the top level anyway, not in the `tests/` subdirectory, and that way it keeps CMake from polluting `tests/` with target-tracking for the coverage target.

My commit b4682ac makes that change, and also adds some additional tooling to, among other things...
1. Turn `DISABLE_TESTS` into a full-fledged CMake `option()`
1. Check if the user has set `DISABLE_TESTS` and `ENABLE_COVERAGE` together, and emit a warning before overriding and force-enabling the tests
1. Check if `CMAKE_BUILD_TYPE` is **unset**, and if so automatically set it to `"Debug"`. (If it's something other than debug, the `WARNING` in `CodeCoverage.cmake` will fire, so no need to handle any of those cases.)

ferdnyc:
In my [`coverage-build`](https://github.com/ferdnyc/libopenshot/tree/coverage-build) branch I've imported the latest changes to `CodeCoverage.cmake` from upstream, and applied the latest modifications from [my own upstream PR](https://github.com/bilke/cmake-modules/pull/39).

CMakeLists.txt
```diff
@@ -85,6 +85,15 @@ include_directories(
   ${CMAKE_CURRENT_SOURCE_DIR}/include
   ${CMAKE_CURRENT_BINARY_DIR}/include)
 
+############## Code Coverage #########################
+option(ENABLE_COVERAGE "Enable coverage reporting." OFF)
+
+if (ENABLE_COVERAGE)
+  include(CodeCoverage)
+  APPEND_COVERAGE_COMPILER_FLAGS()
```
ferdnyc:
```suggestion
  append_coverage_compiler_flags()
```

My PR upstream will lowercase these in the module file, as well — current CMake norms are that `ALLCAPS_IS_FOR_VARIABLES` while `functions_use_lowercase()`.

ferdnyc:
Well, no updates on this in a while, so I decided to [Be Bold](https://en.wikipedia.org/wiki/Wikipedia:Be_bold)™ (...wait, what do you mean "Wrong project?")  and pull the trigger on this, because I still think it's a good idea.

I've merged the branch with my suggested tweaks into @musteresel 's, and once Travis finishes chewing on it I'm going to merge this. It doesn't affect the build _anyway_ unless you specifically ask it to, so it's really low-impact.

After it's set up, I'm going to look into getting coverage reports uploaded to Codacy as well, so it can nag us about the lack of tests. #397 has me wanting to see more of the code covered by unit tests. As much as I hate writing unit tests.

ferdnyc:
Passed. Merging, with my sincere thanks @musteresel !

PR #371, opened 2019-11-21
Use non-subsampled color format for lossless encoding with the h264 encoder

SuslikV:
By default, the 4:2:0 is used, that is not lossless itself. Thus, even if crf=0 was set - the loss of the color info still occurs. This change is step to fix it.

This is part of the: https://github.com/OpenShot/libopenshot/pull/219


SuslikV:
Updated on top of _develop_ branch. Additional code for new "0 cqp" setting.

jonoomph:
This seems harmless enough. Famous last words.

Issue #370, opened 2019-11-20
Public interface for Keyframe


musteresel:
The current public interface of `Keyframe` includes the public member `Points`.  **I'm about to remove this member.  This <s>will</s>** might **break a lot of code** (<s>basically every place where Python code - e.g. in OpenShot-qt - uses `keyframe["Points"]` ... which is currently 88 places in OpenShot-qt</s> *nothing in OpenShot-qt as far as I can see now*).

There's `AddPoint()`, `RemovePoint()`, `GetPoint()` among others.  These should be used instead of direct access to the `Points` member to give `Keyframe` a chance to keep it's internal state in order (e.g. sort order of points, interpolated values, ..).

All in all the public interface of `Keyframe` is a bit chaotic. It supports access to the points similar like a container (add, remove, find, ..) but of course also access to interpolated values (it's main job, after all).

The question is: How should the public interface of `Keyframe` look on the C++ side, and what should it look like on the Python side?

*edit: The question shifts a bit ... it's still important to get a clean public interface. But for openshot-qt the only really important functions are `Keyframe::SetJsonValue()`, `Keyframe::AddPoint`, the "get functions" and `Keyframe::Json()`.*

musteresel:
OpenShot-qt uses `Points` mostly in these ways:

 - iterating over it and indexing into it, giving read **and write** access to each point
 - `append`ing new points
 - `remove`ing points by point value (not index)

This is a huge problem. The python code currently does stuff like this ([src/windows/views/timeline_webview.py#L334](https://github.com/OpenShot/openshot-qt/blob/develop/src/windows/views/timeline_webview.py#L334))

~~~python
brightness["Points"][-1]["co"]["X"] = round(duration * fps_float) + 1
~~~

Removing `Points` and providing only `AddPoint` and `RemovePoint` makes this much more complicated to write for the python side.

One possible approach is the following:

`Keyframe` (C++ side) can provide access to a proxy class which acts like a container of points. E.g. you can iterate over points, remove and add points, and access individual points by index. Instead of giving access to the points however, the proxy container gives access to "proxy points".  These points notify the `Keyframe` instance the point is part of over any changes that are made to them.

Example:

~~~python
p = some_keyframe.points()[20]
p = some_other_point
~~~

This would call `ProxyPoint::operator=` on the C++ side, which in turn would call `RemovePoint()` for the point which is proxied, then `AddPoint()` for the given new point and set itself up as proxy for the new point (which might have another index!)

[...]

PR #374, opened 2019-11-25
New Keyframe implementation

musteresel:
This is a new implementation of the Keyframe class.

**Differences**
The main difference is that the old implementation was "fully cached" - by which I mean that it computed all interpolated values for all X coordinates on first request, and had to re-compute all values when anything changed.

The new implementation is completely "on demand", it does not cache any previously computed value(s) but instead computes just the one requested value for each `GetValue()`, `GetInt()` or `GetLong()` call.  This means modifications to the points do not invalidate any cache (as there is none) and are thus fast.

Regarding points: The previously public data member `Points` is now private and needs to be modified via the provided `AddPoint` and `RemovePoint` functions.  Some code outside of the Keyframe class had to be modified to not access `Points.size()` but rather use the provided `GetCount()`. Same for the now gone `Values` vector: `Values.size()` has been replaced with `GetLength()`.

Many of the other member functions have been modified as well; some just to remove dead / unused code, others to use `GetValue()` instead of the now gone `Values` vector.

I've added new tests to capture the (in part strange, but that's for #370 ) old behavior of the modified member functions and to make sure the new functions follow the same behavior.

Some test values have been modified. This is because the new code produces much smoother Bezier curvers than the old one:

0-to-255-in-12-with-gnuplot

**Performance**

I tested two "scenarios":
 1. Setting up all points of the Keyframe and then querying all values.
 2. Setting up all points of the Keyframe, querying all values while also doing random changes to the Keyframe points.

Because I removed the caching I was expecting to see a performance *loss* in the first case. This is not the case however. The following shows a point for the time each `GetValue` call of a 100000 frames (X values) long Keyframe with random constant, linear, and Bezier curve segments:

no-change

One clearly sees that the old code only starts to produce Values after some time due to the initial calculation of all values. The new code however starts producing values right from the start, and does so also considerably faster. The vertical lines indicate when "constructing" the keyframe (that is adding all points) has been finished, showing that also `AddPoint` has become significantly faster.  This is due to the old `AddPoint` code using selection sort to re-sort the points. The new code just inserts the new point at the correct position.

For the other scenario - the "interactive editing scenario" - I see even more drastic performance improvements:

out-with-change

The old code fails here because on each change it has to re-compute all values of the (long) Keyframe.

The test data has been generated with this program:

~~~c++
#include <OpenShot.h>


#include <algorithm>
#include <chrono>
#include <cstdlib>


using namespace openshot;
using std::chrono::steady_clock;


int main() {
	unsigned sections = 400;
	unsigned const section_length = 250;
	Keyframe f;
	InterpolationType types[] = {CONSTANT, BEZIER, LINEAR};
	double position = 0;
	auto const start = steady_clock::now();
	f.AddPoint(position, rand());
	while (sections-- > 0) {
		position += section_length;
		f.AddPoint(position, rand(), types[0]);
		std::rotate(std::begin(types), std::begin(types) + 1,
					std::end(types));
	}
	std::chrono::duration<double> const init =
		steady_clock::now() - start;
	std::cerr << "init after " << init.count() << std::endl;
	for (int i = 0; i < position; ++i) {
		#ifdef WITH_CHANGE
		if (rand() > (0.99 * RAND_MAX)) {
			f.AddPoint(i + 1, 0);
			std::cerr << " - added point @ " << i + 1 << std::endl;
		}
		#endif
		double const v = f.GetValue(i);
		std::chrono::duration<double> d = steady_clock::now() - start;
		std::cout << i << " = " << v <<  " after " << d.count() << "\n";
	}
}
~~~

**What now?**
Please test the code, also from OpenShot. Can you reproduce the performance difference? Let me hear suggestions where to further improve the code.

Things I want to improve further:

 - `SetJsonValue()` - which as I know now is used a lot from Openshot - currently calls `AddPoint` for each parsed point.  While asymptotically O(n * log n) I think this can be optimized by first storing the points, and then sorting them *if necessary* (as they will come sorted most of the time from Openshot eitherway).

 - Some functions (`RemovePoint`) still use looping instead of binary search on `Points`

 - Code quality: `int64_t` is not the best candidate to use as an index variable. `std::size_t` is the logical choice.

Things originally planned but probably not necessary:

 - Caching some results (e.g. parts of a Bezier curve).  This will add complexity, can make performance *worse* if the cache does not correctly predict the usage pattern and given the good performance is probably not worth the effort.

### DIFF
--- a/tests/KeyFrame_Tests.cpp
+++ b/tests/KeyFrame_Tests.cpp
@@ -51,7 +51,7 @@ TEST(Keyframe_GetPoint_With_1_Points)
 	k1.AddPoint(openshot::Point(2,3));
 
 	CHECK_THROW(k1.GetPoint(-1), OutOfBoundsPoint);
-	CHECK_EQUAL(1, k1.Points.size());
+	CHECK_EQUAL(1, k1.GetCount());
 	CHECK_CLOSE(2.0f, k1.GetPoint(0).co.X, 0.00001);
 	CHECK_CLOSE(3.0f, k1.GetPoint(0).co.Y, 0.00001);
 	CHECK_THROW(k1.GetPoint(1), OutOfBoundsPoint);
@@ -98,7 +98,7 @@ TEST(Keyframe_GetValue_For_Bezier_Curve_2_Points)
 	CHECK_CLOSE(3.81602f, kf.GetValue(40), 0.0001);
 	CHECK_CLOSE(4.0f, kf.GetValue(50), 0.0001);
 	// Check the expected number of values
-	CHECK_EQUAL(kf.Values.size(), 51);
+	CHECK_EQUAL(kf.GetLength(), 51);
 }
 
 TEST(Keyframe_GetValue_For_Bezier_Curve_5_Points_40_Percent_Handle)
@@ -121,7 +121,7 @@ TEST(Keyframe_GetValue_For_Bezier_Curve_5_Points_40_Percent_Handle)
 	CHECK_CLOSE(1.77569f, kf.GetValue(177), 0.0001);
 	CHECK_CLOSE(3.0f, kf.GetValue(200), 0.0001);
 	// Check the expected number of values
-	CHECK_EQUAL(kf.Values.size(), 201);
+	CHECK_EQUAL(kf.GetLength(), 201);
 }
 
 TEST(Keyframe_GetValue_For_Bezier_Curve_5_Points_25_Percent_Handle)
@@ -144,7 +144,7 @@ TEST(Keyframe_GetValue_For_Bezier_Curve_5_Points_25_Percent_Handle)
 	CHECK_CLOSE(1.77569f, kf.GetValue(177), 0.0001);
 	CHECK_CLOSE(3.0f, kf.GetValue(200), 0.0001);
 	// Check the expected number of values
-	CHECK_EQUAL(kf.Values.size(), 201);
+	CHECK_EQUAL(kf.GetLength(), 201);
 }
 
 TEST(Keyframe_GetValue_For_Linear_Curve_3_Points)
@@ -164,7 +164,7 @@ TEST(Keyframe_GetValue_For_Linear_Curve_3_Points)
 	CHECK_CLOSE(4.4f, kf.GetValue(40), 0.0001);
 	CHECK_CLOSE(2.0f, kf.GetValue(50), 0.0001);
 	// Check the expected number of values
-	CHECK_EQUAL(kf.Values.size(), 51);
+	CHECK_EQUAL(kf.GetLength(), 51);
 }
 
 TEST(Keyframe_GetValue_For_Constant_Curve_3_Points)
@@ -185,7 +185,7 @@ TEST(Keyframe_GetValue_For_Constant_Curve_3_Points)
 	CHECK_CLOSE(8.0f, kf.GetValue(49), 0.0001);
 	CHECK_CLOSE(2.0f, kf.GetValue(50), 0.0001);
 	// Check the expected number of values
-	CHECK_EQUAL(kf.Values.size(), 51);
+	CHECK_EQUAL(kf.GetLength(), 51);
 }
 
 TEST(Keyframe_Check_Direction_and_Repeat_Fractions)
--- a/tests/KeyFrame_Tests.cpp
+++ b/tests/KeyFrame_Tests.cpp
@@ -92,10 +92,10 @@ TEST(Keyframe_GetValue_For_Bezier_Curve_2_Points)
 	// Spot check values from the curve
 	CHECK_CLOSE(1.0f, kf.GetValue(-1), 0.0001);
 	CHECK_CLOSE(1.0f, kf.GetValue(0), 0.0001);
-	CHECK_CLOSE(1.00023f, kf.GetValue(1), 0.0001);
-	CHECK_CLOSE(1.14025f, kf.GetValue(9), 0.0001);
-	CHECK_CLOSE(1.91492f, kf.GetValue(20), 0.0001);
-	CHECK_CLOSE(3.81602f, kf.GetValue(40), 0.0001);
+	CHECK_CLOSE(1.0f, kf.GetValue(1), 0.0001);
+	CHECK_CLOSE(1.12414f, kf.GetValue(9), 0.0001);
+	CHECK_CLOSE(1.86370f, kf.GetValue(20), 0.0001);
+	CHECK_CLOSE(3.79733f, kf.GetValue(40), 0.0001);
 	CHECK_CLOSE(4.0f, kf.GetValue(50), 0.0001);
 	// Check the expected number of values
 	CHECK_EQUAL(kf.GetLength(), 51);
@@ -114,11 +114,11 @@ TEST(Keyframe_GetValue_For_Bezier_Curve_5_Points_40_Percent_Handle)
 	// Spot check values from the curve
 	CHECK_CLOSE(kf.GetValue(-1), 1.0f, 0.0001);
 	CHECK_CLOSE(1.0f, kf.GetValue(0), 0.0001);
-	CHECK_CLOSE(1.00023f, kf.GetValue(1), 0.0001);
-	CHECK_CLOSE(2.73656f, kf.GetValue(27), 0.0001);
-	CHECK_CLOSE(7.55139f, kf.GetValue(77), 0.0001);
-	CHECK_CLOSE(4.08102f, kf.GetValue(127), 0.0001);
-	CHECK_CLOSE(1.77569f, kf.GetValue(177), 0.0001);
+	CHECK_CLOSE(1.0f, kf.GetValue(1), 0.0001);
+	CHECK_CLOSE(2.68197f, kf.GetValue(27), 0.0001);
+	CHECK_CLOSE(7.47719f, kf.GetValue(77), 0.0001);
+	CHECK_CLOSE(4.20468f, kf.GetValue(127), 0.0001);
+	CHECK_CLOSE(1.73860f, kf.GetValue(177), 0.0001);
 	CHECK_CLOSE(3.0f, kf.GetValue(200), 0.0001);
 	// Check the expected number of values
 	CHECK_EQUAL(kf.GetLength(), 201);
@@ -137,11 +137,11 @@ TEST(Keyframe_GetValue_For_Bezier_Curve_5_Points_25_Percent_Handle)
 	// Spot check values from the curve
 	CHECK_CLOSE(1.0f, kf.GetValue(-1), 0.0001);
 	CHECK_CLOSE(1.0f, kf.GetValue(0), 0.0001);
-	CHECK_CLOSE(1.00023f, kf.GetValue(1), 0.0001);
-	CHECK_CLOSE(2.73656f, kf.GetValue(27), 0.0001);
-	CHECK_CLOSE(7.55139f, kf.GetValue(77), 0.0001);
-	CHECK_CLOSE(4.08102f, kf.GetValue(127), 0.0001);
-	CHECK_CLOSE(1.77569f, kf.GetValue(177), 0.0001);
+	CHECK_CLOSE(1.0f, kf.GetValue(1), 0.0001);
+	CHECK_CLOSE(2.68197f, kf.GetValue(27), 0.0001);
+	CHECK_CLOSE(7.47719f, kf.GetValue(77), 0.0001);
+	CHECK_CLOSE(4.20468f, kf.GetValue(127), 0.0001);
+	CHECK_CLOSE(1.73860f, kf.GetValue(177), 0.0001);
 	CHECK_CLOSE(3.0f, kf.GetValue(200), 0.0001);
 	// Check the expected number of values
 	CHECK_EQUAL(kf.GetLength(), 201);
@@ -200,7 +200,7 @@ TEST(Keyframe_Check_Direction_and_Repeat_Fractions)
 	CHECK_EQUAL(kf.GetInt(1), 500);
 	CHECK_EQUAL(kf.IsIncreasing(1), false);
 	CHECK_EQUAL(kf.GetRepeatFraction(1).num, 1);
-	CHECK_EQUAL(kf.GetRepeatFraction(1).den, 12);
+	CHECK_EQUAL(kf.GetRepeatFraction(1).den, 13);
 	CHECK_EQUAL(kf.GetDelta(1), 500);
 
 	CHECK_EQUAL(kf.GetInt(24), 498);
@@ -212,13 +212,13 @@ TEST(Keyframe_Check_Direction_and_Repeat_Fractions)
 	CHECK_EQUAL(kf.GetLong(390), 100);
 	CHECK_EQUAL(kf.IsIncreasing(390), true);
 	CHECK_EQUAL(kf.GetRepeatFraction(390).num, 3);
-	CHECK_EQUAL(kf.GetRepeatFraction(390).den, 15);
+	CHECK_EQUAL(kf.GetRepeatFraction(390).den, 16);
 	CHECK_EQUAL(kf.GetDelta(390), 0);
 
 	CHECK_EQUAL(kf.GetLong(391), 100);
 	CHECK_EQUAL(kf.IsIncreasing(391), true);
 	CHECK_EQUAL(kf.GetRepeatFraction(391).num, 4);
-	CHECK_EQUAL(kf.GetRepeatFraction(391).den, 15);
+	CHECK_EQUAL(kf.GetRepeatFraction(391).den, 16);
 	CHECK_EQUAL(kf.GetDelta(388), -1);
 }
 
@@ -307,8 +307,8 @@ TEST(Keyframe_Scale_Keyframe)
 	CHECK_CLOSE(1.0f, kf.GetValue(1), 0.01);
 	CHECK_CLOSE(7.99f, kf.GetValue(24), 0.01);
 	CHECK_CLOSE(8.0f, kf.GetValue(25), 0.01);
-	CHECK_CLOSE(3.68f, kf.GetValue(40), 0.01);
-	CHECK_CLOSE(2.0f, kf.GetValue(49), 0.01);
+	CHECK_CLOSE(3.85f, kf.GetValue(40), 0.01);
+	CHECK_CLOSE(2.01f, kf.GetValue(49), 0.01);
 	CHECK_CLOSE(2.0f, kf.GetValue(50), 0.01);
 
 	// Resize / Scale the keyframe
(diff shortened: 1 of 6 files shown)


jonoomph:
Thanks @musteresel, this looks great so far! Let me grab these changes and run some local tests today. I'll post my results soon.

musteresel:
`IsIncreasing()` works now with points instead of interpolated values.  This is much faster especially with keyframes with slow changes.

tests/KeyFrame_Tests.cpp
```diff
@@ -380,22 +380,105 @@ TEST(Keyframe_Remove_Duplicate_Point)
...
+	// - point where all next values are equal => false
+	// - point where first non-eq next value is smaller => false
+	// - point where first non-eq next value is larger => true
+	Keyframe kf;
+	kf.AddPoint(1, 1, LINEAR); // testing with linear
+	kf.AddPoint(3, 5, BEZIER); // testing with bezier
+	kf.AddPoint(6, 10, CONSTANT); // first non-eq is smaller
+	kf.AddPoint(8, 8, CONSTANT); // first non-eq is larger
+	kf.AddPoint(10, 10, CONSTANT); // all next values are equal
+	kf.AddPoint(15, 10, CONSTANT);
```
ferdnyc: I have to think more about the answer to the comment-question here, but as a general observation, I feel like some of these tests (the new ones, but also the old ones as well — in fact, _especially_ the old ones) could be doing more to test _modifications_ to the data.

IOW, using this test as an example, right now it:
1. Creates a Keyframe
2. Does a bunch of `AddPoint()`s in increasing order along the `X` axis
3. Checks the result of several `IsIncreasing()` calls at different `X` points, to make sure it returns the correct result each time.

But it would probably make for more rigorous testing if we were to _then_ do something like:
```c++
kf.AddPoint(4, 2, BEZIER);
kf.AddPoint(7, 10, CONSTANT);
kf.AddPoint(12, 15, LINEAR);
kf.AddPoint(14, 8, CONSTANT);
```
Followed by more checks at interesting points, to ensure that the `IsIncreasing()` result has changed where it should have changed, **and** that it _hasn't_ changed wherever it should still be the same.

(The original data is packed kind of tight, obviously if it were spread out more along the `X` axis then the subsequent insertions could be made at more natural intervals.)

I don't know if data can be persisted between individual tests in the same source file using UnitTest++. (I don't see why not, if it's moved to file scope instead of being created at function scope. But maybe there are ordering or parallelism concerns?) If it's possible, then it might make sense to build the modification test a separate test instance which makes changes to the `Keyframe` object from the previous instance.

ferdnyc: Speaking of modification tests, I don't see a single test to `openshot::Keyframe::UpdatePoint()` in the entire file, which feels like a huge testing hole. Modifications to existing data (and the unexpected side effects they may have) are a prime opportunity for bugs in the implementation.

I bring all this up because there's an open bug report on OpenShot's keyframe behavior (OpenShot/openshot-qt#2552): It seems that, when the interpolation mode for a keyframe point is changed, the interpolation of the _previous_ point is **ALSO** changed to the same value. Which I suspect is a bug in OpenShot, not libopenshot, and may even be intentional given the somewhat-odd way that interpolation in any segment is computed based on the mode of the _next_ point after the segment.

(It's still a bug, though, because it causes nonsensical side effects.)

For example, say you have keyframe points set up like this...
```c++
kf.AddPoint(1, 1, CONSTANT); // The initial default keyframe, pretend it's the Alpha property
kf.AddPoint(10, 0.5, LINEAR);  // Added by modifying the Alpha value in the GUI
kf.AddPoint(20, 0.8, LINEAR);  // Same
```
The value will be decreasing from `X = 1` to `X = 10`, then increasing from `X = 10` to `X = 20`.

If you now change `(20, 0.8)` to `CONSTANT`, and it _also_ changes `(10, 0.5)` to `CONSTANT`, then the value will be a constant `0.5` from `X = 10` to `X = 19`, but it will **ALSO** be a constant `1.0` from `X = 1` to `X = 9`, which is almost certainly not what was expected to happen.

So, even though as I said, I suspect that particular bug is in OpenShot rather than libopenshot, it would still be a good idea to test that the methods which make Keyframe modifications — meaning `AddPoint`, `UpdatePoint`, and any other method that can alter the contents of the list — all have the expected effects on _existing_ data.

ferdnyc: > given the somewhat-odd way that interpolation in any segment is computed based on the mode of the _next_ point after the segment.

I'm really curious _why_ interpolation works that way, BTW. I'd love it if @jonoomph or someone could fill in the blanks there.

It actually seems more than "somewhat" odd. It's **extremely** odd, to be honest, and not at all what I'd have initially expected — if anything, I'd expect that setting interpolation at a given point would change how the values are computed _after_ that point, not _until_ that point.

Was it just done so that, when a `LINEAR` keyframe is added to a point on a clip that already contains a `CONSTANT` initial point, the value will be interpolated between the initial value and the new point?

If so, wouldn't it have made more sense to just create the initial keyframe point as `LINEAR`, so that any added `LINEAR` points would immediately trigger interpolation _without_ the weird "type-of-next-point-controls-segment-interpolation" rule? (Especially since the value already holds constant at any point _after_ the last keyframe, so a single initial `LINEAR` keyframe would act exactly the same as a single initial `CONSTANT` keyframe.)

These are the questions that keep me up at night. (And yes, I realize how pathetic that sounds.)

musteresel: > I don't know if data can be persisted between individual tests in the same source file using UnitTest++. (I don't see why not, if it's moved to file scope instead of being created at function scope. But maybe there are ordering or parallelism concerns?) If it's possible, then it might make sense to build the modification test a separate test instance which makes changes to the Keyframe object from the previous instance.

We could make a `Keyframe GetStandardTestKeyframe()` function which returns a suitably complex "standard test keyframe".  That avoids any issues with ordering / parallelism but still reduces code. We need to decide on what that "standard test keyframe" looks like then, though. And of course modifications to *that* keyframe will require modifications in **all** tests using it.

Moreover there will be cases not covered by the "standard test keyframe"; these still need to be handled then.  A good balance between a simple "standard test keyframe" and one which catches many (edge) cases will be needed.

> I feel like some of these tests (the new ones, but also the old ones as well — in fact, especially the old ones) could be doing more to test modifications to the data.

Yes, especially the old ones *really* would have needed tests of that. And if we do "black box" testing - testing without assuming anything about the implementation - then you're right, we need tests that modifications don't mess things up.  The *new implementation* however does not care that much about modifications ... it doesn't cache things or something like that.  All it needs is a properly sorted `Points` vector.  A (rather) small test of this there: https://github.com/OpenShot/libopenshot/blob/ed0b0818033cab39da39b37c7bfe5237dd37954a/tests/KeyFrame_Tests.cpp#L403

> It seems that, when the interpolation mode for a keyframe point is changed, the interpolation of the previous point is ALSO changed to the same value. 

**Definitively** not a libopenshot Keyframe issue :) We only get one big pack of JSON from openshot, after all. So we don't get "changes" but rather: "Hey, throw a way all data and re-initialize with this JSON".

musteresel: > It actually seems more than "somewhat" odd. It's extremely odd, to be honest, and not at all what I'd have initially expected — if anything, I'd expect that setting interpolation at a given point would change how the values are computed after that point, not until that point.

Hm. Well, you need to decide on one or the other. Either it's the first point of a "segment" which controls the interpolation *or* the last.  With proper implementation (in Openshot) the user doesn't need to know about any of that, though.

You call `GetClosestPoint(frame_number_where_the_cursor_is)` which **by default** searches to the right and modify *that* points interpolation:

~~~c++
Keyframe kf;
kf.AddPoint(1,10);
kf.AddPoint(10,0);
// Ugh ... I want LINEAR interpolation between these two points!
auto p = kf.GetClosestPoint(Point(5,0)); // we've talked about the public API ... well ...
p.interpolation = LINEAR;
kf.AddPoint(p);
// TADA :)
// Which points interpolation changed doesn't need to care us, it's an implementation detail.
~~~

That said, the *clean* API for that would probably be:

 - `AddPoint(Coordinate)`: Adds a fix point to the Keyframe, no mention of any interpolation yet.
 - `SetInterpolation(int64_t frame, InterpolationType const & t)`: Makes it so that the interpolation used to obtain the value at `frame` is done via `t`.  `InterpolationType` could then be a base class, from which constant, linear etc derive. We could also make a "constant_from_left" and "constant_from_right" interpolation type.

ferdnyc: > We could make a `Keyframe GetStandardTestKeyframe()` function which returns a suitably complex "standard test keyframe". That avoids any issues with ordering / parallelism but still reduces code. We need to decide on what that "standard test keyframe" looks like then, though. And of course modifications to _that_ keyframe will require modifications in **all** tests using it.

The bigger issue with that, I think, is that if the "standard test keyframe" is constructed using the Keyframe code being tested (which it would have to be), it creates a chicken-and-egg problem. If all tests (or a significant number) are dependent on the implementation of the standard keyframe, it seems like it diminishes the value of the individual tests. And the more complex the standard test keyframe gets, the greater that possibility.

Having to update tests when changes are made that affect the standard keyframe isn't even really my concern. My bigger concern is, say a _bug_ is introduced in the Keyframe "setup" code (meaning whatever code is used to create the standard keyframe), but it's a relatively subtle bug that only impacts Keyframes if they happen to have some particular arrangement of points. Two possible scenarios:
1. If the standard keyframe happens to _not_ have that arrangement, then none of the tests catch the bug.
2. If the standard data _does_ trigger the bug, then **all** of the (dependent) tests suddenly start failing all at once, which still doesn't reveal very much about the nature of the bug.

The potential for discovering the true nature of the bug, quite possibly via differential diagnosis based on which tests are impacted and which aren't, would seem to be much higher if every test uses a _different_ set of data. (Except, possibly, the occasional test that picks up from where a previous test left off. — Though TBH I may have just talked myself out of that idea _completely_, for the same reasons.)

ferdnyc: > Hm. Well, you need to decide on one or the other. Either it's the first point of a "segment" which controls the interpolation _or_ the last.

Or both, really. Meaning, _theoretically_ every (interior) point has TWO interpolation modes, one governing the input curve and one governing the output curve. For `BEZIER` that's made explicit with the `handle_left` and `handle_right` curve adjustments, but it's not impossible to define an interpolation function with `LINEAR` attachment on one end  and `BEZIER` on the other (switching over at roughly the halfway point, perhaps?) — really, `CONSTANT` is really the only one that _requires_ a segment have the same interpolation at both e—**!!!**

Aaaand, I just realized that's exactly _why_ the OpenShot bug exists: Because **if** you had separate input and output interpolation parameters for each keyframe point, then a change to the interpolation mode at one end of a line segment _**WOULD**_ require applying the same interpolation to (the connected side of) the point at the other end as well, in order for it to apply along the entire length of the segment.

But our segments are governed _only_ by the setting at the _end_ point. So changing the interpolation setting at both ends mistakenly affects not only the current segment, but the previous _segment_ as well.

jonoomph: @ferdnyc I think you have summarized the issue quite well. I don't remember the exact reasoning, other than the current implementation was the best one I landed on (with regards to which point in a segment controls the interpolation). For whatever reason, the end point made sense to my brain. 

I also agree that the bug is almost certainly contained in openshot-qt, and not libopenshot. I've very excited to merge this Keyframe PR into develop though, as it solves one of the biggest problems with libopenshot and long clips (i.e. long Keyframes more like it). :+1:

jonoomph: Keyframes can now be any length and any complexity, and the performance of GetValue should always be extremely fast (i.e. faster than we need).

musteresel: > Or both, really. Meaning, theoretically every (interior) point has TWO interpolation modes [..]

@ferdnyc I hope to not offend you, but no - you're loosing focus. Think about this: A video, you add a simple zoom effect. Slowly linear in, then a bezier one out. How would you describe this? You start at zoom 1, then have a linear zoom to (say) 1.5, then a zoom back to 1 with a bezier curve. So

 - Fix point at (start_X, 1)
 - Fix point at (zoom_full, 1.5)
 - Fix point at (zoom_end, 1)
 - Linear between start_x and zoom_full
 - Bezier between zoom_full and zoom_end

Try to make this when interpolation takes both end points (= two possibly different interpolation modes!) into account. What interpolation should be set on the first point? The second? The third?

> But our segments are governed only by the setting at the end point. 

Which is an implementation detail. We could also keep a vector of coordinates, and a separate vector of "interpolations". Would IMO be better!

> So changing the interpolation setting at both ends mistakenly affects not only the current segment, but the previous segment as well.

True, which is why you shouldn't do this. If you want to change how a particular frame gets its value, then you call `GetClosestPoint(that_frame)`, change that points interpolation, and "update" via `AddPoint()`.

**There's one thing, though:** `handle_right` is used from the left point, which of course breaks my reasoning :D

musteresel: > The bigger issue with that, I think, is that if the "standard test keyframe" is constructed using the Keyframe code being tested (which it would have to be), it creates a chicken-and-egg problem. 

Hm, being honest: I cannot follow you here ... how's that an issue? I was think about something along the lines of:

~~~c++
Keyframe GetStandardTestKeyframe() {
  Keyframe kf;
  kf.AddPoint(1,0);
  kf.AddPoint(10, 10);
  // ... or from a JSON, but this adds the JSON parsing stuff in...
  return kf;
}
~~~

musteresel:
`GetRepeatFraction()` is now also no longer using a linear scan; but rather binary search to find the range where the value remains constant (when rounding).

This was a hard one ... it passes the tests, but I'm not 100% sure that all edge cases are covered. I'm reasonably sure that the new code will not break/crash, but there might be some edge case where it returns a different value!

jonoomph:
LGTM!

jonoomph:
@musteresel This all looks terrific, and I think it's good to merge at this point. I'm gonna go ahead and merge, and continue testing and verifying the behavior.

tests/KeyFrame_Tests.cpp
ferdnyc: > @ferdnyc I hope to not offend you, but no - you're loosing focus.

Never had it, but that's beside the point.

> **There's one thing, though:** `handle_right` is used from the left point, which of course breaks my reasoning :D

Exactly, because it's meant as the handle for the curve attachment on the keyframe point's (vertex's) right side, each vertex having curve attachments on both sides. If you envision it as a literal handle, a physical lever that allows you to deform the curve at the point it attaches to the vertex (which is how it's represented in Inkscape, the Gimp's path editor, and most other Bezier implementations), then it makes sense that its position be indicated by which side of the vertex it's on. The focus isn't on the segment, it's on the vertices.

e.g. after I drew this rectangle in Inkscape, then converted it to paths and dragged out a couple of attachment handles:
image

You could describe that a lot of ways, but because the handles are attached _to_ the points (not to the line), it makes sense to me to describe it as a line that runs from the right side of the shape's upper left vertex, with its eccentricity described by the vertex's right-hand handle, and terminates at the upper right vertex, which has a left-hand handle.

Linear interpolation _in a sense_ is just a special-case Bezier curve — you can think of it two ways, either it's a Bezier curve with no handle on either end (in which case the curve will just be a straight line), or a curve where both handles point directly at the other vertex. And in fact, if you set up a bezier curve with both handles set that way, it will be exactly linear:
image

Now, that could be two points with really short handles pulled out just a bit from their position, or two points with really _long_ handles that reach almost all the way to the other end of the line — it's the latter, in this case, but it's nearly impossible to tell because the former would produce an identical line. Versus a curve where only one side "has linear attachment" (in a manner of speaking):
image

But a major factor that gave rise to that bug (and a factor that makes it not _strictly_ a bug, I'm now realizing) is this: A _segment's_ interpolation mode CAN'T be described solely by the point at one end or the other (in the current implementation), because the interpolation mode assigned at a point determines _whether or not that **point** has Bezier handles — on **both** sides_.

If a point is set `BEZIER` then it will have _both_ a `handle_left` and a `handle_right`, whereas a point set `LINEAR` or `CONSTANT` will have no handles at all. Which means that, when you set the interpolation mode at the end of a line segment to `BEZIER`, you **DO** have to change the mode at the other end, in order to give the start point a `handle_right` property.

So the problem is twofold:
1. That's being done for all changes, even when setting a point to `CONSTANT` or `LINEAR`
2. It breaks whenever the start point is a `LINEAR` or `CONSTANT` endpoint, because changing it it will affect the previous segment.

So we either need a way to assign interpolation parameters **directly** to line segments, instead of to any points (at either end), or we need a way to add `handle_right` values to `LINEAR ` and `CONSTANT` points when they're followed by a `BEZIER` point, without changing the interpolation mode set on the point itself.

(Or, we just treat all **points** as `BEZIER` vertices, give them all `handle_left` and `handle_right` properties, and find a way to special-case `LINEAR` and `CONSTANT` interpolation modes.) 

And regardless what we do about handles, I would still argue that at least `CONSTANT` interpolation should _begin_ at the point where it's set, because that's much more in line with users' expectations. If they intend for a property to have a constant value of `0.7`, they don't think that should involve going to the _end_ and setting a keyframe that says "use the value `0.7` up until this point". They want to set a keyframe at a certain point that says, "now hold the value constant at `0.7`, until I say otherwise".

ferdnyc: > Hm, being honest: I cannot follow you here ... how's that an issue?

OK, forget that part, let's keep it simple:
> The potential for discovering [any sort of subtle bug] would seem to be much higher if every test uses a different set of data. (Except, possibly, the occasional test that picks up from where a previous test left off. — Though TBH I may have just talked myself out of that idea completely, for the same reasons.)

Building a new ad hoc data set for each test, rather than reusing the same one each time, means we exercise more of the code in different scenarios.

ferdnyc: > And regardless what we do about handles, I would still argue that at least `CONSTANT` interpolation should _begin_ at the point where it's set, because that's much more in line with users' expectations. If they intend for a property to have a constant value of `0.7`, they don't think that should involve going to the _end_ and setting a keyframe that says "use the value `0.7` up until this point". They want to set a keyframe at a certain point that says, "now hold the value constant at `0.7`, until I say otherwise".

Er, sorry! That's not even how it works. It's worse than that.

In the _current_ logic, setting a point to `CONSTANT` means, "hold the _value_ of the **previous point** constant, up until this point". Which is just... **#GUH????**

musteresel: @ferdnyc  Ah, now I understand what you mean. I was concerned that you wanted to make interpolation depend on *two* interpolation modes (that of the point left, that of the point right). Basically an "interpolation" of the interpolation...

Well, what about my suggestion here: https://github.com/OpenShot/libopenshot/pull/374#discussion_r354165165

Or developed a bit further:

 - `AddFixPoint(Coordinate)` to add a fix point to the keyframe curve
 - `SetInterpolation(SegmentNum, InterpolationMode)` to change how the points between two fix points are interpolated
 - `GetSegmentNum(FrameNum)` to ease with the conversion between frame numbers (e.g. clicking on the timeline between the points -> frame number -> segment number)
 - `RemovePoint(CoordinateIndex, useLeft)` to remove a fix point. Need to specify how the interpolation of the two adjacent segments should be merged.

SuslikV: @ferdnyc Small remark: the handle can't leave the rectangle, because it is in percent from the coordinate. So, the posted pictures is not accurate in sense of libopenshot's Bezier control points implementation.

Issue #202, opened 2019-03-15
A few optimizations 


schauveau:
Generally speaking, libopenshot is well written and quite efficient but I noticed a few things that could be optimized:
- ImageReader is creating a Magick::Image that is converted into a new QImage for each frame (via Frame::AddMagickImage). That conversion is not very efficient and could be avoided by reusing the same QImage.    
-  `QImage::bits()` and `QImage::scanline()` provide a non-const pointer to the image data. If the QImage is shared (because of a shallow copy), this causes an automatic duplication of the data. Consequently, if the data is not going to be modified then it could be beneficial to use `QImage::constBits()` or `QImage::constScanline()` instead. Potential gains are in `Frame::GetPixels()`,  `Frame::AddImage(std::shared_ptr<QImage> new_image, bool only_odd_lines)`, `Frame::GetPixels(int row)`. and maybe a few other places.
-  In several places, the pixels of a QImage are processed with a loop of the form `for (n = 0; n < image->columns() * image->rows(); n += 1)`. This is highly inefficient because `QImage::rows()`  and `QImage::columns()` are functions that cannot be inlined by the compiler. Consequently, the compiler cannot assume that their values are constant and must perform the 2 calls at each iteration. Also, the code cannot be vectorized. This is found in `Timeline::add_layer` and in most of the effect classes. A similar problem can be found in `Frame::AddMagickImage` in the loop that converts all pixels of the Magick::Image to RGBA.

ferdnyc:
> * `QImage::bits()` and `QImage::scanline()` provide a non-const pointer to the image data. If the QImage is shared (because of a shallow copy), this causes an automatic duplication of the data. Consequently, if the data is not going to be modified then it could be beneficial to use `QImage::constBits()` or `QImage::constScanline()` instead. Potential gains are in `Frame::GetPixels()`,  `Frame::AddImage(std::shared_ptr<QImage> new_image, bool only_odd_lines)`, `Frame::GetPixels(int row)`. and maybe a few other places.

On this point: As I noted in #375, `QImage::bits()` and `QImage::scanLine()` are overloaded functions, and in `const` context they're equivalent to `constBits()` and `constScanline()` automatically. We merged the change to use the `const...()` forms explicitly, for the sake of code readability and clarity, but it was purely code-cosmetic — the `const` context of the previous calls meant that they were already avoiding any unnecessary copying.

PR #375, opened 2019-11-26
Frame.cpp: Avoid unnecessary copy of image data

musteresel:
As mentioned in issue #202 `QImage::bits()` and `QImage::scanLine()` make a deep copy of the source image.  This is completely unnecessary when read-only access to the pixel data is required.  Changing to `QImage::constBits()` and `QImage::constScanLine()` solves this.  Both functions were introduced in Qt 4.7.

Tested using some test project in OpenShot, not sure if I covered all cases, though!

https://doc.qt.io/qt-5/qimage.html#constBits

### DIFF
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -692,7 +692,7 @@ void Frame::Thumbnail(std::string path, int new_width, int new_height, std::stri
 
 		// Get pixels
 		unsigned char *pixels = (unsigned char *) thumbnail->bits();
-		unsigned char const *mask_pixels = (unsigned char *) mask->constBits();
+		const unsigned char *mask_pixels = (const unsigned char *) mask->constBits();
 
 		// Convert the mask image to grayscale
 		// Loop through pixels
@@ -922,7 +922,7 @@ std::shared_ptr<Magick::Image> Frame::GetMagickImage()
 		AddColor(width, height, "#000000");
 
 	// Get the pixels from the frame image
-	QRgb const *tmpBits = (const QRgb*)image->constBits();
+	const QRgb *tmpBits = (const QRgb*)image->constBits();
 
 	// Create new image object, and fill with pixel data
 	std::shared_ptr<Magick::Image> magick_image = std::shared_ptr<Magick::Image>(new Magick::Image(image->width(), image->height(),"RGBA", Magick::CharPixel, tmpBits));
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -280,7 +280,7 @@ const unsigned char* Frame::GetWaveformPixels(int width, int height, int Red, in
 	wave_image = GetWaveform(width, height, Red, Green, Blue, Alpha);
 
 	// Return array of pixel packets
-	return wave_image->bits();
+	return wave_image->constBits();
 }
 
 // Display the wave form
@@ -473,14 +473,14 @@ const unsigned char* Frame::GetPixels()
 		AddColor(width, height, color);
 
 	// Return array of pixel packets
-	return image->bits();
+	return image->constBits();
 }
 
 // Get pixel data (for only a single scan-line)
 const unsigned char* Frame::GetPixels(int row)
 {
 	// Return array of pixel packets
-	return image->scanLine(row);
+	return image->constScanLine(row);
 }
 
 // Check a specific pixel color value (returns True/False)
@@ -692,7 +692,7 @@ void Frame::Thumbnail(std::string path, int new_width, int new_height, std::stri
 
 		// Get pixels
 		unsigned char *pixels = (unsigned char *) thumbnail->bits();
-		unsigned char *mask_pixels = (unsigned char *) mask->bits();
+		unsigned char const *mask_pixels = (unsigned char *) mask->constBits();
 
 		// Convert the mask image to grayscale
 		// Loop through pixels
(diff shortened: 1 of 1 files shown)


ferdnyc:
@musteresel I looked at this a bit after #202 was filed, and at the time I wasn't completely sold on the actual issue here. (For reasons I'll come to shortly.)

Given that you'd filed this PR, though, I checked it out and ran some tests to compare the effects.

### `valgrind` memcheck tests

I first ran `openshot-test` from both the current `develop` branch code and this PR branch through `valgrind`'s memcheck scanner. `valgrind` did show slightly fewer logged 'errors' with the PR code. (Any flagged event is an error, even "potentially lost" memory blocks... and in truth the difference may be down to my not having some debug symbols installed on the first run. If I were being 100% rigorous I'd rebuild and rerun the first test in the exact same environment.)

The end-result memory stats were _exactly_ the same, though, which I wouldn't have expected to be the case if lots of unnecessary copies were happening.

The result summary from a baseline `openshot-test` run without these changes:
```
$ cd build
$ valgrind --trace-children=yes --leak-check=full tests/openshot-test
[...]
==179087== LEAK SUMMARY:
==179087==    definitely lost: 73,246 bytes in 135 blocks
==179087==    indirectly lost: 615,383 bytes in 1,023 blocks
==179087==      possibly lost: 42,021,360 bytes in 66 blocks
==179087==    still reachable: 601,332 bytes in 2,898 blocks
==179087==                       of which reachable via heuristic:
==179087==                         newarray           : 1,536 bytes in 16 blocks
==179087==         suppressed: 0 bytes in 0 blocks
==179087== Reachable blocks (those to which a pointer was found) are not shown.
==179087== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==179087==
==179087== Use --track-origins=yes to see where uninitialised values come from
==179087== For lists of detected and suppressed errors, rerun with: -s
==179087== ERROR SUMMARY: 130 errors from 113 contexts (suppressed: 0 from 0)
```

For `openshot-test` with the PR applied:
```
$ cd build_pr375
$ valgrind --trace-children=yes --leak-check=full tests/openshot-test            
==180335== LEAK SUMMARY:
==180335==    definitely lost: 73,246 bytes in 135 blocks
==180335==    indirectly lost: 615,383 bytes in 1,023 blocks
==180335==      possibly lost: 42,021,360 bytes in 66 blocks
==180335==    still reachable: 601,332 bytes in 2,898 blocks
==180335==                       of which reachable via heuristic:
==180335==                         newarray           : 1,536 bytes in 16 blocks
==180335==         suppressed: 0 bytes in 0 blocks
==180335== Reachable blocks (those to which a pointer was found) are not shown.
==180335== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==180335==
==180335== Use --track-origins=yes to see where uninitialised values come from
==180335== For lists of detected and suppressed errors, rerun with: -s
==180335== ERROR SUMMARY: 118 errors from 101 contexts (suppressed: 0 from 0)
```

The full output from both runs:
* [pr375_baseline_openshot-test_20191127.txt](https://github.com/OpenShot/libopenshot/files/3898419/pr375_baseline_openshot-test_20191127.txt)
* [pr375_enabled_openshot-test_20191127.txt](https://github.com/OpenShot/libopenshot/files/3898420/pr375_enabled_openshot-test_20191127.txt)

Running `valgrind` did point me at an uninitialized-variable issue in FrameMapper, though, so the effort was doubly useful. I'll file a PR on that shortly.

### OpenShot export `memusage` tests

Now, it's possible that the change would reduce only memory _consumption_ during the program execution, but as long as all of that memory was subsequently cleaned up properly, there wouldn't have been any traces of it in valgrind. So, not seeing any change there isn't inherently conclusive. 

Which is why I also re-ran one of my OpenShot export memory-consumption tests (from OpenShot/openshot-qt#2811). First on both the latest `develop` branch code of both projects, again as a baseline, and then on a libopenshot with these changes applied.

The results speak for themselves: Memory consumption is _identical_ between the two runs. The only discernible deviations in the graphs are attributable to extra time I took in starting the export process, or terminating the application after it completed.

[memusage graph comparison.pdf](https://github.com/OpenShot/libopenshot/files/3898476/memusage.graph.comparison.pdf)

image

### What's going on here?

The secret, I suspected but am now completely certain, has to do with the fact that `QImage::bits()` is [an overloaded function](https://doc.qt.io/qt-5/qimage.html#bits-1), and in `const` context the `constBits()` form is _automatically_ used. So, Qt was automatically optimizing these calls on our behalf.

I'm still in favor of merging this, since it doesn't hurt to be more explicit in how we use QImage. Relying on context to choose which algorithm is used just makes the code harder to read. Thanks!

src/Frame.cpp
```diff
@@ -692,7 +692,7 @@ void Frame::Thumbnail(std::string path, int new_width, int new_height, std::stri
 
 		// Get pixels
 		unsigned char *pixels = (unsigned char *) thumbnail->bits();
-		unsigned char *mask_pixels = (unsigned char *) mask->bits();
+		unsigned char const *mask_pixels = (unsigned char *) mask->constBits();
```
ferdnyc: **This** change _will_ actually reduce the memory consumption during thumbnailing, since previously the context _wasn't_ `const`. So `QImage` would have been copying the mask bits here.

Minor nit: My preference would be for this more traditional ordering (especially since it makes the `const` far more obvious):
```suggestion
		const unsigned char *mask_pixels = (unsigned char *) mask->constBits();
```

src/Frame.cpp
```diff
@@ -922,7 +922,7 @@ std::shared_ptr<Magick::Image> Frame::GetMagickImage()
 		AddColor(width, height, "#000000");
 
 	// Get the pixels from the frame image
-	QRgb const *tmpBits = (const QRgb*)image->bits();
+	QRgb const *tmpBits = (const QRgb*)image->constBits();
```
ferdnyc: I'd prefer changing this one as well so it matches its cast, while we're in there:
```suggestion
	const QRgb *tmpBits = (const QRgb*)image->constBits();
```

musteresel:
Applied the suggested edits. Yes, it seems that this doesn't solve memory issues ... but as you said: being explicit doesn't hurt :)

PR #376, opened 2019-11-27
Minor adjustments to Doxygen API docs

ferdnyc:
- Define `USE_IMAGEMAGICK` and `USE_BLACKMAGIC` unconditionally when building docs, so that the classes will be documented.
- Improve handling of `std::`-prefixed types in doxygen output.


ferdnyc:
Merging as it only affects the docs.

PR #377, opened 2019-11-27
look for rc_buffer_size too

chad3814:
`rc_buffer_size` is missing from the `if()` but is in the clause. Need this to do crf + bit rate limiting.


chad3814:
ugh I had originally used master as the base for the commit, so then github told me there was a merge conflict but that I could edit it right there, alas I failed at doing that correctly... so 3 commits to add `name == "rc_buffer_size" || `...

ferdnyc:
Ooh, good catch! Makes sense to me, merging.

> 3 commits to add `name == "rc_buffer_size"`

No worries, it happens, and the commits are easy enough to squash on merge.

PR #379, opened 2019-12-02
Wrap parentheses around assignment in conditional

ferdnyc:
When warnings are enabled, the compiler will flag any lines of the form:
```c++
if (result = get_result_for(thing))
```
due to the possibility that it's either _meant_ to be an equality test, or will be _interpreted_ as an equality test by someone reading the code in the future.

The warnings can be silenced by adding a second set of parens, i.e.
```c++
if ((result = get_result_for(thing)))
```
or, as this PR makes it (for readability, at the end of the line):
```c++
if ( (result = get_result_for(thing)) )
```

Double parens are added to the three remaining places in the code which trigger that warning.

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -914,7 +914,7 @@ void Clip::SetJsonValue(Json::Value root) {
 
 			if (!existing_effect["type"].isNull()) {
 				// Create instance of effect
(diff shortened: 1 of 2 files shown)


ferdnyc:
Merging, as the change is source-code-cosmetic only.

PR #380, opened 2019-12-03
WIP: contributing proposed code changes

aopa-frogslayer:
I know this code is outdated and we spoke on the phone about some of these changes. I would just like to contribute these changes back.


include/ZmqLogger.h
```diff
@@ -29,17 +29,17 @@
 #define OPENSHOT_LOGGER_H
 
 
-#include "JuceLibraryCode/JuceHeader.h"
-#include <iostream>
-#include <iomanip>
-#include <fstream>
-#include <stdlib.h>
+// #include "JuceLibraryCode/JuceHeader.h"
```
musteresel: Hm ... this is not pretty. If it's commented out it should be removed. Same for the other similar changes here.

src/CMakeLists.txt
```diff
@@ -398,9 +398,10 @@ set(LIB_INSTALL_DIR lib${LIB_SUFFIX}) # determine correct lib folder
 
 # Install primary library
 INSTALL(  TARGETS openshot
-		ARCHIVE DESTINATION ${LIB_INSTALL_DIR}
-		LIBRARY DESTINATION ${LIB_INSTALL_DIR}
-		COMPONENT library )
+  RUNTIME DESTINATION ${LIB_INSTALL_DIR}
```
musteresel: Why put executables in the library directory?

src/Qt/PlayerPrivate.cpp
```diff
@@ -54,7 +55,7 @@ namespace openshot
 		// bail if no reader set
 		if (!reader)
 			return;
-
+
```
musteresel: Whitespace change, please remove :)

src/QtPlayer.cpp
```diff
@@ -156,7 +156,9 @@ void QtPlayer::Reader(ReaderBase *new_reader)
 	// Set new reader. Note: Be sure to close and dispose of the old reader after calling this
 	reader = new_reader;
 	p->reader = new_reader;
-	p->videoCache->Reader(new_reader);
+
+	Timeline *timelineReader = dynamic_cast<Timeline*>(new_reader);
```
musteresel: **Bad** idea. At least *check* that the return value is NON-NULL.

src/ZmqLogger.cpp
```diff
@@ -60,102 +60,102 @@ ZmqLogger *ZmqLogger::Instance()
...
-
-	} catch (zmq::error_t &e) {
-		cout << "ZmqLogger::Connection - Error binding to " << connection << ". Switching to an available port." << endl;
-		connection = "tcp://*:*";
-		publisher->bind(connection.c_str());
-	}
-
-	// Sleeping to allow connection to wake up (0.25 seconds)
-	usleep(250000);
+	// const genericscopedlock<criticalsection> lock(loggercriticalsection);
```
musteresel: Commented out code?

src/ZmqLogger.cpp
```diff
@@ -42,8 +42,8 @@ ZmqLogger *ZmqLogger::Instance()
 		m_pInstance = new ZmqLogger;
 
 		// init ZMQ variables
-		m_pInstance->context = NULL;
-		m_pInstance->publisher = NULL;
+		//m_pInstance->context = NULL;
```
musteresel: Commented out code?

src/CMakeLists.txt
ferdnyc: That's actually present in our _current_ build configs too, for Historical Reasons™.

(CMake on Windows writes DLL files to the `RUNTIME DESTINATION`, which is _meant_ to be the same directory as executables — but our packaging scripts expect them to be in the `${LIB_INSTALL_DIR}`. Because once upon a time, there was only one `DESTINATION` parameter, which had the same effect of lumping everything into the same place. ...I still hope to fix that one day, and have the Windows build put the DLL files in the `bin` directory where — as counterintuitive as it seems to those of us used to Unix norms — they actually do belong.)

jonoomph:
@aopa-frogslayer Looks like we'll need to clean up some of this, remove commented code, etc... and then try and resolve a bunch of these conflicts, so this PR is more readable. It's kinda hard to view in the current form, but I appreciate sharing the changes with the community! If there are good performance improvements hidden in here, I want 'em. :stuck_out_tongue_winking_eye:

ferdnyc:
This one, I think, is a bit too daunting to attempt conflict-resolution. Sorry!

PR #383, opened 2019-12-07
Move feature summary to root CMakeLists

ferdnyc:
PR #367 by @musteresel adds optional coverage reporting to the CMake build, which is controlled from the root `CMakeLists.txt` and `tests/CMakeLists.txt`.

Until now our FeatureSummary output has been done in the `src/CMakeLists.txt`, but it makes sense to list coverage reporting as a build feature. So, the FeatureSummary generation is moved to the bottom of the root `CMakeList.txt` where it can collect feature information from all build components.


ferdnyc:
Merging this, as it's just a small organizational change to the build.

PR #386, opened 2019-12-11
FFmpegWriter: Some fixes for building with libav

ferdnyc:
In #384 some issues were identified compiling libopenshot with `libav`, instead of `ffmpeg`. These adjustments make our source compatible with both.

- libav defines `AV_CODEC_ID_HEVC` but not `AV_CODEC_ID_H265`, so it makes sense to only use `_HEVC` in the source.

- `AVCodecContext->framerate` was actually added in FFmpeg 2.2, which corresponds to libavcodec 56.26. Our version check was using `LIBAVFORMAT_VERSION` instead of `LIBAVCODEC_VERSION`, and wasn't precise enough.

Fixes #384

PR #387, opened 2019-12-15
Ensure switch statements on enums cover all values

ferdnyc:
When compiling with `-Wall`, GCC will issue a warning whenever a `switch()` on an enum type doesn't account for all of the possible enum values, either through explicit `case` statements or by including a `default` case.

This PR silences all of those warnings, by ensuring that all enum values are covered in `switch()` statements.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -191,10 +191,10 @@ static enum AVPixelFormat get_hw_dec_format(AVCodecContext *ctx, const enum AVPi
 #if defined(__APPLE__)
 			// Apple pix formats
 			case AV_PIX_FMT_VIDEOTOOLBOX:
-					hw_de_av_pix_fmt_global = AV_PIX_FMT_VIDEOTOOLBOX;
-					hw_de_av_device_type_global = AV_HWDEVICE_TYPE_VIDEOTOOLBOX;
-					return *p;
-					break;
+				hw_de_av_pix_fmt_global = AV_PIX_FMT_VIDEOTOOLBOX;
+				hw_de_av_device_type_global = AV_HWDEVICE_TYPE_VIDEOTOOLBOX;
+				return *p;
+				break;
 #endif
 				// Cross-platform pix formats
 			case AV_PIX_FMT_CUDA:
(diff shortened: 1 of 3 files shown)


ferdnyc:
Huh, I don't see how this could POSSIBLY be causing failures, but the jobs fail in the tests with FFmpeg 3 for some reason. (FFmpeg 4 is errored offline due to the PPA issue I mentioned in #386.)

I'll poke around and see if I can figure out what's changed, perhaps as a side-effect.

ferdnyc:
_Crazy_. Heres' the complete diff between defeadf and bd4d2bf, which made the difference between the tests succeeding and not:

```diff
diff --git a/src/FFmpegWriter.cpp b/src/FFmpegWriter.cpp
index cccaf02..4765285 100644
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -1532,7 +1532,7 @@ void FFmpegWriter::write_audio_packets(bool is_final) {
 		ZmqLogger::Instance()->AppendDebugMethod("FFmpegWriter::write_audio_packets", "is_final", is_final, "total_frame_samples", total_frame_samples, "channel_layout_in_frame", channel_layout_in_frame, "channels_in_frame", channels_in_frame, "samples_in_frame", samples_in_frame, "LAYOUT_MONO", LAYOUT_MONO);
 
 		// Keep track of the original sample format
-		AVSampleFormat output_sample_fmt;
+		AVSampleFormat output_sample_fmt = audio_codec->sample_fmt;
 
 		AVFrame *audio_frame = NULL;
 		if (!is_final) {
@@ -1564,7 +1564,7 @@ void FFmpegWriter::write_audio_packets(bool is_final) {
 					break;
 				}
 				default: {
-					output_sample_fmt = audio_codec->sample_fmt;
+					// This is only here to silence unused-enum warnings
 					break;
 				}
 			}

```

Ohhhh, _now_ I see it. The `if (!is_final)` is the crucial bit — with my first change, when `is_final` was true `output_sample_fmt` would end up being unset. D'oh.

jonoomph:
Noice! Thanks again!

PR #388, opened 2019-12-15
Travis: allow FFmpeg4 build failures

ferdnyc:
As I noted in https://github.com/OpenShot/libopenshot/pull/386#issuecomment-564573049 :

The failed Travis check is due to our FFmpeg 4 build failing in the setup phase, as it's not able to install the ffmpeg 4 libs via apt. It seems the maintainer of the PPA for ffmpeg 4 on Bionic has chosen to [take his toys and go home](https://launchpad.net/~jonathonf/+archive/ubuntu/ffmpeg-4):

<blockquote>
This PPA has been removed from public access as part of a protest against the abuse of open-source projects by large companies. For more detail visit the main page here: <a href="https://launchpad.net/~jonathonf">https://launchpad.net/~jonathonf</a>

If you are a company and you would like this PPA to continue then let me know your preferred route for contributions and I will arrange something.

If we have already been in contact then ping me your Launchpad ID and I will add you to a private PPA in the meantime.
</blockquote>

🙄 Presumably some saner person will set up a replacement PPA, but until then it looks like we'll have to figure out whether we want to build FFmpeg 4 ourselves for the Travis environment, or disable the FFmpeg 4 branch of the CI builds.

...For now, allowing failures of the FFmpeg4 build (only) will get us back online as far as being able to develop and merge PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,11 @@ addons:
     - libswresample-dev
 
 matrix:
+
+  # The FFmpeg4 PPA is currently down as a protest
+  allow_failures:
+  - env: BUILD_VERSION=ffmpeg4
+
   include:
     - name: "FFmpeg 2 GCC (Ubuntu 16.04 Xenial)"
       env: BUILD_VERSION=ffmpeg2


ferdnyc:
Merging this, as it has the intended effect of restoring our CI to working order. It would be a very good thing to get FFmpeg 4 builds working again, but how to do that is a larger question that shouldn't halt all development.

eisneinechse:
I had the same problem. I compiled a version myself and included AV1 codecs that are not normally in there. Maybe what I did might help to get a self compiled version here working.

[Install_ffmpeg4.txt](https://github.com/OpenShot/libopenshot/files/3969003/Install_ffmpeg4.txt)

ferdnyc:
@eisneinechse 

Thanks! That looks like it could save me some time.

And that [cargo-c](https://dev.to/luzero/building-crates-so-they-look-like-c-abi-libraries-1ibn) tool the `rav1e` developer uses/built is interesting in its own right. It could potentially help with packaging up ReSVG as a dependency as well, since it currently has no "install"-type process at all. I should bounce that idea off RazrFalcon.

eisneinechse:
@ferdnyc 
A few points:
The SVT encoders want to be inside the ffmpeg directory but I wanted them outside. Therefore, I made a link as can bee seen in the text. And there are also very fast encoders for VP9 and HEVC (=h265) in SVT, easy to find on github; I just wanted this text to be easily readable.
Rav1e can be compiled for one architecture but when the library should be distributed it is better to not use this option (native) which I left out in this text.
I did some testing with the resulting ffmpeg with svt_av1 and rav1e and it works including the options. The libraries I did use to compile libopenshot and I tested openshot-qt too, it works but the options cannot be set yet. And that's what I was looking into when I started and realized that the ppa was gone. My plan would be to find a way to set this options in openshot but they are quite different from the h264 ones. Setting this options could be combined with implementing the interlaced export, finally.
And I too found the cargo-c tool extremely interesting.

ferdnyc:
@eisneinechse 
> And I too found the cargo-c tool extremely interesting.

Talk about missed opportunities, though — _**how**_ is it not called "cargo-ship"!??! **#DUH.** :wink:

eisneinechse:
@ferdnyc 

I guess it has historical reasons.
I can now change the quality settings for the av1 encoders. It is a hack that should under NO circumstances be used, just to check if I am right.
index 645c0cc..b1de96e 100644
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -462,6 +462,7 @@ void FFmpegWriter::SetOption(StreamType stream, std::string name, std::string va
                                        case AV_CODEC_ID_AV1 :
                                                c->bit_rate = 0;
                                                av_opt_set_int(c->priv_data, "crf", std::min(std::stoi(value),63), 0);
+                                               av_opt_set_int(c->priv_data, "qp", std::min(std::stoi(value),63), 0);
                                                break;
 #endif
                                        case AV_CODEC_ID_VP8 :

What we really need is a new user interface for the Advanced Settings-> Video. I think this would be a good time as we already change the UI. We need a way to set things other than CRF and Mb/s. The newer encoders have some other way to define the quality. (with the self compiled ffmpeg see: ffmpeg -help encoder=av1)
Also:
For SVT-AV1 to produce video files that a searchable ffmpeg has to have the option -forced-idr or SvtAv1EncApp has to have the option -irefresh-type 2 . I think the main direction SVT-AV1 is developed is for streaming not for producing stored videos.

ferdnyc:
@eisneinechse 
Word. Honestly I'd just love to get rid of that accordion-panel nonsense. I've come to view it as a _terrible_ UI (especially with only **one** field in the Image Sequence section), and it sucks so hard not being able to see related settings at the same time.

It's so tedious making export adjustments using that dialog (and having to re-do them _every_ time) that I'm actually tempted to say we should just make the JSON panel _editable_, let users compose Export parameters by directly manipulating the JSON, and just enhance our parsing a little to expect that some requests will be ill-formed and result in parser errors. (And I'm fine with just rejecting those outright, we don't have to work too hard trying to process bad requests.)

The _interface_ part of the dialog could still be there for those who prefer to make adjustments that way, but ultimately they'd be able to use the JSON representation to not only examine, but tweak — and ultimately maybe even store and reuse — their export parameters.

(If nothing else, they'd always have the option of copy-pasting their settings to a text editor, as a self-service save/load workaround.)

ferdnyc:
Annoyances like #2881 also contribute to the whole "death by a thousand paper cuts" experience, with the current Export dialog.

eisneinechse:
@ferdnyc 
Sorry I cannot control when I have time right now so there may be delays.
Yes, it is hard to implement all the new settings for the new codecs that way and there should be a way to come back to the settings used last time IMHO. Now should be the time to make the changes as the new codecs are now really coming to the costumers.
It might be interesting for this subject that now there is the first TV with AV1 support (LG ZX range) and the export with svt-av1 (modes 6 and up have no memory leak) or rav1d work when included in the ffmpeg libraries and a few small changes are made to openshot. But the settings are a bad hack with this UI. Unfortunately svt-vp9 is buggy right now and cannot be used which is a shame as there are many devices and services now that support VP9 (I lost a lot of time trying to make it work in openshot; I have a ffmpeg compiled with dav1d, rav1d, svt-av1, svt-hevc, and svt-vp9).
I also think that we should include these implementations if we compile our own ffmpeg4.
Also, editing 4k videos needs extremely fast memory paths (RAM, HD, GPU). That's from my own experience and now I have seen more and more sources saying the same. It is no wonder 4k 60 fps editing sucks in openshot when one has a normal fast computer; one needs PCIe x16 speed for the connections, like in raid0 with four 3.6GB/s(read) / 3GB/s(write) NVMe SSDs. Something I don't have.

PR #389, opened 2019-12-15
Don't compare differently-signed types

ferdnyc:
A frequent cause of complaining from `gcc -Wall` is comparison between types with different signedness. For instance, both of these will produce a warning:

```c++
std::vector<openshot::Point> points = {};
for (int i = 0; i < points.size(); i++) {
  // stuff
}

Json::Value root = {};
for (int j = 0; j < root["clips"].size(); j++) {
  // json stuff
}
```

Fortunately, both are easily solved:
```c++
std::vector<openshot::Point> points = {};
for (std::vector<openshot::Point>::size_type i = 0; i < points.size(); i++) {
  // stuff
}

Json::Value root = {};
for (Json::Value::ArrayIndex j = 0; j < root["clips"].size(); j++) {
  // json stuff
}
```

This PR applies similar solutions to all comparisons which cause a warning to be emitted.

### DIFF
--- a/src/AudioBufferSource.cpp
+++ b/src/AudioBufferSource.cpp
@@ -43,7 +43,7 @@ AudioBufferSource::~AudioBufferSource()
 {
 	// forget the AudioSampleBuffer. It still exists; we just don't know about it.
 	buffer = NULL;
-};
+}
 
 // Get the next block of audio samples
 void AudioBufferSource::getNextAudioBlock (const juce::AudioSourceChannelInfo& info)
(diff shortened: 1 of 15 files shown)


jonoomph:
@ferdnyc Love the new loop syntax!

ferdnyc:
@jonoomph

Yeah, I could've made this a spin-off of my Awkward Code series, "Don't WRITE Awkward Code". C++11 finally lets us leave management of loop variables to the compiler, since that's exactly the sort of job that it should be doing for us.

Do we give up the ability to do weird, magical things with odd increments and confusing typecasts? Well, no, actually, we _could_ still do that stuff if we wanted to. But we **shouldn't**!

Issue #272, opened 2019-07-11
FFmpegReader.cpp CreateFrame need lock


laochen:
CreateFrame need lock, avoid  working_cache  one requested_frame Add once more 
like this：
```
std::shared_ptr<Frame> FFmpegReader::CreateFrame(int64_t requested_frame) {
	// Check working cache
	std::shared_ptr<Frame> output = working_cache.GetFrame(requested_frame);	
	if (!output) {
		// Lock
		const GenericScopedLock <CriticalSection> lock(processingCriticalSection);	
		output = working_cache.GetFrame(requested_frame);
		if(output) return output;
		// Create a new frame on the working cache
		output = std::make_shared<Frame>(requested_frame, info.width, info.height, "#000000", Frame::GetSamplesPerFrame(requested_frame, info.fps, info.sample_rate, info.channels), info.channels);
		output->SetPixelRatio(info.pixel_ratio.num, info.pixel_ratio.den); // update pixel ratio
		output->ChannelsLayout(info.channel_layout); // update audio channel layout from the parent reader
		output->SampleRate(info.sample_rate); // update the frame's sample rate of the parent reader

		working_cache.Add(output);

		// Set the largest processed frame (if this is larger)
		if (requested_frame > largest_frame_processed)
			largest_frame_processed = requested_frame;
	}

	// Return new frame
	return output;
}
```

ferdnyc:
@eisneinechse , can I trouble you for a consult? Sounds like your area of expertise.

@laochen caught a leak in my updated JsonCPP code (#271), leading to my penance in #273. 👍

eisneinechse:
@ferdnyc Sure, there might be delays though as I might not be online all the time.
It all looks very logical. And if valgrind is fine with it then it is a good start. Valgrind is slow but catches a lot. I would say the patch works. Only the real world, Openshot, will tell if I am wrong.
(BTW thw memory leek  for the new calls in ffmpeg 4 are also identified correctly by valgrins, unfortunately I haven't had time to find a solution but I should have more time very soon.)

ferdnyc:
@eisneinechse *nod* I saw that as well, running the test executable back through valgrind after fixing the `Json::CharReader` leak. Some of them got pretty sizeable, too. Since `valgrind` is nice enough to sort the leaks by size, the last one kind of jumps out at you. :laughing: 
<pre>
==9297== <b>22,118,423 bytes</b> in 1 blocks are possibly lost in loss record 527 of 527
[...]

PR #391, opened 2019-12-19
fix incorrect buffer size being passed to avcodec_fill_audio_frame

ckirmse:
- most of the time the fixed number being sent in was large enough,
  but not always
- also added logging if avcodec_fill_audio_frame fails

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -1482,7 +1482,8 @@ void FFmpegWriter::write_audio_packets(bool is_final) {
 		ChannelLayout channel_layout_in_frame = LAYOUT_MONO; // default channel layout
 
 		// Create a new array (to hold all S16 audio samples, for the current queued frames
-		int16_t *all_queued_samples = (int16_t *) av_malloc((sizeof(int16_t) * (queued_audio_frames.size() * AVCODEC_MAX_AUDIO_FRAME_SIZE)));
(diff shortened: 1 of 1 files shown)


ferdnyc:
I got all excited because I was hoping this might be the solution to some of the invalid reads `valgrind` has been flagging (see https://github.com/OpenShot/libopenshot/issues/272#issuecomment-511258961), which live in `write_audio_packets()`.

Unfortunately, they're still there with this code applied, so that's not the case. Still, it looks like a sensible change to me.

ckirmse:
Yeah this is a different one, but it also causes a crash on certain inputs.ChrisOn Dec 19, 2019 3:52 AM, Frank Dana <notifications@github.com> wrote:I got all excited because I was hoping this might be the solution to some of the invalid reads valgrind has been flagging (see #272 (comment)), which live in write_audio_packets().
Unfortunately, they're still there with this code applied, so that's not the case. Still, it looks like a sensible change to me.

ckirmse:
@ferdnyc OK I'd appreciate help getting this merged :) To maybe grease the skids, I have a possible solution for your #272. The buffer size being allocated is correct, but sometimes nb_samples is shorter than the code expects (see https://stackoverflow.com/questions/39587839/libswresample-swr-convert-not-producing-enough-samples )

I think the fix is moving the initialization of remaining_frame_samples to just after the SWR_CONVERT call and setting it to this:
`
// Set remaining samples
remaining_frame_samples = nb_samples * av_get_bytes_per_sample(AV_SAMPLE_FMT_S16);
`

This fixed the valgrind issues in my testing.

jonoomph:
LGTM! Thanks for this!

PR #393, opened 2019-12-19
ColorShift effect: Correct class_name

ferdnyc:
The `info.class_name` was set to `"Color Shift"`, an unlikely-seeming class name considering it contains a space character.


jonoomph:
Good catch!

PR #394, opened 2019-12-19
EffectBase: Remove unused short_name

ferdnyc:
Not a single effect uses `info.short_name`, it is _always_ set to the empty string. And not a single consumer of the class or its derived classes (in either libopenshot or OpenShot) ever looks at it.

### DIFF
--- a/include/EffectBase.h
+++ b/include/EffectBase.h
@@ -50,7 +50,6 @@ namespace openshot
 	struct EffectInfoStruct
 	{
 		std::string class_name; ///< The class name of the effect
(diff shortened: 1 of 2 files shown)


jonoomph:
I guess we weren't ever using this? Hmmm....

jonoomph:
Good bye short_name

ferdnyc:
@jonoomph
> I guess we weren't ever using this? 

Not as far as I can tell. It's kind of redundant with `class_name`, I suppose. No real need for THREE names per effect.

PR #395, opened 2019-12-25
Skip painter transform for the Clip when video disabled

SuslikV:
Attempt to fix the: https://github.com/OpenShot/openshot-qt/issues/2882

**Edit:** as soon as PR was updated the info below is not actual anymore.

It is hard to read it as git differences/changed but I have added only:
```	// Clip that forced no video should has zero size output
	if (source_clip->has_video.GetInt(clip_frame_number) == 0) {
		source_size = QSize(0, 0);

		// Debug output
		ZmqLogger::Instance()->AppendDebugMethod("Timeline::add_layer (has_video = 0)", "source_frame->number", source_frame->number, "source_width", source_size.width(), "source_height", source_size.height());
	}
	else

```
before the `switch()`. I tried other formatting as well, but git changes still shown apart...

### DIFF
--- a/src/Timeline.cpp
+++ b/src/Timeline.cpp
@@ -379,8 +379,9 @@ void Timeline::add_layer(std::shared_ptr<Frame> new_frame, Clip* source_clip, in
 
 	}
 
-	// Skip out if only an audio frame
-	if (!source_clip->Waveform() && !source_clip->Reader()->info.has_video)
+	// Skip out if video was disabled or only an audio frame (no visualisation in use)
+	if (source_clip->has_video.GetInt(clip_frame_number) == 0 ||
+	    (!source_clip->Waveform() && !source_clip->Reader()->info.has_video))
 		// Skip the rest of the image processing for performance reasons
 		return;
 


ferdnyc:
Heh, yeah the formatting changes are a mess, but that's mostly because the original code is a mess (formatting-wise).

This change looks OK, but looking over the code I'm wondering if this can be solved even more simply (and with better performance) using the same test you introduced, but using it to short-circuit even more of the process.

Setting the `source_size` to `0×0` will take care of the scaling error, but all of the translation, shearing, rotation, compositing, etc. code — and even scaling — will still get executed, and... what for? There's literally nothing for it to process, but it'll still go through all of the unnecessary computations.

What if, instead, this part higher up:
https://github.com/OpenShot/libopenshot/blob/a1158ee278e6f66e7aacff77e9b46244eee44b19/src/Timeline.cpp#L382-L385
was changed to:
```c++
// Skip out if video is disabled or not present
if (!source_clip->Waveform() && 
    (source_clip->has_video.GetInt(clip_frame_number) == 0
     || !source_clip->Reader()->info.has_video))
	// Skip the rest of the image processing for performance reasons
	return;
```

(It might even be possible to combine the second and third conditions, if `has_video.GetInt()` will return `0` whenever a clip has no video at all. But since the default is `-1` I'm not sure about that. But if it's smart enough, then that could just be...)
```c++
if (source_clip->has_video.GetInt(clip_frame_number) == 0 && !source_clip->Waveform())
```

SuslikV:
In current implementation of libopenshot the has_video is not const. It gives to user explicit control over the video stream. You can disable it when ever you wish. If source clip selected to display waveform of the sound then the code (with `not` at second variable):
```
source_clip->has_video.GetInt(clip_frame_number) == 0 && !source_clip->Waveform()
```
wouldn't work. The `has_video` as keyframe shouldn't be masked by any other variable, I think.

My thoughts was here such as: the `source_clip->Reader()->info.has_video` points to the presence of the video stream in the source (not all imported files has video).
And original part:
https://github.com/OpenShot/libopenshot/blob/a1158ee278e6f66e7aacff77e9b46244eee44b19/src/Timeline.cpp#L382-L385
only means if source lacks the both video stream and user didn't set the waveform (to display), then there is nothing to process for transform. While simple sound wave visualization (if enabled) still can be rotated etc.

Probably, it is possible to cut earlier but not exactly as proposed above. Do you have any better solution?

ferdnyc:
> The `has_video` as keyframe shouldn't be masked by any other variable, I think.

Mmm, OK, yeah since `source_video->Waveform()` is clip-wide (non-keyframed), we'd still want `source_clip->has_video` to override if it's explicitly set to `0`. That way, they can use `Waveform()` and still control the visibility of the waveform display with "Enable Video" keyframes if they so choose.

Well... in that case, we can keep the existing condition as-is, and just _add_ the keyframe check to it:
```c++
if (source_clip->has_video.GetInt(clip_frame_number) == 0 ||
    (!source_clip->Waveform() && !source_clip->Reader()->info.has_video))
{
    // Skip the rest of the image processing for performance reasons
    return;
}
```
Point is, the `source_clip->has_video.GetInt(clip_frame_number) == 0` case is the same as the no-video, no-waveform case: Once the audio data is handled, there's nothing more that needs to be done for this particular frame, so `add_layer()` can just terminate early.

ferdnyc:
> for example,
> `source_clip->has_video.GetInt(clip_frame_number) == 0` is _false_ (user don't mind to display video or not in the current frame)
> then without the `!source_clip->Reader()->info.has_video` check - all transform code will be still processed no matter is there video in the file (Clip) or not.

Indeed. Of course, that's only true in the first place because the code insists on creating "dummy" video / audio data for frames even when they _don't have_ any video or audio data. If `Reader->info.has_video == 0` _actually_ meant that there was no video in the Frame returned by `clip->GetFrame(num)`, then we could just check that.

Huh... in fact, `openshot::Frame` actually _has_ `has_image_data` and `has_audio_data` members. What's more, turns out `Clip::GetFrame(num)` actually _already checks_ `has_video.GetInt(num)` before copying the video image into the Frame:
https://github.com/OpenShot/libopenshot/blob/a1158ee278e6f66e7aacff77e9b46244eee44b19/src/Clip.cpp#L302-L312
https://github.com/OpenShot/libopenshot/blob/a1158ee278e6f66e7aacff77e9b46244eee44b19/src/Clip.cpp#L325-L340

The **PROBLEM** is that the frame is initialized with:
```c++
new Frame(new_frame_number, 1, 1, "#000000", original_frame->GetAudioSamplesCount(), original_frame->GetAudioChannelsCount())
```
Which means `frame->has_image_data` and `frame->has_audio_data` will **ALWAYS** be true, because even when the clip's frame data isn't copied in, it's filled with a `1px × 1px` black square and a silent audio waveform as its initial data.

What's more, there are no methods at all to _remove_ video or audio data from a frame. There's `Frame::ClearWaveform()`, to delete the _that_ image, but that's the only one. There's no `Frame::RemoveImage()` to go with `Frame::AddImage()`, there's no `Frame::RemoveAudio()` to go with `Frame::AddAudio()`, and there's absolutely no way for `has_image_data` or `has_audio_data` to ever become `false` again, after they've been set to `true`.

I still don't understand the thinking behind that. It seems like it'd be cleaner if we could use the _presence_ of video or audio data in the frame to control whether or not those elements are processed at all. If `has_image_data == false`, no video processing. If `has_audio_data == false`, no audio processing.

ferdnyc:
Hah! Oh, wow. And I just found a bug in the Frame copy routine:
https://github.com/OpenShot/libopenshot/blob/a1158ee278e6f66e7aacff77e9b46244eee44b19/src/Frame.cpp#L104-L113

I'm gonna take a leap and say that should probably be:
```c++
// Copy data and pointers from another Frame instance
void Frame::DeepCopy(const Frame& other)
{
	number = other.number;
	channels = other.channels;
	width = other.width;
	height = other.height;
	channel_layout = other.channel_layout;
	has_audio_data = other.has_audio_data;
	has_image_data = other.has_image_data;
```

Of course, since like I said they're _always_ true, in practice that bug probably doesn't matter at all. :unamused:

ferdnyc:
> Which means `frame->has_image_data` and `frame->has_audio_data` will **ALWAYS** be true, because even when the clip's frame image isn't copied in, it's filled with a `1px × 1px` black square and a silent audio waveform as its initial data.

Hm! No, I take that back. Even though those values are passed in, `has_image_data` and `has_audio_data` are initialized to `false`, and they won't get set true unless `AddColor()`, `AddImage()`, `AddAudio()`, or `AddAudioSilence()` are called.
https://github.com/OpenShot/libopenshot/blob/a1158ee278e6f66e7aacff77e9b46244eee44b19/src/Frame.cpp#L74-L85

Of course, `Frame::GetImage()` _calls_ `AddColor()` if there's no image pointer, which will _set_ `has_image_data` true. ARRRRRRGH.
https://github.com/OpenShot/libopenshot/blob/a1158ee278e6f66e7aacff77e9b46244eee44b19/src/Frame.cpp#L904-L913

SuslikV:
Nice findings! Indeed, the `Frame::DeepCopy` should be fixed. (I hope you will fill PR for this!)

As for the original issue and earlier cut, I just want to be sure now, that the clip has (0, 0) size and nothing will be shown even if rendered.

SuslikV:
Logic to resolve the issue was changed and includes proposed optimizations now.

------
Also, while experimenting with the setting the field `Enable Video` of the clip Properties in OpenShot editor, found a bug that makes loops in keyframes setting (not ordered keyframes):
```
'has_video': 
{'Points': [
{'co': {'X': 1.0, 'Y': -1.0}, 'interpolation': 2},
{'co': {'X': 9, 'Y': 0}, 'interpolation': 1},
{'co': {'X': 4, 'Y': 1}, 'interpolation': 1}
]}
```
Don't know if it was reported earlier.

src/Timeline.cpp
```diff
@@ -379,8 +379,8 @@ void Timeline::add_layer(std::shared_ptr<Frame> new_frame, Clip* source_clip, in
 
 	}
 
-	// Skip out if only an audio frame
-	if (!source_clip->Waveform() && !source_clip->Reader()->info.has_video)
+	// Skip out if video was disabled or only an audio frame (no visualisation in use)
+	if (source_clip->has_video.GetInt(clip_frame_number) == 0 || !source_clip->Waveform() && !source_clip->Reader()->info.has_video)
```
jonoomph: Are we missing some parenthesis?
`if (source_clip->has_video.GetInt(clip_frame_number) == 0 || (!source_clip->Waveform() && !source_clip->Reader()->info.has_video))`

SuslikV: Is it required change? If indentation needed I can make it look like this:
```
if (source_clip->has_video.GetInt(clip_frame_number) == 0 ||
    !source_clip->Waveform() && !source_clip->Reader()->info.has_video)
```
Maybe it will be easier to read than additional pair of parentheses.

SuslikV:
Just updated the code listing (two rows instead of one to make the second condition more noticeable).

jonoomph:
Thx, but I would prefer we add parenthesis around the final 2 conditions, to disambiguate them, as I included in my example above.

jonoomph:
LGTM! Thx

jonoomph:
Did this merge the wrong commit? Hmmm

SuslikV:
Cosmetic changes (second condition included in parentheses).

Operator precedence is fundamental rule of the programming language, hardly believable that it will be changed anytime soon.

jonoomph:
Github is acting strange... I think it was merged into `develop` correctly.

SuslikV:
It has number of history modifications. I rebased it each time during edit but didn't update it on the top of the current `develop` branch (because this file wasn't modified - all merging was OK).

PR #397, opened 2019-12-27
Frame: Copy has_audio_data correctly in DeepCopy

ferdnyc:
Fix a typo I spotted in https://github.com/OpenShot/libopenshot/pull/395#issuecomment-568930523.

### DIFF
--- a/tests/Frame_Tests.cpp
+++ b/tests/Frame_Tests.cpp
@@ -46,7 +46,7 @@ TEST(Default_Constructor)
 	// Create a "blank" default Frame
 	std::shared_ptr<Frame> f1(new Frame());
 
-	REQUIRE CHECK(f1 != nullptr);  // Test aborts here if we didn't get a Frame
+	CHECK(f1 != nullptr);  // Test aborts here if we didn't get a Frame
 
 	// Check basic default parameters
 	CHECK_EQUAL(1, f1->GetHeight());
@@ -61,7 +61,7 @@ TEST(Default_Constructor)
 	// Calling GetImage() paints a blank frame, by default
 	std::shared_ptr<QImage> i1 = f1->GetImage();
 
-	REQUIRE CHECK(i1 != nullptr);
+	CHECK(i1 != nullptr);
 
 	CHECK_EQUAL(true,f1->has_image_data);
 	CHECK_EQUAL(false,f1->has_audio_data);
@@ -79,7 +79,7 @@ TEST(Data_Access)
 	// Get first frame
 	std::shared_ptr<Frame> f1 = c1.GetFrame(1);
 
-	REQUIRE CHECK(f1 != nullptr);
+	CHECK(f1 != nullptr);
 
 	CHECK_EQUAL(1, f1->number);
 	CHECK_EQUAL(1280, f1->GetWidth());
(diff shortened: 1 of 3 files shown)


SuslikV:
I didn't found `DeepCopy` usage in openshot-qt (at least from my tests), and overall logic of the changes seems to be OK. Should be merged.

ferdnyc:
> I didn't found `DeepCopy` usage in openshot-qt (at least from my tests),

Not directly, but it is used in the assignment operator and copy constructor for the Frame class, which is probably used in libopenshot.

Which is why I decided to at least start on some unit tests for the class, and add them in to the PR.

Didn't get very far in before I seem to have bumped up against another bug related to DeepCopy.

tests/Frame_Tests.cpp
```diff
@@ -0,0 +1,150 @@
...
+	CHECK_EQUAL(par1.num, par2.num);
+	CHECK_EQUAL(par1.den, par2.den);
+
+
+	CHECK_EQUAL(f1.SampleRate(), f2.SampleRate());
+	CHECK_EQUAL(f1.GetAudioChannelsCount(), f2.GetAudioChannelsCount());
+	CHECK_EQUAL(f1.ChannelsLayout(), f2.ChannelsLayout());
+
+	CHECK_EQUAL(f1.GetBytes(), f2.GetBytes());
+	CHECK_EQUAL(f1.GetAudioSamplesCount(), f2.GetAudioSamplesCount());
```
ferdnyc: The last line of this test function will fail, typically with a message like this:
```
----------------------------
     RUNNING ALL TESTS
----------------------------
[libvpx @ 0xa1c380] v1.8.1
/home/ferd/rpmbuild/REPOS/libopenshot/tests/Frame_Tests.cpp:147:1: error:
Failure in Copy_Constructor: Expected 0 but was -65536
FAILURE: 1 out of 92 tests failed (1 failures).
Test time: 38.86 seconds.
----------------------------
```
It seems when the copy constructor sets up the new object, it doesn't do anything to ensure that the copied class will have sane data for all of its accessor methods.

`GetAudioSamplesCount()` just blindly returns the value of the `max_audio_sample` class member. That value isn't copied or otherwise handled by `openshot::Frame::DeepCopy()`, so there's nothing to ensure that it would receive the proper value. (And it clearly doesn't.)

ferdnyc:
Ugh. Although, on Ubuntu Xenial `Frame_Tests.cpp` failed to compile, because it seems its UnitTest++ is too old to support `REQUIRE` in tests. I'll have to drop those.

SuslikV:
Now you did it even worse. If first commit just was the quick typofix for the logic. The next are resolving nothing but adding new feature - new tests.

> Not directly, but it is used in the assignment operator and copy constructor for the Frame class, which is probably used in libopenshot.

I made my checks by adding debug message into the `DeepCopy()` method and using OpenShot (edit & export). And didn't found evidence of its usage. So, let's wait for the issue to be filled first or maybe this feature require review and elimination from the code, completely. My thoughts.

jonoomph:
Yup, I'm still seeing the failed unit test.

jonoomph:
So, in theory, we just need to copy `max_audio_sample` in the copy constructor, and this error should go away.

jonoomph:
Unittests now work correctly! Merging this. Thanks @ferdnyc and @SuslikV !

PR #398, opened 2019-12-27
correctly calculate remaining_frame_samples

ckirmse:
- sometimes the output from the SWR_CONVERT is not the exact amount we asked for
- without this fix, the old code would sometimes read past the end of a buffer
- valgrind complained about the old code
- read https://stackoverflow.com/questions/39587839/libswresample-swr-convert-not-producing-enough-samples
- @ferdnyc this is my holiday present to you :)

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -1417,7 +1417,7 @@ void FFmpegWriter::open_video(AVFormatContext *oc, AVStream *st) {
 				// unless "qp" was set for CQP, switch to VBR RC mode
 				av_opt_set(video_codec->priv_data, "rc_mode", "VBR", 0);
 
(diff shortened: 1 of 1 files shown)


SuslikV:
> sometimes the output from the SWR_CONVERT is not the exact amount we asked for

Is this was fixed by the proposed code changes?

> without this fix, the old code would sometimes read past the end of a buffer

How it acts for the OpenShot processing? Noises, crashes etc. Conditions to reproduce the issue itself maybe? (info to track the corresponding issues if any)

ckirmse:
If you run the tests with valgrind, it reports the invalid memory read. Yes, this change properly handles the return code of swr_convert and fixes the invalid memory read.

jonoomph:
Nice! Imma merge this also, and do some additional testing after it's merged. Thanks again!

PR #410, opened 2020-01-08
Fix Resvg image format

ferdnyc:
resvg should be rendering to `QImage::Format_ARGB32_Premultiplied`, according to @RazrFalcon.

Also, tightened up the code a bit. Works fine on my system, but the AppImage will be the ultimate test.

PR #411, opened 2020-01-08
Set IMPORTED_NO_SONAME on RESVG target

ferdnyc:
Prevents the library from being linked with its full path, instead of letting `ld` locate it at runtime.


ferdnyc:
Closing as this is included in #410 .

PR #412, opened 2020-01-09
Travis: Add libjsoncpp-dev to apt package list

ferdnyc:
Even though we bundle a copy of jsoncpp, it makes more sense to test our builds against the distros' system packages. Our copy is static, whereas the system packages could be updated. If that happens, testing in CI will hopefully help us catch any incompatibilities that get introduced (from either end).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,7 @@ matrix:
           packages:
           - *ff_common
           - qt5-default
+          - libjsoncpp-dev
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
Decided to split the difference: FFmpeg 3 and 4 GCC builds will use system jsoncpp, the other builds (including Clang FFmpeg 3) will continue to use the bundled copy.

PR #415, opened 2020-01-12
Use Codecov.io for coverage reporting

ferdnyc:
[Codacy](https://codacy.com) shows some promise, and I'd still like to use it eventually because it does a whole lot of complex tracking for repo changes, and as noisy and alarmist as it can be, for the most part I like its static analysis reporting.

But, so far setting coverage reporting up hasn't worked, the reports just don't seem to be getting there. meanwhile, setting the repo up at [codecov.io](https://codecov.io/) was a total breeze, it Just Worked™ right away. So, let's funnel our reporting there for now. It's something, at least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ addons:
     - qtmultimedia5-dev
     - doxygen
     - graphviz
+    - curl
     packages: &ff_common  # Common set of FFmpeg packages
     - *p_common
     - libfdk-aac-dev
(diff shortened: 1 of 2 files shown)


ferdnyc:
(See? Neato! Just needs some config tweaking...)

.travis.yml
```diff
@@ -94,6 +80,23 @@ matrix:
           - libavresample4
           - libswresample3
```
ferdnyc: Because Travis plays favorites, tending to kick off matrix jobs in the order they're listed in the Travis configs, having our _least_ critical (and quickest) build listed **first** was causing unnecessary delays even _starting_ the later, more-critical ones.

Now, the Coverage build is listed first (because it requires extra processing time to generate and upload the report data), then the FFmpeg 4 build, followed by FFmpeg3 (GCC, Clang), and finally FFmpeg 2. Because nobody really cares that much anymore, whether we can still build libopenshot with FFmpeg 2.

ferdnyc:
Merging, as that'll hopefully then make this commit the source of the baseline coverage report it keeps complaining about not having. (Also, it affects nothing but the repo metasources.)

PR #416, opened 2020-01-14
Raise preview cache to CPUs*8 frames, max 64

ferdnyc:
At least on my 4-core, far from high-performance system, the effect this has on playback stuttering and smoothness is _immense_.

With a 32-frame cache, even with debug logging enabled the preview is _largely_ smooth, at least when playing a Timeline with a single, unedited clip on it. With only 8 frames of buffer to work with, playback is jerky and stuttering even in that simple, best-case scenario.


jonoomph:
LGTM!

PR #421, opened 2020-01-31
FFmpegWriter: Overload Set___Options() methods

ferdnyc:
Add overloaded forms of `SetVideoOptions()` and `SetAudioOptions()` that apply some sensible defaults to rarely-changed parameters.

### DIFF
--- a/include/FFmpegReader.h
+++ b/include/FFmpegReader.h
@@ -76,11 +76,11 @@ namespace openshot {
 	 *
 	 * @code
 	 * // Create a reader for a video
-	 * FFmpegReader r("MyAwesomeVideo.webm");
+	 * openshot::FFmpegReader r("MyAwesomeVideo.webm");
 	 * r.Open(); // Open the reader
 	 *
 	 * // Get frame number 1 from the video
-	 * std::shared_ptr<Frame> f = r.GetFrame(1);
+	 * std::shared_ptr<openshot::Frame> f = r.GetFrame(1);
 	 *
 	 * // Now that we have an openshot::Frame object, lets have some fun!
 	 * f->Display(); // Display the frame on the screen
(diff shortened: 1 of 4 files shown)


ferdnyc:
As auto-documented in the Python bindings, the forms are now:
```python
>>> import openshot
>>> help(openshot.FFmpegWriter)
Help on class FFmpegWriter in module openshot:

class FFmpegWriter(WriterBase)
 |  FFmpegWriter(path)
 |  
 |  Proxy of C++ openshot::FFmpegWriter class.
 |  
 |  SetAudioOptions(self, *args)
 |      SetAudioOptions(FFmpegWriter self, bool has_audio, std::string codec, int sample_rate, int channels, openshot::ChannelLayout channel_layout, int bit_rate)
 |      SetAudioOptions(FFmpegWriter self, std::string codec, int sample_rate, int bit_rate)
 |  
 |  SetVideoOptions(self, *args)
 |      SetVideoOptions(FFmpegWriter self, bool has_video, std::string codec, Fraction fps, int width, int height, Fraction pixel_ratio, bool interlaced, bool top_field_first, int bit_rate)
 |      SetVideoOptions(FFmpegWriter self, std::string codec, Fraction fps, int width, int height, int bit_rate)
 |  
```

ferdnyc:
I thought about it some more, and decided to reorder the arguments to the overloaded `setVideoOptions()` to be consistent with the Timeline constructor (having the FPS value _after_ width and height), even though it makes it inconsistent with the other signature. It just feels more natural.

But I'm open to arguments that it's a bad idea or will cause problems, if anyone thinks so...

I also did some cleanup on the docs/example code.

jonoomph:
I have no problems with this at all. We still have the original method signatures, so old code will still work, and we have more consistant signatures moving forward. I like it.

Issue #420, opened 2020-01-31
Error while writing raw video frame


jeffski:
I am getting this error when rendering a reasonably complex video, it has video, images, about 39 clips, a few simple keyframe animations (zoom effect on images, and its about 2 minutes long. I am getting this error consistently for a particular sequence of clips. When I remove the last clip it works, but if I swap that clip with a different video it also fails. I have tried a lot of combinations basically to try and narrow down what the cause might be.

I realise that this is caused by a complex piece of code and there are probably several variables that could cause the issue but perhaps someone has seen this before, knows a bit more about where to start looking, likely culprits, etc... or maybe there are some settings in libopenshot that can reduce the chance of the error occurring.

Below is the error from my application:

```
terminate called after throwing an instance of 'openshot::ErrorEncodingVideo'
  what():  Error while writing raw video frame
Caught signal 6 (SIGABRT)

  /lib64/libc.so.6               ( abort                                     + 0x148 )  [0x7ff7a8d73f88]
  /usr/bin/lib/libstdc++.so.6 ( __gnu_cxx::__verbose_terminate_handler()  + 0x11f )  [0x7ff7a96d20ef]
  /usr/bin/lib/libstdc++.so.6 (                                           + 0x8fe86)  [0x7ff7a96cfe86]
  /usr/bin/lib/libstdc++.so.6 (                                           + 0x8fed1)  [0x7ff7a96cfed1]
  /usr/bin/lib/libstdc++.so.6 (                                           + 0x90104)  [0x7ff7a96d0104]
  /usr/bin/lib/libopenshot.so.17 ( openshot::FFmpegWriter::write_queued_frames()  + 0x4cd )  [0x7ff7ae17f459]
  /usr/bin/lib/libopenshot.so.17 ( openshot::FFmpegWriter::WriteFrame(std::shared_ptr<openshot::Frame>)  + 0x53e )  [0x7ff7ae17ee0c]
  /usr/bin/lib/libopenshot.so.17 ( openshot::FFmpegWriter::WriteFrame(openshot::ReaderBase*, long, long)  + 0x333 )  [0x7ff7ae17f8cd]
  /usr/bin/myapplication       (                                                   )  [0x40fd08]
  /usr/bin/myapplication       (                                                   )  [0x4049de]
  /lib64/libc.so.6               ( __libc_start_main                         + 0xea  )  [0x7ff7a8d6002a]
  /usr/bin/myapplication       (                                                   )  [0x404e5a]
```

[...]

PR #422, opened 2020-02-01
ZmqLogger: Add optional dumping to stderr

ferdnyc:
Add an `openshot::Settings` boolean `DEBUG_TO_STDERR`, which when set `true` will direct all `AppendDebugMethod()` messages to `std::clog`, in addition to (possibly) being sent over the ZeroMQ wire.

Set to WIP because I want to add unit tests and check it over a bit more, right now this is a quick and dirty proof-of-concept implementation intended for #420.

Also:
* Replaced `usleep()` call with the C++11 `std::this_thread::sleep_for()`
* Added `std::` prefixes throughout `ZmqLogger.cpp`, and removed the `using namespace std;`
* Corrected some cut-and-paste comment errors in `Settings.h`
* Reformatted a bunch of code for readability / differing tab widths


jonoomph:
@ferdnyc Is this PR ready for merge? It seems complete to me, but it still has the WIP decorator. If it's ready, slam that merge button! :+1:  Thx.

ferdnyc:
@jonoomph 

I think it is ready. I had originally wanted to add unit tests, but I've since reconsidered given the issues I've already encountered trying to unit-test singletons. (Plus, I think we'd need a more full-stack unit test framework than just UnitTest++, to **test** output to stderr in any meaningful way.)

I'll see what state it's in and finish up whatever's needed, it would be good to have. I also wanted to integrate the functionality into OpenShot somehow, and with `--debug` support merged things are in a much more conducive state over on that side.

jonoomph:
Thanks, LGTM!

PR #423, opened 2020-02-03
Optimized, parallelized Blur effect

ferdnyc:
This PR optimizes the Blur effect (by far our slowest effect), streamlining the code and (most importantly) adding OMP `#pragma`s to multi-thread the execution where the operations can be cleanly parallelized.

<strike>It still won't win any speed trophies (in fact, it's still slow as _fuck_), but testing on my 4-core Intel i5 system, during an export of an 11-second Big Buck Bunny clip with a default-settings Blur applied:
* CPU usage went up from 25% to 85%-95%
* Export rate estimation increased from 0.57 FPS to > 1.2 FPS, making the new code more than twice as fast
* I don't know what the full export time would have been for the old code because I gave up and canceled after 2 minutes when it was only at 20%. But with the new code, the full export completed in `4m37s` — so, again, more than twice as fast.
* Results appear identical, at least to my eye.</strike>

I took a second run at this, after re-re-rereading the OpenMP docs, and improved the parallelization significantly. Now, exporting a slightly different clip than my previous tests, the results 
on my 4-core Intel Core i5 system are:
* CPU usage went up from 25% to around 85%
* **Export rate estimate increased from 0.58 FPS to over 5.1 FPS**, making the new code **nearly 10x as fast**
* The **old export would've taken about 13 minutes** to complete (again, I aborted before it finished), whereas **the new code can export the same thing in 1m 28s**
* The results are identical; a [video diff](http://dericed.com/2012/display-video-difference-with-ffmpegs-overlay-filter/) of the two exports produces a solid gray field (no differences)
* Unlike my previous attempt, I expect the new code now **will** benefit from additional cores, and run even faster

#### Note
You may notice that in my new code, I don't use `initBoxes`, and the `sigma` property does nothing. In fact, I removed the `initBoxes` function (and all references to it) _completely_.

The reason for this is simple: **In the _current_ (old) Blur effect, the results of `initBoxes` are never used, and the `sigma` property does nothing.**

I don't know if that was intentional or a bug, but in the existing Blur code, `initBoxes()` is a useless function — even though it's called, the results are never used in the code. So as things stood, there was no point in having it in the code. Which is why I removed it.

### DIFF
--- a/src/effects/Blur.cpp
+++ b/src/effects/Blur.cpp
@@ -77,43 +77,9 @@ std::shared_ptr<Frame> Blur::GetFrame(std::shared_ptr<Frame> frame, int64_t fram
 	int w = frame_image->width();
 	int h = frame_image->height();
 
-	// Declare 2-column arrays for each color channel
-	typedef struct {
-		unsigned char *red;
-		unsigned char *green;
-		unsigned char *blue;
-		unsigned char *alpha;
-	} channels;
-	
-	channels arrays_in {
-		new unsigned char[w * h](),
-		new unsigned char[w * h](),
-		new unsigned char[w * h](),
-		new unsigned char[w * h]()
-	};
-	channels arrays_out {
-		new unsigned char[w * h](),
-		new unsigned char[w * h](),
-		new unsigned char[w * h](),
-		new unsigned char[w * h]()
-	};
-
-	// Loop through pixels and split RGBA channels into separate arrays
-	unsigned char *pixels = (unsigned char *) frame_image->bits();
-
-	#pragma omp parallel for
-	for (int pixel = 0; pixel < w * h; ++pixel)
-	{
-		// Get the RGBA values from each pixel
-		arrays_in.red[pixel] = arrays_out.red[pixel] = pixels[pixel * 4];
-		arrays_in.green[pixel] = arrays_out.green[pixel] = pixels[pixel * 4 + 1];
-		arrays_in.blue[pixel] = arrays_out.blue[pixel] = pixels[pixel * 4 + 2];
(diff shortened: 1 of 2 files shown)


src/effects/Blur.cpp
```diff
@@ -74,136 +74,147 @@ std::shared_ptr<Frame> Blur::GetFrame(std::shared_ptr<Frame> frame, int64_t fram
...
-			// Remove boxes
-			delete[] bxs;
-
-			// Copy blur_<chan> back to <chan> for vertical blur or next iteration
-			for (int i = 0; i < (frame_image->width() * frame_image->height()); i++) red[i] = blur_red[i];
-			for (int i = 0; i < (frame_image->width() * frame_image->height()); i++) green[i] = blur_green[i];
-			for (int i = 0; i < (frame_image->width() * frame_image->height()); i++) blue[i] = blur_blue[i];
-			for (int i = 0; i < (frame_image->width() * frame_image->height()); i++) alpha[i] = blur_alpha[i];
-		}
+		#pragma omp parallel num_threads(4) shared (chans_a, chans_b, output)
```
ferdnyc: Hey, one question, if anyone knows — this isn't something I was able to find in the docs so far, though I'll keep looking.

If this is run on a system with _fewer_ than 4 cores, will it still execute with exactly four threads? The way I have the code written, I need that to happen, because the thread ID controls which channel that thread processes. There have to be exactly four, no more no less, or it won't work right.

ferdnyc: Never mind — completely re-worked how OMP parallel processing is applied, and eliminated that bottleneck.

jonoomph:
Awesome job! LGTM

PR #424, opened 2020-02-03
Parallelize and streamline Brightness effect

ferdnyc:
As with #423, this streamlines the Brightness code (less severely, as it was already fairly efficient), and adds a `#pragma omp parallel for` decorator to the main `for()` loop so that it will be multi-threaded.

On my 4-core system, a Brightness/Contrast effect on the same 11-second Big Buck Bunny clip exported in 19 seconds, down from 48.


ferdnyc:
(Unlike #423, I expect users with additional processor cores _will_ see even greater speedups with this code, as processing of pixel arrays in a `for()` loop is a textbook example of code that's **extremely** parallelizable.)

jonoomph:
LGTM!

PR #425, opened 2020-02-03
Streamline and parallelize Wave effect

ferdnyc:
You know the drill by now. Main `for()` loop decorated with `#pragma omp parallel for`, and streamlined for cleaner parallelization.

In this one I also eliminated an unnecessary copy of the frame image data, taking advantage of QImage's implicit data sharing and automatic copying functionality to use the frame image's `constBits()` as the _source_ pixel data, and (an implicitly-copied) `bits()` as the destination.

11-second Big Buck Bunny clip export time on my 4-core system: `28s`, down from `1m24s` (nearly 3x as fast).


jonoomph:
LGTM! Nice work!

PR #426, opened 2020-02-03
Rewrite of Pixelate effect

ferdnyc:
No real improvements this time, to be honest, since there's no real room for parallelization or optimization in the Pixelate code — it's pretty tight already.

However, I did rewrite it to make use of Qt's `QRect` and `QMargins`, in defining the area being pixelated. Which I think makes for marginally cleaner code (and certainly quite a bit shorter!), despite the lack of any significant performance impact.


ferdnyc:
Ironically, if I _hadn't_ rewritten it, the old code could've been parallelized since it looped over scanlines doing a `memcpy()` — but I don't think the minor benefits there would outweigh the advantages of the cleaner rewrite.

jonoomph:
Simple is good! LGTM

PR #427, opened 2020-02-06
Hue: Optimize and parallelize

ferdnyc:
Hue was already pretty efficient, but `#pragma omp parallel for` and some streamlining still produced moderate speedups.

* Export rate increased from 6.60 FPS to 12.90 FPS, very nearly twice as fast
* Time to export a 15-second test clip down from 1m 08s to 35s.


jonoomph:
LGTM!

PR #428, opened 2020-02-06
Saturation: streamline and parallelize

ferdnyc:
Once again, Saturation was fairly efficient already, but adding `#pragma omp parallel for` and eliminating some intermediate variable storage allows it to run roughly twice as fast on my system, and I would expect even better results from a more powerful CPU.


jonoomph:
I love these optimizations, so long as they don't crash on systems. :+1:

ferdnyc:
I love them _even if_ they crash! **#LivingDangerously** (Besides, crashes can be fixed. :angel: )

PR #437, opened 2020-02-10
Limit HW accel support to FFmpeg 3.4+

### LINKED ISSUE #433: AV_HWDEVICE_TYPE_NONE was not declared
I see that recent FFMpegReader.h has:
~~~
#if IS_FFMPEG_3_2
		AVPixelFormat hw_de_av_pix_fmt = AV_PIX_FMT_NONE;
		AVHWDeviceType hw_de_av_device_type = AV_HWDEVICE_TYPE_NONE;
#endif
~~~
However AV_HWDEVICE_TYPE_NONE is not defined in ffmpeg until version 4 (git tag n4.0), despite the suggestion in doc/ HW-ACCEL.md that:
~~~
HW accel is supported from FFmpeg version 3.2 (3.3 for nVidia drivers)
~~~
Having searched various ffmpeg git tags, I think that claim is wrong. Maybe some distros have somehow backlported the hardware acceleraton stuff from 4.x to earlier versions (unlikely?).

As a result, libopenshot-0.2.4 cannot build with (for instance) ffmpeg-3.2.4 with error:
~~~
/var/tmp/SBo/libopenshot-0.2.4/src/../include/FFmpegReader.h:152:41: error: 'AV_HWDEVICE_TYPE_NONE' was not declared in this scope
   AVHWDeviceType hw_de_av_device_type = AV_HWDEVICE_TYPE_NONE;
~~~

### LINKED ISSUE #436: Our hardware-decoding code never supported FFmpeg 3.2
This is a follow-on from #433, as fixing the issue reported there only revealed other issues building libopenshot with FFmpeg 3.2.

It appears that our hardware-accelerated processing code was never actually compatible with FFmpeg 3.2 — even [the initial commit](https://github.com/OpenShot/libopenshot/commit/340803e31eb3385d32a0b8de5e24f227c2b07a1b) that added it to the codebase references `pCodecCtx->hw_device_ctx`, which doesn't exist in FFmpeg 3.2. [FFmpeg 3.2's `AVCodecCtx`](https://ffmpeg.org/doxygen/3.2/structAVCodecContext.html) contains a `hw_frames_ctx`, but no `hw_device_ctx` — that wasn't added [until FFmpeg 3.4](https://ffmpeg.org/doxygen/3.4/structAVCodecContext.html).

It may be possible to add FFmpeg 3.2 support to our hardware-accelerated code, but given that it was never truly supported, as an initial fix it probably makes more sense to just correctly claim that we support hardware acceleration in FFmpeg 3.4+ only.

ferdnyc:
As noted in #436, our hw-accel code was never really compatible with FFmpeg 3.2.

This PR also adds an FFmpeg 3.2 build to our Travis matrix, ensuring that builds against that version are tested in the future.

Fixes #436 
Fixes #433

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ matrix:
 
     - name: "Coverage (Ubuntu 18.04 Bionic)"
       env:
-        - BUILD_VERSION=coverage_ffmpeg3
+        - BUILD_VERSION=coverage_ffmpeg34
         - CMAKE_EXTRA_ARGS="-DENABLE_COVERAGE=1"
         - TEST_TARGET=coverage
       os: linux
@@ -80,9 +80,9 @@ matrix:
           - libavresample4
           - libswresample3
 
-    - name: "FFmpeg 3 GCC (Ubuntu 18.04 Bionic)"
+    - name: "FFmpeg 3.4 GCC (Ubuntu 18.04 Bionic)"
       env:
-        - BUILD_VERSION=ffmpeg3
+        - BUILD_VERSION=ffmpeg34
         - CMAKE_EXTRA_ARGS=""
         - TEST_TARGET=test
       os: linux
@@ -97,9 +97,9 @@ matrix:
           - qt5-default
           - libjsoncpp-dev
 
-    - name: "FFmpeg 3 Clang (Ubuntu 18.04 Bionic)"
+    - name: "FFmpeg 3.4 Clang (Ubuntu 18.04 Bionic)"
       env:
-        - BUILD_VERSION=clang_ffmpeg3
+        - BUILD_VERSION=clang_ffmpeg34
         - CMAKE_EXTRA_ARGS=""
         - TEST_TARGET=test
       os: linux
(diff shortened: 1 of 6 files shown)


ferdnyc:
I'm going to hold off a bit on merging this, in case others (@jonoomph? @eisneinechse?) want to take a different approach, or even fix the FFmpeg 3.2 hw-accel support. But AFAICT that would take quite a bit of work (too much to be worth it), because I've still found no evidence that our existing hw-accel code was ever actually compatible with FFmpeg 3.2 — the use of `hw_device_ctx` in the codec context makes that impossible.

eisneinechse:
From the history of libavcodec I can see:
 2017-02-13 - c1a5fca06f - lavc 57.80.100 - avcodec.h
  Add AVCodecContext.hw_device_ctx.
That was after ffmpeg 3.2 was released. Therefore, setting the minimum required libavcodec to 57.80.100 from 57.64.101 would be right. Maybe using 57. 89.100 (ffmpeg 3.3) would be better or even using 3.4+.
I don't have a version with ffmpeg 3.2 anymore and cannot test the hw acceleration using this version anymore. The programming started with ffmpeg 3.2.

ferdnyc:
@eisneinechse I went with libavcodec 57.107.100, the version released as FFmpeg 3.4, since there's also the question of when `AV_HWDEVICE_FORMAT_NONE` was added to the `AVHWDeviceType` enum (also some time between 3.2 and 3.4), and any other differences there might be between the two versions.

Thanks to the CI system, we've been testing FFmpeg 3.4 all along (it's the version included in Ubuntu Bionic), but since there was no LTS Ubuntu release that included FFmpeg 3.2, it was never part of the testing infrastructure. (In fact, it was never even on our testing radar until yesterday.) So, I have a feeling code slipped in early on that wasn't compatible with FFmpeg 3.2 (but was compatible with 3.4), and since all the CI tests were still passing it understandably went unnoticed.

I figure, we **might** be able to support a slightly earlier version, but since we can't know for sure, we'd have to go through the hassle of testing to figure out exactly what that would be. And... for what? To support hardware-accel de/encoding on a few additional, very-out-of-date systems? Hardly seems worth the effort.

ferdnyc:
I'm going to merge this now, because @cwilling reported that it solves the build issues with FFmpeg 3.2 (as our own newly-upgraded CI matrix also demonstrates, with this merge), and hwaccel support for 3.2/3.3 can always be revisited/restored in a future PR.

PR #439, opened 2020-02-10
Resolve ambiguous abs() call

cwilling:
Signed-off-by: Christoph Willing <chris.willing@linux.com>

This resolves compilatin problem:
~~~
error: call of overloaded 'abs(double)' is ambiguous`
~~~
by  using fabs(double) from cmath rather than trying abs(int) from stdlib.h or  std::abs(long long int) from cstdlib

### DIFF
--- a/src/KeyFrame.cpp
+++ b/src/KeyFrame.cpp
@@ -71,7 +71,7 @@ namespace {
 			}
 			double const x = p0.X * B[0] + p1.X * B[1] + p2.X * B[2] + p3.X * B[3];
 			double const y = p0.Y * B[0] + p1.Y * B[1] + p2.Y * B[2] + p3.Y * B[3];
-			if (abs(target - x) < allowed_error) {
+			if (fabs(target - x) < allowed_error) {
 				return y;
 			}
 			if (x > target) {


ferdnyc:
Hmmm. OOC, is the call still ambiguous if it's written as `std::abs()`?

I have no objection to the use of `std::fabs()` instead, especially if it solves build problems, just trying to get a handle on the issue. (And always on the lookout for issues potentially caused by the IMHO-bad practice of `using namespace std;` in our code.)

What compiler/system gave you that error, BTW? Compiling with the various GCC and Clang versions in our CI matrix didn't flag it, interestingly.

cwilling:
Yes the problem remains when using std::abs(). From the error msg the compiler seems to find only int abs(int) (from stdlib.h), or long int std::abs(long int) and long long int std::abs(long long int) (both from cstdlib). Since we want abs(double), using fabs() resolves the problem.

It could be due to age of compiler gcc-g++-5.5.0 but we're stuck with that on Slackware 14.2 until release of 15.0 which has gcc-g++-9.2.0 (probably still some months away).

Here's the full error message:
~~~
/var/tmp/SBo/libopenshot-0.2.4/src/KeyFrame.cpp: In function 'double {anonymous}::InterpolateBezierCurve(const openshot::Point&, const openshot::Point&, double, double)':
/var/tmp/SBo/libopenshot-0.2.4/src/KeyFrame.cpp:74:27: error: call of overloaded 'abs(double)' is ambiguous
    if (std::abs(target - x) < allowed_error) {
                           ^
In file included from /usr/include/c++/5.5.0/cstdlib:72:0,
                 from /usr/include/c++/5.5.0/ext/string_conversions.h:41,
                 from /usr/include/c++/5.5.0/bits/basic_string.h:5352,
                 from /usr/include/c++/5.5.0/string:52,
                 from /usr/include/c++/5.5.0/bits/locale_classes.h:40,
                 from /usr/include/c++/5.5.0/bits/ios_base.h:41,
                 from /usr/include/c++/5.5.0/ios:42,
                 from /usr/include/c++/5.5.0/ostream:38,
                 from /usr/include/c++/5.5.0/iostream:39,
                 from /var/tmp/SBo/libopenshot-0.2.4/src/../include/KeyFrame.h:34,
                 from /var/tmp/SBo/libopenshot-0.2.4/src/KeyFrame.cpp:31:
/usr/include/stdlib.h:774:12: note: candidate: int abs(int)
 extern int abs (int __x) __THROW __attribute__ ((__const__)) __wur;
            ^
In file included from /usr/include/c++/5.5.0/ext/string_conversions.h:41:0,
                 from /usr/include/c++/5.5.0/bits/basic_string.h:5352,
                 from /usr/include/c++/5.5.0/string:52,
                 from /usr/include/c++/5.5.0/bits/locale_classes.h:40,
                 from /usr/include/c++/5.5.0/bits/ios_base.h:41,
                 from /usr/include/c++/5.5.0/ios:42,
                 from /usr/include/c++/5.5.0/ostream:38,
                 from /usr/include/c++/5.5.0/iostream:39,
                 from /var/tmp/SBo/libopenshot-0.2.4/src/../include/KeyFrame.h:34,
                 from /var/tmp/SBo/libopenshot-0.2.4/src/KeyFrame.cpp:31:
/usr/include/c++/5.5.0/cstdlib:166:3: note: candidate: long int std::abs(long int)
   abs(long __i) { return __builtin_labs(__i); }
   ^
/usr/include/c++/5.5.0/cstdlib:174:3: note: candidate: long long int std::abs(long long int)
   abs(long long __x) { return __builtin_llabs (__x); }
   ^
src/CMakeFiles/openshot.dir/build.make:638: recipe for target 'src/CMakeFiles/openshot.dir/KeyFrame.cpp.o' failed
make[2]: *** [src/CMakeFiles/openshot.dir/KeyFrame.cpp.o] Error 1
CMakeFiles/Makefile2:161: recipe for target 'src/CMakeFiles/openshot.dir/all' failed
```

ferdnyc:
Huh... so odd that it still considers `abs` from `stdlib.h` a candidate. But since there are two `std::abs()` candidates, it's kind of immaterial. And, indeed, `fabs()` is used at multiple other points in the code. I'm good with it! Merging, thanks.

PR #442, opened 2020-02-13
Add -no-integrated-cpp for G++ < 9

ferdnyc:
This change (from #254) got accidentally dropped soon after merge, when 4e08ab32c1386366305212875925ab81ed1e257a happened. (Which explains why Travis is showing deprecation warnings during the build.)


ferdnyc:
Merging; we thought it was already there!

PR #443, opened 2020-02-13
ColorShift effect: Don't init with random values

ferdnyc:
For no discernible reason, the Color Shift effect initializes itself with the following properties:
|Property|Value|
|-----------|-------|
|Alpha X Shift|0.00|
|Alpha Y Shift|0.00|
|<b>Red X Shift</b>|<b>-0.05</b>|
|Red Y Shift|0.00|
|<b>Green X Shift</b>|<b>0.05</b>|
|Green Y Shift|0.00|
|Blue X Shift|0.00|
|Blue Y Shift|0.00|

The random "defaults" are arbitrary and unnecessary, and just complicate management of the effect parameters. This PR loses them, and inits all of the shifts to `0.00`.

### DIFF
--- a/src/effects/ColorShift.cpp
+++ b/src/effects/ColorShift.cpp
@@ -33,7 +33,7 @@
 using namespace openshot;
 
 /// Blank constructor, useful when using Json to load the effect properties
-ColorShift::ColorShift() : red_x(-0.05), red_y(0.0), green_x(0.05), green_y(0.0), blue_x(0.0), blue_y(0.0), alpha_x(0.0), alpha_y(0.0) {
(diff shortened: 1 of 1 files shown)


ferdnyc:
(Note: This change won't affect existing projects, since all property values are exported directly into the project file — only newly-added instances of ColorShift will use different initial property values.)

ferdnyc:
> I think the intention on the initial values was to "show" a difference when the effect was applied. 

Yeah, and I get that — with some effects, it makes a lot of sense, because they have some sort of linear adjustment that can be defaulted to some "does its thing" point along that linear scale. But these defaults just felt so _arbitrary_ and weird, even relative to the other effects' arbitrary/weird defaults. :laughing:

PR #446, opened 2020-02-22
WIP: A few more Frame tests

musteresel:
I've expanded the test suite for the `Frame` class a bit, and also corrected some (seemingly?) outdated comments in `Frame.h`.  The tests still do not cover most of what a `Frame` can do. A lot of the behavior which I'd like to test more is actually quite ... surprising, though.  I'll open an issue regarding this.

Depending on the conclusion of the issue I'll further extend these tests to capture either the existing behavior as good as possible or the IMO more consistent behavior.  Therefore I mark this as "Work In Process" for now :)

### DIFF
--- a/include/Frame.h
+++ b/include/Frame.h
@@ -73,17 +73,17 @@ namespace openshot
 	 * There are many ways to create an instance of an openshot::Frame:
 	 * @code
 	 *
-	 * // Most basic: a blank frame (300x200 blank image, 48kHz audio silence)
+	 * // Most basic: a blank frame (1x1 blank image, 44.2kHz audio silence)
 	 * Frame();
 	 *
-	 * // Image only settings (48kHz audio silence)
+	 * // Image only settings (44.2kHz audio silence)
 	 * Frame(1, // Frame number
 	 *       720, // Width of image
 	 *       480, // Height of image
 	 *       "#000000" // HTML color code of background color
 	 *       );
 	 *
-	 * // Audio only (300x200 blank image)
+	 * // Audio only (1x1 blank image)
 	 * Frame(number, // Frame number
 	 *       44100, // Sample rate of audio stream
 	 *       2 // Number of audio channels
@@ -131,13 +131,13 @@ namespace openshot
 		bool has_image_data; ///< This frame has been loaded with pixel data
 
 
-		/// Constructor - blank frame (300x200 blank image, 48kHz audio silence)
+		/// Constructor - blank frame (1x1 blank image, 44.2kHz audio silence)
 		Frame();
 
-		/// Constructor - image only (48kHz audio silence)
+		/// Constructor - image only (44.2kHz audio silence)
 		Frame(int64_t number, int width, int height, std::string color);
 
-		/// Constructor - audio only (300x200 blank image)
+		/// Constructor - audio only (1x1 blank image)
 		Frame(int64_t number, int samples, int channels);
 
 		/// Constructor - image & audio
--- a/tests/Frame_Tests.cpp
+++ b/tests/Frame_Tests.cpp
@@ -57,6 +57,21 @@ TEST(Default_Constructor)
 	CHECK_EQUAL(false, f1.has_audio_data);
 }
 
+TEST(ImageOnly_Constructor)
+{
+	Frame f(42, 100, 111, "#FAFBFC");
+
+	CHECK_EQUAL(100, f.GetWidth());
+	CHECK_EQUAL(111, f.GetHeight());
+	CHECK_EQUAL(44100, f.SampleRate());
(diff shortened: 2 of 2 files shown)


musteresel:
Regarding the failing CI: This is due to me using `REQUIRE` from unittest++ (since version 1.6.0) but unittest++ on these CI instances is at version 1.4 (from *2008* ..). Thoughts on that?

SuslikV:
What is the issue with the existing tests? There is no clear explanation given.

musteresel:
The existing tests cover almost none of `Frame`s code; which is why I'd like to extend the tests.

include/Frame.h
```diff
@@ -73,17 +73,17 @@ namespace openshot
 	 * There are many ways to create an instance of an openshot::Frame:
 	 * @code
 	 *
-	 * // Most basic: a blank frame (300x200 blank image, 48kHz audio silence)
+	 * // Most basic: a blank frame (1x1 blank image, 44.2kHz audio silence)
```
ferdnyc:
```suggestion
	 * // Most basic: a blank frame (1x1 blank image, 44.1kHz audio silence)
```

All of those `44.2`s should be `44.1`.

ferdnyc:
> The existing tests cover almost none of Frames code; which is why I'd like to extend the tests.

Our tests desperately need more coverage, so I'm very much in favor of that under any circumstances.

As far as the Ubuntu 16.04 failures, unfortunately that release is still supported for another year+ (not 2 months, as I'd previously said -- I was off by a year), so our tests need to be compatible. If there's a way to use `REQUIRE` but stay compatible with 16.04, then that's fine. Maybe something along these lines?
```c++
#ifndef REQUIRE
  #define REQUIRE {}
#endif
```

Unfortunately we won't get an updated coverage report until the CI run passes all jobs.

ferdnyc:
> As far as the Ubuntu 16.04 failures, unfortunately that release is still supported for another year+ (not 2 months, as I'd previously said -- I was off by a year), so our tests need to be compatible. If there's a way to use `REQUIRE` but stay compatible with 16.04, then that's fine. Maybe something along these lines?
> 
> ```c++
> #ifndef REQUIRE
>   #define REQUIRE
> #endif
> ```

I just pushed a commit to do exactly that, we'll see if Travis likes it...

ferdnyc:
That worked, just waiting on the rest of the builds to finish so we can get updated coverage stats.

musteresel:
Fixed the 44.2 to 44.1 :)

ferdnyc:
Looking at the [full coverage map for `Frame.cpp`](https://codecov.io/gh/OpenShot/libopenshot/pull/446/src/src/Frame.cpp) as of this PR, there's a lot of stuff in there that isn't really a priority for unit-testing. (Things like `Frame::Play()` and `Frame::Display()` that really aren't _API_ calls at all — the ones where it's open for debate whether they should even be methods of the Frame class at all.) So, 100% coverage wouldn't, IMHO, really be the goal.

But there is still plenty of other stuff we _could_ be testing, like (just walking down the source from top to bottom...):
- [ ] `Frame::operator=()` assignment
- [ ] `Frame::GetWaveform()`
- [ ] `Frame::ClearWaveform()`
- [ ] `Frame::GetAudioSample()` 
    - [ ] with `channel == 0`
    - [ ] and with `channel > 0`
- [ ] `Frame::GetPlanarAudioSamples()`
    - [ ] with and
    - [ ] without resampling
- [ ] `Frame::GetAudioSampleBuffer()`
- [ ] calling `Frame::GetPixels()` on a frame with no `image`
- [ ] calling `Frame::CheckPixel()` with invalid `row`/`col`
- [ ] `Frame::GetSamplesPerFrame()`
- [ ] calling `Frame::Save()` with a modified aspect ratio
- [ ] `Frame::Thumbnail()`
    - [ ] with and
    - [ ] without a modified aspect ratio
    - [ ] with and
    - [ ] without `ignore_aspect`
    - [ ] with and
    - [ ] without an `overlay_path`
    - [ ] with and
    - [ ] without a `mask_path`
- [ ] `Frame::constrain()`
- [ ] calling `Frame::AddImage()` on an interlaced frame, with non-blank source and destination images — code that I contend is currently very broken (#418)
- [ ] `Frame::ApplyGainRamp()`
- [ ] calling `Frame::GetMagickImage()` on a frame with no `image`

(...I'm just throwing that list out there as a sort of general status-check. It's certainly not any**one's** to-do list. Or, maybe it's _every**one's**_ to-do list. Meaning, the list is there — and please feel free to amend or correct it — for anyone who'd like to contribute tests for any of those branches. I'm hoping it'll help me guilt myself into writing at least a few of them, too.)

PR #448, opened 2020-02-24
Change frame rate detection

SuslikV:
FFmpeg re-encoding algorithm don't uses `avg_frame_rate`.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -706,9 +706,16 @@ void FFmpegReader::UpdateVideoInfo() {
 	info.vcodec = pCodecCtx->codec->name;
 	info.video_bit_rate = (pFormatCtx->bit_rate / 8);
 
-	// set frames per second (fps)
-	info.fps.num = pStream->avg_frame_rate.num;
-	info.fps.den = pStream->avg_frame_rate.den;
+	// Frame rate from the container and codec
+	AVRational framerate = av_guess_frame_rate(pFormatCtx, pStream, NULL);
+	info.fps.num = framerate.num;
+	info.fps.den = framerate.den;
+
+	ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::UpdateVideoInfo", "info.fps.num", info.fps.num, "info.fps.den", info.fps.den);
+
+	// TODO: remove excessive debug info in the next releases
+	// The debug info below is just for comparison and troubleshooting on users side during the transition period
+	ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::UpdateVideoInfo (pStream->avg_frame_rate)", "num", pStream->avg_frame_rate.num, "den", pStream->avg_frame_rate.den);
 
 	if (pStream->sample_aspect_ratio.num != 0) {
 		info.pixel_ratio.num = pStream->sample_aspect_ratio.num;
--- a/tests/FFmpegReader_Tests.cpp
+++ b/tests/FFmpegReader_Tests.cpp
@@ -36,13 +36,16 @@
 using namespace std;
 using namespace openshot;
 
-TEST(FFmpegReader_Invalid_Path)
+SUITE(FFmpegReader)
+{
+
+TEST(Invalid_Path)
 {
 	// Check invalid path
 	CHECK_THROW(FFmpegReader(""), InvalidFile);
 }
 
-TEST(FFmpegReader_GetFrame_Before_Opening)
+TEST(GetFrame_Before_Opening)
 {
 	// Create a reader
 	stringstream path;
@@ -53,7 +56,7 @@ TEST(FFmpegReader_GetFrame_Before_Opening)
 	CHECK_THROW(r.GetFrame(1), ReaderClosed);
 }
 
-TEST(FFmpegReader_Check_Audio_File)
+TEST(Check_Audio_File)
 {
 	// Create a reader
 	stringstream path;
@@ -83,7 +86,7 @@ TEST(FFmpegReader_Check_Audio_File)
 	r.Close();
 }
 
-TEST(FFmpegReader_Check_Video_File)
+TEST(Check_Video_File)
 {
 	// Create a reader
 	stringstream path;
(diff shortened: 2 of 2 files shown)


src/FFmpegReader.cpp
```diff
@@ -706,9 +706,12 @@ void FFmpegReader::UpdateVideoInfo() {
 	info.vcodec = pCodecCtx->codec->name;
 	info.video_bit_rate = (pFormatCtx->bit_rate / 8);
 
-	// set frames per second (fps)
-	info.fps.num = pStream->avg_frame_rate.num;
-	info.fps.den = pStream->avg_frame_rate.den;
+	// Frame rate from the container and codec
+	AVRational framerate = av_guess_frame_rate(pFormatCtx, pStream , NULL);
```
ferdnyc:
```suggestion
	AVRational framerate = av_guess_frame_rate(pFormatCtx, pStream, NULL);
```
Remove the spurious space character, please.

src/FFmpegReader.cpp
```diff
@@ -706,9 +706,12 @@ void FFmpegReader::UpdateVideoInfo() {
...
+	info.fps.num = framerate.num;
+	info.fps.den = framerate.den;
+
+	ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::UpdateVideoInfo", "fps_num", info.fps.num, "fps_den", info.fps.den, "avg_fps_num", pStream->avg_frame_rate.num, "avg_fps_den", pStream->avg_frame_rate.den);
```
ferdnyc: For testing/comparison purposes during development I totally see the value, but in terms of the library's logging going forward, it doesn't really make sense to log values we're _not_ using (`pStream->avg_frame_rate_{num,den}`). If we're switching to `av_guess_frame_rate()` because it produces better values (and if you say that it does, good enough for me), then IMHO we should just log **those** values. Also, tweaked the labels of logged values to reflect how they're stored in the code.
```suggestion
	ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::UpdateVideoInfo (set info.fps)", "info.fps.num", info.fps.num, "info.fps.den", info.fps.den);
```

ferdnyc: (If we want to see comparisons in the short-term, to "monitor" the accuracy of `av_guess_frame_rate()`, then I'd say let's log the `pstream->avg_frame_rate` values in a **separate** log message. Later, we can just delete that line.)
```cpp
ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::UpdateVideoInfo (pStream->avg_frame_rate)", "num", pStream->avg_frame_rate.num, "den", pStream->avg_frame_rate.den);
```

SuslikV: Of course, message can be omitted completely. This strings is called rarely (on file open only, thus when seeking will be fixed - this will be called just once per file) and I placed debugging info just for comparison. The clip FPS has huge impact on everything else in OpenShot (start/duration etc.) that calculated in seconds, so this change requires a bit more attention. **I think**, this solution is better than the one the OpenShot has right now.

I'm choosing the message split.

SuslikV:
I hope, now it looks a bit better. Updated on top of the current _develop_ branch.

src/FFmpegReader.cpp
```diff
@@ -706,9 +706,16 @@ void FFmpegReader::UpdateVideoInfo() {
 	info.vcodec = pCodecCtx->codec->name;
 	info.video_bit_rate = (pFormatCtx->bit_rate / 8);
 
-	// set frames per second (fps)
-	info.fps.num = pStream->avg_frame_rate.num;
-	info.fps.den = pStream->avg_frame_rate.den;
+	// Frame rate from the container and codec
+	AVRational framerate = av_guess_frame_rate(pFormatCtx, pStream, NULL);
+	info.fps.num = framerate.num;
+	info.fps.den = framerate.den;
```
ferdnyc: This reminds me that I tried to add operators and conversion methods into `openshot::Fraction` so it would automatically convert from/to `AVRational` as needed, once. But I think the fact that `AVRational` is implemented in C instead of C++ probably doomed that idea right from the start.

ferdnyc:
Are there unit tests that do any checks on frame-rate detection at all? I don't think so... in fact the only frame rate check I see is this one, which doesn't involve FFmpegReader at all.
https://github.com/OpenShot/libopenshot/blob/c38c55fbfa128c16b0df497ac4e151027a22d94b/tests/Timeline_Tests.cpp#L82-L90

Oh, no, there is one in FFmpegWriter that checks the rate of a newly-written file, after reading it back in.

ferdnyc:
I added an additional frame rate check in the FFmpegReader_Tests suite, using the standard sintel video source. (Which is uninteresting at 24fps, but like I said it's a sanity check.) With that in place, I'll merge this once Travis is done with it. I was hoping to get @eisneinechse 's take, but it's been a month and I don't want to hold this up forever.

SuslikV:
All these unit tests is based on _libopenshot_ own implementations, so they exist here just for fun. This is my own thoughts about these kind of stuff in _libopenshot_. With the broken instrument you'll be unable to make right measurements (it is more like freezing bugs at some level for the current build ^_^).

ferdnyc:
> All these unit tests is based on _libopenshot_ own implementations, so they exist here just for fun.

Well, no — it's _**exactly**_ what unit tests are. They check that the implementation matches expectations, in isolation. And, just as importantly (and the reason they're run on every build), they catch when changes to the code **alter** the implementation in unexpected ways.

The [Android developer intro](https://developer.android.com/training/testing/unit-testing) sums it up pretty well:
> Unit tests are the fundamental tests in your app testing strategy. By creating and running unit tests against your code, you can easily verify that the logic of individual units is correct. Running unit tests after every build helps you to quickly catch and fix software regressions introduced by code changes to your app.
> 
> A unit test generally exercises the functionality of the smallest possible unit of code (which could be a method, class, or component) in a repeatable way. You should build unit tests when you need to verify the logic of specific code in your app. For example, if you are unit testing a class, your test might check that the class is in the right state. Typically, the unit of code is tested in isolation; your test affects and monitors changes to that unit only. 

Unit tests don't test whether your structure is put together correctly, only that the logical pieces it's assembled from are really the shape we think they are.

Testing how it all fits together is where integration testing comes into play. But integration testing is pointless when you're assembling untested pieces.

PR #449, opened 2020-02-25
CMake: Limit scope of AUTOMOC

ferdnyc:
- Only set `AUTOMOC` property on library target
- Only use the Qt headers in `target_sources()`
- Make Qt headers `PRIVATE` sources, to avoid migration to other targets

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,8 +27,6 @@
 # Collect and display summary of options/dependencies
 include(FeatureSummary)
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
This change prevents a `<target>_autogen` directory from being created for **every** target in the build (a consequence of global `CMAKE_AUTOMOC`), and prevents the Qt headers from being MOC'd separately for each target (a consequence of the `PUBLIC` headers).

PR #450, opened 2020-02-25
Update CodeCoverage.cmake module

ferdnyc:
There have been some enhancements made to `CodeCoverage.cmake` in [bilke/cmake-modules](https://github.com/bilke/cmake-modules), incorporating them into our local copy.


ferdnyc:
Merging as it only affects the coverage build (and even that, not meaningfully)

PR #455, opened 2020-03-05
FFMpegWriter: Ease technical debt by dropping support for FFmpeg versions below 2.4

### LINKED ISSUE #490: AV_GET_CODEC_TYPE() points to different enum types. Drop support for FFmpeg below 0.8 and below
The AV_GET_CODEC_TYPE() returns `codec_type` field that itself has different types depending on FFmpeg version

Compare types of the `codec_type` field:
https://ffmpeg.org/doxygen/0.5/structAVCodecContext.html
vs
https://ffmpeg.org/doxygen/1.0/structAVCodecContext.html

This also has workaround in the code, that doesn't uses AV_GET_CODEC_TYPE() call at all (stream not in use here):
https://github.com/OpenShot/libopenshot/blob/166f69389b6d8a601b51e8d559c418f56a144bba/src/FFmpegWriter.cpp#L1074-L1078

https://github.com/OpenShot/libopenshot/blob/166f69389b6d8a601b51e8d559c418f56a144bba/src/FFmpegWriter.cpp#L1157-L1161

and next part of the code missing this version check:
here the `AVMEDIA_TYPE_AUDIO` can be `CODEC_TYPE_AUDIO` and vise versa.
https://github.com/OpenShot/libopenshot/blob/166f69389b6d8a601b51e8d559c418f56a144bba/src/FFmpegWriter.cpp#L798-L804

If FFmpeg below 0.8 isn't supported then some checks can be omitted.
Similar to already  proposed changes: https://github.com/OpenShot/libopenshot/pull/455

ferdnyc:
A bunch of code in FFmpegWriter was checking for library versions that corresponded to an FFmpeg version from the 2.x series (or even **1.x**, in one case) — versions that are far older than we need to support.

This PR drops that code entirely, making FFmpeg version 2.4 our minimum compatible version. Which seems more than adequate, considering 2.4.1 was released 2014-09-22.

In addition:
1. Comments are added to more version tests, translating things like
    ```c++
    #if (LIBAVFORMAT_VERSION_MAJOR >= 58)
    ```
    into
    ```c++
    #if (LIBAVFORMAT_VERSION_MAJOR >= 58)
        // FFmpeg 4.0+
    ```
1. (This one is arguably unrelated and I should probably have made it a separate PR)
    Uses of `fprintf(stderr...)` and `std::cerr <<` are replaced either with ZeroMQ logging, or stream IO to the `std::clog` output stream, which (unlike either of the two output methods previously used) is [documented as being thread-safe](https://en.cppreference.com/w/cpp/io/clog).

1. Some of the logging calls are cleaned up, `av_err2str()` is added to more of them to produce human-readable FFmpeg errors, and the spelling of "Encodiing device" is fixed because it's been driving me crazy forever.

1. A bunch of calls with long lines are reformatted to be more readable.
    In particular, sections like this:
    ```c++
    avcodec_fill_audio_frame(frame_final, audio_codec->channels, audio_codec->sample_fmt, (uint8_t *) final_samples,
    						 audio_encoder_buffer_size, 0);
    ```
    (Which might even line up past the opening parens, depending on your tab width)
    instead become either this:
    ```c++
    avcodec_fill_audio_frame(frame_final, audio_codec->channels, audio_codec->sample_fmt,
    	(uint8_t *) final_samples, audio_encoder_buffer_size, 0);
    ```
    (that's a _single_ tab indent on the continuing line)
    or in the most extreme cases, they become this:
    ```c++
    avcodec_fill_audio_frame(
    	frame_final, audio_codec->channels, audio_codec->sample_fmt,
    	(uint8_t *) final_samples, audio_encoder_buffer_size, 0
    );
    ```

Fixes #490


ferdnyc:
(Now, fingers crossed that the default FFmpeg in Ubuntu Xenial is >= 2.4. I think it is, I think it's like 2.8 or something.)

jonoomph:
Thanks for trimming the tech debt on this one! Looks good to me!

jonoomph:
I'll be honest, this PR was large... but we'll see if anything breaks!

PR #457, opened 2020-03-09
CMake: Use GNUInstallDirs for install paths

ferdnyc:
This is something we're already doing for libopenshot-audio, but it never migrated here.

Automatically does the right thing with `lib` vs. `lib64` vs. `lib/x86_64-linux-gnu`, etc.


src/CMakeLists.txt
```diff
@@ -409,42 +412,40 @@ add_executable(openshot-player Qt/demo/main.cpp)
...
 # Install primary library
-INSTALL(TARGETS openshot
-		ARCHIVE DESTINATION ${LIB_INSTALL_DIR}
-		LIBRARY DESTINATION ${LIB_INSTALL_DIR}
-		RUNTIME DESTINATION ${LIB_INSTALL_DIR}
-		COMPONENT library )
+install(TARGETS openshot
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
ferdnyc: I accidentally included a change here to install the library DLLs into `bin'`, instead of into `lib/`, on Windows. Which is something I do want to do (as it's more correct when building Windows code), but we're not ready for that yet.

```suggestion
  RUNTIME DESTINATION ${CMAKE_INSTALL_LIBDIR})
```

ferdnyc:
With my accidental directory target change reverted, this is very safe and a desirable improvement to the build configs. Merging.

PR #458, opened 2020-03-09
bindings: Relative paths, overridable

ferdnyc:
This is an attempt to fix the install paths for the bindings files, in particular the Ruby bindings which have been installing to an absolute path (which causes problems when building as non-root). In addition, the Gitlab builder scripts were _moving_ Python artifacts from their installed locations to `${CMAKE_INSTALL_PREFIX}/python`, which is just silly.

Now, `PYTHON_MODULE_PATH` or `RUBY_MODULE_PATH` can be defined on the CMake command line, and they'll override the detection process. (They're interpreted relative to the install prefix, so just `-DPYTHON_MODULE_PATH=python` recreates the previous scripted output.

In addition, the module paths on Windows and MacOS default to simply "python" or "ruby" instead of bothering with detection, since we don't install on the system there anyway.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -16,13 +16,11 @@ linux-builder:
     - unzip artifacts.zip
     - export LIBOPENSHOT_AUDIO_DIR=$CI_PROJECT_DIR/build/install-x64
     - mkdir -p build; cd build;
-    - mkdir -p install-x64/python;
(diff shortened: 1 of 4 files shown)


ferdnyc:
If the Gitlab builders like this, I'm probably going to merge it along with #469, so the bindings can stop being a pain point in our build/install process.

ferdnyc:
Everything wound up where it was meant to, on all four platforms. Merging.

PR #459, opened 2020-03-09
Adding ability for a Clip to auto-detect and instantiate a Timeline Reader

jonoomph:
Adding ability for a Clip to auto-detect and instantiate a Timeline Reader from the *.osp file type. Added new Timeline constructor, to auto load UTF-8 JSON file, and regex convert all paths to absolute. Fixed a dead lock issue when a Timeline loads another Timeline.

This is the first big step towards allowing multiple *.osp files to exist inside an OpenShot Project. This opens up an infinite number of cool possibilities, such as splitting projects into scenes, and then combining scenes at the end. Or importing templates (i.e. *.osp projects) with the complexity hidden from the user (since it looks like a normal clip). I have big plans for this soon!!!

However, performance seems suspect currently, and we'll need to dig deeper, and determine how to most efficiently set thread limits and cache limits for nested Timeline readers.

### DIFF
--- a/src/Timeline.cpp
+++ b/src/Timeline.cpp
@@ -117,7 +117,7 @@ Timeline::Timeline(std::string projectPath, bool convert_absolute_paths) :
 	QDir asset_folder(filePath.dir().filePath(asset_name));
 	if (!asset_folder.exists()) {
 		// Create directory if needed
-		asset_folder.makeAbsolute();
+		asset_folder.mkpath(".");
 	}
 
 	// Load UTF-8 project file into QString
@@ -125,20 +125,46 @@ Timeline::Timeline(std::string projectPath, bool convert_absolute_paths) :
 	projectFile.open(QFile::ReadOnly);
 	QString projectContents = QString::fromUtf8(projectFile.readAll());
 
-	// Convert all relative paths into absolute paths (does not support relative paths with ../)
-	// In otherwords, assets and files must be located in a child/sub-folder (and not from outside this folder)
+	// Convert all relative paths into absolute paths (if requested)
 	if (convert_absolute_paths) {
-		// Convert all paths into absolute (if requested)
-		QRegularExpression pathRegex(QStringLiteral("\"(image|path)\":.*?\"(?:\\./)?(?!@assets|@transitions+)(.*?)\""));
-		projectContents.replace(pathRegex, "\"\\1\": \"" + filePath.absoluteDir().absoluteFilePath("\\2") + "\"");
 
-		// Convert all transitions paths into absolute (if requested)
-		QRegularExpression transRegex(QStringLiteral("\"(image|path)\":.*?\"@transitions/*(.*?)\""));
-		projectContents.replace(transRegex, "\"\\1\": \"" + openshotTransPath.absoluteFilePath("\\2") + "\"");
+		// Find all "image" or "path" references in JSON (using regex). Must loop through match results
+		// due to our path matching needs, which are not possible with the QString::replace() function.
+		QRegularExpression allPathsRegex(QStringLiteral("\"(image|path)\":.*?\"(.*?)\""));
+		std::vector<QRegularExpressionMatch> matchedPositions;
+		QRegularExpressionMatchIterator i = allPathsRegex.globalMatch(projectContents);
+		while (i.hasNext()) {
+			QRegularExpressionMatch match = i.next();
+			if (match.hasMatch()) {
+				// Push all match objects into a vector (so we can reverse them later)
+				matchedPositions.push_back(match);
+			}
+		}
 
-		// Convert all assets paths into absolute
-		QRegularExpression assetRegex(QStringLiteral("\"(image|path)\":.*?\"@assets/*(.*?)\""));
-		projectContents.replace(assetRegex, "\"\\1\": \"" + asset_folder.absoluteFilePath("\\2") + "\"");
+		// Reverse the matches (bottom of file to top, so our replacements don't break our match positions)
+		std::vector<QRegularExpressionMatch>::reverse_iterator itr;
+		for (itr = matchedPositions.rbegin(); itr != matchedPositions.rend(); itr++) {
+			QRegularExpressionMatch match = *itr;
+			QString relativeKey = match.captured(1); // image or path
+			QString relativePath = match.captured(2); // relative file path
+			QString absolutePath = "";
+
+			// Find absolute path of all path, image (including special replacements of @assets and @transitions)
+			if (relativePath.startsWith("@assets")) {
+				absolutePath = QFileInfo(asset_folder.absoluteFilePath(relativePath.replace("@assets", "."))).canonicalFilePath();
+			} else if (relativePath.startsWith("@transitions")) {
+				absolutePath = QFileInfo(openshotTransPath.absoluteFilePath(relativePath.replace("@transitions", "."))).canonicalFilePath();
+			} else {
+				absolutePath = QFileInfo(filePath.absoluteDir().absoluteFilePath(relativePath)).canonicalFilePath();
+			}
+
+			// Replace path in JSON content, if an absolute path was successfully found
+			if (!absolutePath.isEmpty()) {
+				projectContents.replace(match.capturedStart(0), match.capturedLength(0), "\"" + relativeKey + "\": \"" + absolutePath + "\"");
+			}
+		}
+		// Clear matches
+		matchedPositions.clear();
 	}
 
 	// Set JSON of project
--- a/src/Timeline.cpp
+++ b/src/Timeline.cpp
@@ -34,7 +34,7 @@ using namespace openshot;
 
 // Default Constructor for the timeline (which sets the canvas width and height)
 Timeline::Timeline(int width, int height, Fraction fps, int sample_rate, int channels, ChannelLayout channel_layout) :
-		is_open(false), auto_map_clips(true), managed_cache(true)
+		is_open(false), auto_map_clips(true), managed_cache(true), path("")
 {
 	// Create CrashHandler and Attach (incase of errors)
 	CrashHandler::Instance();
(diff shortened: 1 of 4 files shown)


src/Timeline.cpp
```diff
@@ -73,6 +75,107 @@ Timeline::Timeline(int width, int height, Fraction fps, int sample_rate, int cha
...
+	QFile projectFile(QString::fromStdString(path));
+	projectFile.open(QFile::ReadOnly);
+	QString projectContents = QString::fromUtf8(projectFile.readAll());
+
+	// Convert all relative paths into absolute paths (does not support relative paths with ../)
+	// In otherwords, assets and files must be located in a child/sub-folder (and not from outside this folder)
+	if (convert_absolute_paths) {
+		// Convert all paths into absolute (if requested)
+		QRegularExpression pathRegex(QStringLiteral("\"(image|path)\":.*?\"(?:\\./)?(?!@assets|@transitions+)(.*?)\""));
+		projectContents.replace(pathRegex, "\"\\1\": \"" + filePath.absoluteDir().absoluteFilePath("\\2") + "\"");
```
jonoomph: @ferdnyc This works great for converting our relative paths into absolute ones: `Project Path` + relative `FilePath`, but canonicalFilePath() doesn't work on my system, and always returns an empty string. So, I'm using absoluteFilePath() instead for the moment, which has an interesting consequence.

**ProjectPath** = `/home/jonathan/project1/`
**FilePath** = `@assets/titles/title1.svg`
**Calculated Path** = `/home/jonathan/project1/project1_assets/titles/title1.svg` :heavy_check_mark: 

However, if your relative paths are for a file outside of the project folder, such as:
**FilePath** = `../../Pictures/titles/title1.svg`
**Calculated Path** = `/home/jonathan/project1/../../Pictures/titles/title1.svg` :stop_sign: 

Thus, this PR only supports *.osp files with relative content, until we can find a solution for canonicalFilePath() that doesn't replace all my file paths with an empty string. :wink:. 

However, a big design goal of this, is to support more advanced templates, downloadable templates, advanced titles, and more, that depend on having a more complex layout (i.e. *.osp file behind the scenes).

SuslikV:
```
std::experimental::filesystem::path p = L"/home/jonathan/project1/../../Pictures/titles/title1.svg"; // wide string
std::cout << "canonical path = " << std::experimental::filesystem::canonical(p) << '\n';
```
How about this? If file exist, of course. (requires `-lstdc++fs` at compile)

src/Timeline.cpp
```diff
@@ -73,6 +75,107 @@ Timeline::Timeline(int width, int height, Fraction fps, int sample_rate, int cha
...
+	// Load UTF-8 project file into QString
+	QFile projectFile(QString::fromStdString(path));
+	projectFile.open(QFile::ReadOnly);
+	QString projectContents = QString::fromUtf8(projectFile.readAll());
+
+	// Convert all relative paths into absolute paths (does not support relative paths with ../)
+	// In otherwords, assets and files must be located in a child/sub-folder (and not from outside this folder)
+	if (convert_absolute_paths) {
+		// Convert all paths into absolute (if requested)
+		QRegularExpression pathRegex(QStringLiteral("\"(image|path)\":.*?\"(?:\\./)?(?!@assets|@transitions+)(.*?)\""));
```
ferdnyc: OK, first off, this is a horror show. :laughing: C++11 supports raw string literals, same as Python (but with a slightly different format)
```suggestion
		QRegularExpression pathRegex(QStringLiteral(
			R"__(
				"(image|path)":.*?"(?:\./)?(?!@assets|@transitions+)(.*?)
			")__", QRegularExpression::ExtendedPatternSyntaxOption
		));
```
...Much more readable. And if I plug that into [Debuggex (live session link)](https://www.debuggex.com/r/uFO61iT7iS165S-E
), a couple of parts are iffy — `@transitions+` is the major red flag, since it would match `@transitions`, `@transitionss`, @transitionssssssss`... unlikely to be the intent.

I think this is more like what you want:
```c++
R"__(
	"(image|path)":\s*?"(?:\./)?(?!@(?:assets|transitions)/)(.*)"
)__"
```
...However, that's still assuming QRegularExpression supports all of the advanced features in use, including negative lookahead — RE implementations can be surprisingly lacking. Still trying to confirm that in the documentation.

src/Timeline.cpp
ferdnyc: @SuslikV The STL finally has a [`std::fileystem` implementation](https://en.cppreference.com/w/cpp/filesystem) that does all that kind of stuff, and would theoretically be helpful here — but it requires C++17. (As well as libstdc++fs.) Even if we wanted to migrate the code to the newer standard, there's no way it's going to be supported by gcc 4.8.4 (Linux & Mac builders). Only the Windows builders are running gcc 9.1.0, currently. Still, Qt's implementation _should_ serve, and it's been around a lot longer.

ferdnyc: @jonoomph 
> but canonicalFilePath() doesn't work on my system, and always returns an empty string.

That makes sense, actually, as `canonicalPath()` is documented as returning the empty string if the path does not exist, and I can _guaran-damn-tee_ that there is no file named `\2` in that directory. :smile:

Because the path methods are going to be evaluated first, `canonicalPath()` (or `absoluteFilePath()` will be getting the unexpanded `"\\2"` string as their argument — fortunately, `absoluteFilePath()` doesn't check whether the file passed to it actually exists, so it'll just pass it through unmolested. 

How you'd do this with `canonicalPath()`, I am not entirely sure, since `QString::replace()` doesn't have the same callback-style semantics for the replacement text that Python's `re.sub()` offers, and I think that'd be necessary to do this this way. I think if you're going to use regular expressions for this, the only chance is to `split()` the QString into a `QStringList` by newline, then iterate over it and do max. one replacement at a time, using a simple string-based `replace()` after generating the new path out-of-band.

ferdnyc: > Because the path methods are going to be evaluated first, `canonicalPath()` (or `absoluteFilePath()` will be getting the unexpanded `"\\2"` string as their argument — fortunately, `absoluteFilePath()` doesn't check whether the file passed to it actually exists, so it'll just pass it through unmolested.

(That's actually why `absoluteFilePath()` isn't doing the right thing, either. It'll only be called **once** when the arguments to the `QString::replace()` are being assembled, and it'll be asked to generate the `absoluteFilePath()` to `/path/to/some/asset/dir/\2`. Which it will happily do, and then the replacement ultimately becomes this:
```c++
projectContents.replace(pathRegex, R"__("\1": "/path/to/some/asset/dir/\2")__");
```
...A replacement it will happily execute, prefixing each of your matched path strings in an absolute path to the starting location, but never resolving them to remove all of the relative-directory backtracking.)

ferdnyc: I'll say this: I now have a far greater appreciation for just how powerful & flexible Python's `re.sub()` is, simply by virtue of accepting a callable as the replacement argument.

Being able to generate a whole new replacement string on-the-fly for each individual match opens up a whole new dimension in what you can do with regular expressions.

jonoomph:
Oh, and I accidentally started a "review" on this PR, and it's threatening me that if I cancel the review, it will delete all our comments. So... hmmm, I'm gonna just ignore it. :wink:

src/Timeline.cpp
```diff
@@ -73,6 +75,107 @@ Timeline::Timeline(int width, int height, Fraction fps, int sample_rate, int cha
...
+	if (!openshotTransPath.exists()) {
+		throw InvalidFile("PATH_OPENSHOT_INSTALL/transitions could not be found.", openshotTransPath.path().toStdString());
+	}
+
+	// Determine asset path
+	QString asset_name = filePath.baseName().left(30) + "_assets";
+	QDir asset_folder(filePath.dir().filePath(asset_name));
+	if (!asset_folder.exists()) {
+		// Create directory if needed
+		asset_folder.makeAbsolute();
```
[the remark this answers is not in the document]
jonoomph: Fixed, thx!

ferdnyc:
> Oh, and I accidentally started a "review" on this PR, and it's threatening me that if I cancel the review, it will delete all our comments. So... hmmm, I'm gonna just ignore it.

The review functionality gets _really_ weird once reviews have been started on a PR, I've noticed — even normal conversational replies like this one can end up being submitted as reviews, and that's apparently just the way it's meant to work. It's thrown me a few times too, though. Whenever it starts going on about reviews, I just submit everything that way, and it seems to work out.

`¯\_(ツ)_/¯` **#GithubIsConfusing**

src/Timeline.cpp
ferdnyc: @jonoomph 
> One crazy thought I had though, was to add some additional regex expressions, which would match `../` and `../../` and `../../../` and so on (up to a reasonable point), and replace them separately.

That way lies madness, I suspect. Yes, you can match for any <code><var>directoryname</var></code>/../</code> and replace it with simply `/`, but that only works if your paths are non-relative to _begin_ with — IOW, it's only possible after you've already prepended the absolute location of your starting point. So it's minimum two passes through the entire dataset.

The other option for doing this, in my mind, would be to load in the entire JSON data set as-is, and then iterate over the _**Json**_ data structure (I guess it's a nested `Json::Value` tree, really), and do the fixup in-place.

jsoncpp does provide iterators that should theoretically make that pretty simple, and while it is potentially a **lot** of structure to walk, the actual operation becomes bog-simple. Instead of dealing with carefully matching sections of text in a serialized string buffer, you just walk the tree looking for any `path` or `image` keys in dictionary objects. When you find one, you:
1. String-replace any leading `@marker` placeholders with the appropriate path
2. Run the value through `canonicalPath()` and assign it back to itself

Boom, done. You don't even have to _look_ at any of the rest of the data, so other than the time it takes to walk the structure, it's mostly irrelevant how much there actually is.

Whereas a buffer-wide regexp has to parse every single character of the serialization, just to edit a relatively few bytes of its contents. That's gotta be expensive, no matter how efficient the QRegularExpression implementation is (and I'm sure it's very efficient). Especially for complex project files that can get to be up in the dozens or even hundreds of kilobytes of string data.

jonoomph:
Merging, excited about this one!

PR #460, opened 2020-03-09
Adding stale bot for github issue management (libopenshot)

jonoomph:
Experimental support for `stale` plugin, to help with issue management on GitHub. https://github.com/marketplace/stale

Related PR for libopenshot-audio: https://github.com/OpenShot/libopenshot-audio/pull/93

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,7 +1,7 @@
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: 60
+daysUntilStale: 90
 # Number of days of inactivity before a stale issue is closed
-daysUntilClose: 7
+daysUntilClose: 10
 # Issues with these labels will never be considered stale
 exemptLabels:
   - pinned
@@ -16,3 +16,7 @@ markComment: >
   for your contributions.
 # Comment to post when closing a stale issue. Set to `false` to disable
 closeComment: false
+# Only close issues
+only: issues
+# Don't close issues which are assigned to somebody
+exemptAssignees: true
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - enhancement
+# Label to use when marking an issue as stale
+staleLabel: stale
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,18 @@
...
+  - pinned
+  - security
+  - enhancement
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as **stale** because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
```
ferdnyc: I wonder if it's possible to use any sort of variable substitution in this message, to customize it with the actual parameters governing Stale's activities? The repo I first noticed it in had it posting a message something like,
> This issue has been automatically marked as **stale** because it has been more than 30 days since the last update. It will automatically be closed in 7 days if no further activity occurs. Thank you for your contributions.

...Which is a little less brusque and more informative to the users whose issues it's threatening to close. But I have no idea if those parameters were pulled from the config, or if they were simply hard-coded into the text.

Then again, the config is YAML — it supports things like references out of the box. It might be possible to assemble the message using references to the YAML parameter values. I'll play with it a bit.

ferdnyc: Huh, I guess you [really can't](https://stackoverflow.com/questions/25852464/dry-up-yaml-with-variables) — support for any sort of variable replacement would have to be implemented by the bot author, in whatever language they use to consume the configuration. They _might_ have implemented that support, but it's not a built-in language feature. And the bot's docs are... spare. (Though I suppose I could try to remember which repo it was where I originally saw Stale in action, and go look at their config.)

ferdnyc: Well, fortunately that wasn't too hard.

As I feared, they simply [hardcoded the timespan values into the message text](https://github.com/travis-ci/travis-ci/blob/master/.github/stale.yml).

jonoomph: I suppose we can always hardcode a better message in the future too. Especially if any confusion appears around this.

ferdnyc: Confusion, even, I can deal with. "Vehement backlash and incoherent rage" is more my worry. Internet haters gotta internet-hate.

jonoomph:
Merging 'stale-bot' to libopenshot also!

ferdnyc:
Well! Stale sure has been busy. :laughing: 

Which is a good thing, but now I have a _lot_ of notifications for issues I want to revisit.

PR #462, opened 2020-03-11
ReaderBase_Tests: 100% internal coverage

ferdnyc:
This PR completes coverage for the `ReaderBase_Tests.cpp` file _itself_, by adding code to exercise the derived class that's defined in the test source. (It does nothing to affect the coverage of any libopenshot source code, merely the coverage of the source code for the unit test itself.)


ferdnyc:
Merging as painfully uncontroversial.

Issue #456, opened 2020-03-08
Audio corruption at max amplitude


ferdnyc:
When combined amplitude of mixed audio exceeds the maximum (1.0), clipping is the expected result. However, libopenshot appears to be doing something that corrupts the audio beyond simple clipping, with the values swinging around wildly instead of being hard-limited to 1.0. 

The effect can most easily be seen in this Audacity screenshot. (Lower waveform, with the upper being the same mix, but with the volume levels reduced in OpenShot):

![image](https://user-images.githubusercontent.com/538020/76172452-88573400-616c-11ea-82c4-3b2c1ec4bba6.png)

This has the effect of distorting the audio _beyond_ what would simply be heard from normal clipping, causing audible popping and crackling.

cc: @SuslikV , cc: OpenShot/openshot-qt#3256

musteresel:
Is this happening at *every* time where clipping would be expected? (Might be related to the missing samples from some frames I'm currently working on)

ferdnyc:
> Is this happening at _every_ time where clipping would be expected?

Near as I can _tell_, yes, and only when clipping would be expected. It seems like code that's meant to be constraining the amplitude to ±1.0, whenever it goes over/under, is instead making up unpredictable values other than ±1.0.

I don't _think_ it's any sort of **missing** data, per se, just because the distortions are so small. I mean, in that screenshot each labeled tickmark is just `0.005` seconds apart. If data were missing, I'd expect it to be in larger chunks than that.

SuslikV:
There is screenshot at higher zoom:
https://github.com/OpenShot/openshot-qt/issues/3256#issuecomment-599053363

original issue is seems to be this: https://github.com/OpenShot/openshot-qt/issues/1069 so, it is long time issue.

I can look at this (no time deadlines), but if you have better skills...

ferdnyc:
@SuslikV 

If you're able to take a look at it, that would be great, thanks! It'd be a few days at least before I'd be able to get to it.

I made the mistake of discovering a [new tool (`clazy`)](https://github.com/KDE/clazy) for scanning and checking Qt code, and after about 12 hours I've got it running on the libopenshot codebase with no warning messages at all. (Except for the couple I turned off, because like _all_ of our polymorphic classes are copyable and we'll just have to live with that.)

[...]

PR #466, opened 2020-03-16
Fix distortion during mix of audio

SuslikV:
Fixes: https://github.com/OpenShot/libopenshot/issues/456

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -1525,11 +1525,23 @@ void FFmpegWriter::write_audio_packets(bool is_final) {
 			// Calculate total samples
 			total_frame_samples = samples_in_frame * channels_in_frame;
 
-			// Translate audio sample values back to 16 bit integers
-			for (int s = 0; s < total_frame_samples; s++, frame_position++)
-				// Translate sample value and copy into buffer
-				all_queued_samples[frame_position] = int(frame_samples_float[s] * (1 << 15));
+			// Translate audio sample values back to 16 bit integers with saturation
+			float valF;
+			int16_t conv;
+			const int16_t max16 = 32767;
+			const int16_t min16 = -32768;
+			for (int s = 0; s < total_frame_samples; s++, frame_position++) {
+				valF = frame_samples_float[s] * (1 << 15);
+				if (valF > max16)
+					conv = max16;
+				else if (valF < min16)
+					conv = min16;
+				else
+					conv = int(valF + 32768.5) - 32768; // +0.5 is for rounding
 
+				// Copy into buffer
+				all_queued_samples[frame_position] = conv;
+			}
 
 			// Deallocate float array
 			delete[] frame_samples_float;
--- a/src/FrameMapper.cpp
+++ b/src/FrameMapper.cpp
@@ -785,10 +785,23 @@ void FrameMapper::ResampleMappedAudio(std::shared_ptr<Frame> frame, int64_t orig
 	// Create a new array (to hold all S16 audio samples for the current queued frames)
  	int16_t* frame_samples = (int16_t*) av_malloc(sizeof(int16_t)*total_frame_samples);
 
-	// Translate audio sample values back to 16 bit integers
-	for (int s = 0; s < total_frame_samples; s++)
-		// Translate sample value and copy into buffer
-		frame_samples[s] = int(frame_samples_float[s] * (1 << 15));
+	// Translate audio sample values back to 16 bit integers with saturation
+	float valF;
+	int16_t conv;
+	const int16_t max16 = 32767;
+	const int16_t min16 = -32768;
+	for (int s = 0; s < total_frame_samples; s++) {
+		valF = frame_samples_float[s] * (1 << 15);
+		if (valF > max16)
+			conv = max16;
+		else if (valF < min16)
+			conv = min16;
+		else
+			conv = int(valF + 32768.5) - 32768; // +0.5 is for rounding
+
+		// Copy into buffer
+		frame_samples[s] = conv;
+	}
 
 
 	// Deallocate float array
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -1627,9 +1627,8 @@ void FFmpegReader::ProcessAudioPacket(int64_t requested_frame, int64_t target_fr
 			else
 				partial_frame = true;
 
-			// Add samples for current channel to the frame. Reduce the volume to 98%, to prevent
-			// some louder samples from maxing out at 1.0 (not sure why this happens)
-			f->AddAudio(true, channel_filter, start, iterate_channel_buffer, samples, 0.98f);
+			// Add samples for current channel to the frame.
+			f->AddAudio(true, channel_filter, start, iterate_channel_buffer, samples, 1.0f);
 
 			// Debug output
 			ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::ProcessAudioPacket (f->AddAudio)", "frame", starting_frame_number, "start", start, "samples", samples, "channel", channel_filter, "partial_frame", partial_frame, "samples_per_frame", samples_per_frame);
--- a/tests/FFmpegReader_Tests.cpp
+++ b/tests/FFmpegReader_Tests.cpp
@@ -76,8 +76,8 @@ TEST(FFmpegReader_Check_Audio_File)
 	CHECK_CLOSE(0.0f, samples[50], 0.00001);
 	CHECK_CLOSE(0.0f, samples[100], 0.00001);
 	CHECK_CLOSE(0.0f, samples[200], 0.00001);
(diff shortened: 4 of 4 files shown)


SuslikV:
New sample values for the test sound `piano.wav` form the `FFmpegReader_Tests.cpp` (samples numbers starts from 0):
```
samples[230] = 0.16406
samples[300] = -0.06250
```
Because no gain of x0.98 any more.

src/FFmpegWriter.cpp
```diff
@@ -1525,11 +1525,23 @@ void FFmpegWriter::write_audio_packets(bool is_final) {
...
+				if (valF > max16)
+					conv = max16;
+				else if (valF < min16)
+					conv = min16;
+				else
+					conv = int(valF + 32768.5) - 32768; // +0.5 is for rounding
+
+				// Copy into buffer
+				all_queued_samples[frame_position] = conv;
+			}
```
ferdnyc: C++ defines `INT16_MAX` and `INT16_MIN` in the standard. (They're technically marked as optional, but if you `#include <cinttypes>` they should be there in any sane compiler.) So, if nothing else those can be used in place of `max16` and `min16`.

But you're absolutely right that it's a problem of bad limiting — it's really the `int()` there that's the killer. It allows the value to temporarily exceed `INT16_MIN`/`INT16_MAX` _during_ the computation, and then the overflow will produce bad values. Setting an `int16_t` variable to `int(1.1 * INT16_MAX)` doesn't produce `32767`, it produces `-29493`.

However, if we don't use `int()` for the math, the value will be limited correctly. This should work, without the temporary variables or manual limiting:
```c++
# At the top of the file
#include <cinttypes>

for (int s = 0; s < total_frame_samples; s++, frame_position++) {
    all_queued_samples[frame_position] = int16_t(frame_samples_float[s] * INT16_MAX));
}
```
(`INT16_MAX` is what `(1 << 15)` is manually computing.)

I was going to say that otherwise `valF` and/or `conv`should be declared _inside_ the loop for multithreading safety (despite the extra allocations), but this entire function is wrapped in `#pragma omp task firstprivate(is_final)` (and it contains plenty of other unsafe allocations already), so that's a moot point. Still, avoiding them is cleaner.

ferdnyc: Er, sorry, `#include <cinttypes>` is for the `std::fprintf` format constants. The `_MIN/MAX` typedefs should be included using `#include <cstdint>`. (Which cinttypes no doubt includes, itself, so it works either way. But it's more correct to use the cstdint header.) Above comment corrected.

ferdnyc: If you want to enforce the rounding direction, C++11 allows a [floating-point environment](https://en.cppreference.com/w/cpp/numeric/fenv/feround) to be established and configured. So, this would ensure that values `> N.5` are always rounded upwards to `N+1`:
```c++
#include <cfenv>
for (int s = 0; s < total_frame_samples; s++, frame_position++) {
    #pragma STDC FENV_ACCESS ON
    std::fesetround(FE_TONEAREST);
    all_queued_samples[frame_position] = int16_t(frame_samples_float[s] * INT16_MAX));
}
```
The scope will automatically be limited to the enclosing block (the inside of the for loop).

ferdnyc: \*sigh* Of course, it turns out that `#pragma` isn't supported, and `std::fesetround()` currently has no effect, in GCC or Clang. **#LAME.**

So, yeah, adding 0.5 seems to be the safest option for rounding.
```c++
for (int s = 0; s < total_frame_samples; s++, frame_position++) {
    all_queued_samples[frame_position] = int16_t(frame_samples_float[s] * INT16_MAX + 0.5));
}

Comment (ferdnyc) [reply src/FFmpegWriter.cpp]: ARRGH. I'm wrong. It turns out the compiler is smart and fixes things, but only when it can do so trivially at compile time. e.g. this:
```c++
    int16_t out[4] = {
        int16_t(1.1 * (1 << 15)),
        int16_t(1.1 * INT16_MAX),
        int16_t(1.1 * (1 << 15) + 0.5),
        int16_t(1.1 * INT16_MAX + 0.5)
    };
    std::cout << std::setw(5) << 1.1 << ": "
              << std::setw(6) << out[0] << " "
              << std::setw(6) << out[1] << " "
              << std::setw(6) << out[2] << " "
              << std::setw(6) << out[3] << "\n";
    int16_t out2[4] = {
        int16_t(-1.1 * (1 << 15)),
        int16_t(-1.1 * INT16_MAX),
        int16_t(-1.1 * (1 << 15) + 0.5),
        int16_t(-1.1 * INT16_MAX + 0.5)
    };
    std::cout << std::setw(5) << -1.1 << ": "
              << std::setw(6) << out2[0] << " "
              << std::setw(6) << out2[1] << " "
              << std::setw(6) << out2[2] << " "
              << std::setw(6) << out2[3] << "\n";
```
Produces the correctly limited values:
```
  1.1:  32768  32768  32768  32768
 -1.1: -32768 -32768 -32768 -32768
```
But this:
```c++
void x(float f) {
    int16_t out[4] = {
        int16_t(f * (1 << 15)),
        int16_t(f * INT16_MAX),
        int16_t(f * (1 << 15) + 0.5),
        int16_t(f * INT16_MAX + 0.5)
    };
    std::cout << std::setw(5) << f << ": "
              << std::setw(6) << out[0] << " "
              << std::setw(6) << out[1] << " "
              << std::setw(6) << out[2] << " "
              << std::setw(6) << out[3] << "\n";
    int16_t out2[4] = {
        int16_t(-f * (1 << 15)),
        int16_t(-f * INT16_MAX),
        int16_t(-f * (1 << 15) + 0.5),
        int16_t(-f * INT16_MAX + 0.5)
    };
    std::cout << std::setw(5) << -f << ": "
              << std::setw(6) << out2[0] << " "
              << std::setw(6) << out2[1] << " "
              << std::setw(6) << out2[2] << " "
              << std::setw(6) << out2[3] << "\n";
}
```
...Does *NOT*, if you call `x(1.1);`:
```
  1.1: -29492 -29493 -29491 -29492
 -1.1:  29492  29493  29492  29493
```

Apologies for the misinformation. Marking this as "Resolved" since it's not helpful commentary as it turns out.

ferdnyc: Let's try this again, with code I've actually tested first:

This can be implemented without the temporary variables, using `std::min()`, `std::max()`, and the `INT16_MIN` / `INT16_MAX` typedefs. e.g:
```suggestion
			for (int s = 0; s < total_frame_samples; s++, frame_position++) {
				all_queued_samples[frame_position] = std::min(INT16_MAX,
				    std::max(INT16_MIN, int(frame_samples_float[s] * (1 << 15)))
				);
			}
```
(Using `(1 << 15)` instead of `INT16_MAX` does make a difference when `frame_samples_float[s]` is exactly `-1.0`, turns out.)

ferdnyc: Or, with the rounding (which does seem like a good idea):
```c++
all_queued_samples[frame_position] = std::min(INT16_MAX,
    std::max(INT16_MIN, (int(frame_samples_float[s] * (1 << 15) + 32768.5) - 32768))
);

Comment (SuslikV) [reply src/FFmpegWriter.cpp]: My code may has better optimization by compiler for branching prediction routines.

**Comment by SuslikV** [comment]
Resampler has similar code, fixing this too.

File: src/FFmpegWriter.cpp (hunk 1 above)
Comment (ferdnyc) [reply src/FFmpegWriter.cpp]: Eh. I'm _surprised_ that it doesn't optimize better, but I'll buy that I guess. (You tested even with `-O2` / `-DCMAKE_BUILD_TYPE=Release`, and it made a measurable difference?)

Normally the temporary variables would make me nervous about multithreading, but like I said this entire function is already marked non-threadsafe so that's a moot point. OK, works for me.

File: src/FrameMapper.cpp [hunk 2]
```diff
@@ -785,10 +785,23 @@ void FrameMapper::ResampleMappedAudio(std::shared_ptr<Frame> frame, int64_t orig
...
-	// Translate audio sample values back to 16 bit integers
-	for (int s = 0; s < total_frame_samples; s++)
-		// Translate sample value and copy into buffer
-		frame_samples[s] = int(frame_samples_float[s] * (1 << 15));
+	// Translate audio sample values back to 16 bit integers with saturation
+	float valF;
+	int16_t conv;
+	const int16_t max16 = 32767;
+	const int16_t min16 = -32768;
+	for (int s = 0; s < total_frame_samples; s++) {
```
ferdnyc: Wait, there are **two** places where this exact same conversion is done!?

...Why the heck does `openshot::Frame::GetInterleavedSamples()` return a float array, then? Instead of at least having an overload that converts it first, and passes back an array in the format we actually need!? ARRRGH, I hate cut-and-paste code duplication.

Well, anyway, that part's not affected by these changes, so not really relevant here.

ferdnyc:
I'm good with it, and fixing the distortion will be a big improvement. Merging. Thanks!

PR #467, opened 2020-03-18
Various: Remove unused variables

ferdnyc:
Eliminate several data members and local variables across the codebase, because they were flagged by static analysis tools as never being used anywhere in the code. First part of my massive [clazy](https://github.com/KDE/clazy)/[Codacy](https://codacy.com) cleanup sweep.

Also, the `src/bindings/ruby/test.rb` file is renamed to `src/examples/Example.rb`. I hadn't intended to include that in this PR, but it makes sense so I'll leave it.


ferdnyc:
Whoops, missed one. Fixed.

ferdnyc:
They're unused, it's right in the name. Merging.

PR #468, opened 2020-03-19
Clip_Tests: Remove try/catch blocks

ferdnyc:
UnitTest++ has its own exception handling (you can even `CHECK_THROW()` to test that a call throws an exception as expected), so the try/catch blocks in the Clip unit tests serve no purpose except to add unreachable code.

This PR replaces them with `REQUIRE CHECK_EQUAL(true, success);` checks on the result var from the previous call (the one that, when it was found false, the test used to `throw()`).

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -9,7 +9,8 @@ coverage:
 ignore:
   - "/src/examples"
   - "/src/Qt/demo"
-  - "/thirdparty"
+  - "/thirdparty/jsoncpp/*.cpp"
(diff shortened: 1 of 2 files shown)


ferdnyc:
I believe this should rase the **internal** coverage of our unit tests to 100%, meaning that every line of the _test code_ (not the _**tested**_ code) is reached during the test run. As it should be.

Now we  just have to get the _overall_ coverage up.

ferdnyc:
Merging, as the change has the intended effect.

PR #470, opened 2020-03-20
Add version gating for Zmq and Qt deprecations

ferdnyc:
This PR resolves both of the deprecations showing up in the codebase when building with current dependency versions.

* Qt's <strike>`QImage::sizeInBytes()`</strike>, used once in QtImageReader, <strike>is deprecated</strike> starting with Qt 5.10. (Correction: `QImage::byteCount()` is deprecated, `QImage::sizeInBytes()` is its replacement.)
* ZeroMQ deprecated the `send()` form we've been using starting with version 4.3.1.

The replacement ZMQ API includes a second `flags` argument. We take advantage of the new field to specify `zmq::send_flags::dontwait`, flagging the message for immediate delivery. (It's possible we could use `dontwait` in older ZMQ versions, so perhaps the version check could be broadened. It's also possible that `dontwait` has a negative performance impact, in which case it might be better to remove it. Unit tests would be helpful, but they do not yet exist.)

### DIFF
--- a/src/QtImageReader.cpp
+++ b/src/QtImageReader.cpp
@@ -111,7 +111,12 @@ void QtImageReader::Open()
 		info.has_audio = false;
 		info.has_video = true;
 		info.has_single_image = true;
-		info.file_size = image->byteCount();
+		#if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
+			// byteCount() is deprecated from Qt 5.10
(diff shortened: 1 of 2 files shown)


ferdnyc:
Oh, hey, Travis finally woke up. Neat.

ferdnyc:
Merging this, as these are both small, future-proofing changes to the code.

PR #475, opened 2020-03-22
Memory-safety fixes to FFmpegWriter, FrameMapper, DummyReader

ferdnyc:
The biggest change here is actually the least consequential: I renamed `audio_codec` and `video_codec` to `audio_codec_ctx` and `video_codec_ctx` throughout FFmpegWriter, because they both have type `AVCodecContext`. There _are_ other variables in use with type `AVCodec`, and it hurt my brain too much trying to follow the code when the variables called "codec" weren't codecs.

After doing that, I was able to read the code, so I fixed a possible memory leak flagged by valgrind, where new contexts were assigned to a pointer without checking that it didn't _already_ have a context assigned. If it does, we free it as required by the FFmpeg API.

In addition, two reads on unused variables are fixed:
1. `is_open` in FrameMapper (Eliminated, because that read was the **only** time it was ever used, _period_ — its `IsOpen()` method returns the value of `reader->is_open`.)
1. `is_open` in DummyReader (Added to the constructors' initializer lists.)

Eliminating those two issues cleared up something like 25 valgrind errors in `openshot-test` (thanks to how often FrameMapper's destructor gets called).


ferdnyc:
Oh, right, I forgot to mention the AVRationals. Apparently some compilers don't like the construct <code>(AVRational){1, <var>value</var>}</code> very much, so those got flagged on analysis. Fortunately, the ffmpeg developers are aware of that, and provided `av_make_q()` for just that reason.

eisneinechse:
I think this is a good idea.
I myself have trouble remembering (researching) what is used.

ferdnyc:
Thanks @eisneinechse , sounds good to me. Merging!

PR #476, opened 2020-03-25
ZeroMQ's std::string support is too new

ferdnyc:
Whoops. Seems I broke ZeroMQ on both the Windows and macOS builders, trying to use `std::string` with `zmq::message_t`. (In my defense, it works fine on Fedora 31! Must be a recent feature.)

I've reverted to the old `memcpy()` version of message construction, for all ZeroMQ versions.

### DIFF
--- a/cmake/Modules/FindZeroMQ.cmake
+++ b/cmake/Modules/FindZeroMQ.cmake
@@ -5,14 +5,16 @@ pkg_check_modules(PC_LIBZMQ QUIET libzmq)
 set(ZeroMQ_VERSION ${PC_LIBZMQ_VERSION})
 
 find_path(ZeroMQ_INCLUDE_DIR zmq.h
-	  PATHS ${ZeroMQ_DIR}/include
-                ${PC_LIBZMQ_INCLUDE_DIRS})
+  PATHS
+    ${ZeroMQ_DIR}/include
+    ${PC_LIBZMQ_INCLUDE_DIRS})
 
 find_library(ZeroMQ_LIBRARY
-	     NAMES zmq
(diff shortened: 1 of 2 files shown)


ferdnyc:
Still broken on non-Linux. I just added version _reporting_ to the ZeroMQ find module in CMake (we were capturing it, before, but not passing it back to the parent context), so hopefully I can figure out what version the other builders are using and why it's breaking.

ferdnyc:
Well, this is very frustrating. The macOS builder claims to be running 4.3.2, same as my Linux machine... and Windows has 4.3.1 installed, which is _documented_ as the release where the previous `send()` signature was deprecated. Yet, the new signature isn't found, so it's not compiling!

...I wonder if perhaps there's a version mismatch between ZeroMQ and **cppzmq**, the C++ bindings. @jonoomph , is there any chance that ZeroMQ was upgraded on the Mac/Win builders, but _without_ upgrading the C++ bindings header from cppzmq? That's the only thing I can think of that would explain this.

ferdnyc:
Mmm, I think that's it. Because looking at the builders, I notice neither Win nor Mac was complaining about the deprecated form of `send()` _before_ I fixed it. Which, if they were fully installed with ZeroMQ 4.3.1 or 4.3.2, they should've been. The old form was deprecated starting in 4.3.1.

ferdnyc:
Shot @jonoomph an email detailing my theory. Let's wait a bit to see if he might be able to fix the Win/Mac builders, before I back out these changes and #470.

ferdnyc:
@jonoomph is on the case.

PR #481, opened 2020-03-26
Update copyright to 2020

ChillerDragon:


ChillerDragon:
Btw on openshot.org it still says:

OpenShot™ Video Editor. Copyright © 2008-**2017** OpenShot Studios, LLC. All Rights Reserved | Privacy | Terms | Language

ferdnyc:
Yeah, the website isn't managed here, that's all solo-@jonoomph.

As for copyright in the files, coincidentally enough I was just reading Thomas E. Dickey's [extensive thoughts](https://invisible-island.net/personal/copyrights.html) on that last night. (Dickey is the maintainer of ncurses, has been a free software developer since the 1970s, and has spent most of a lifetime looking into these things.)

He has a very different — and informed by actual case law and discussions with the US Copyright Office — take on the file copyrights issue:

> #### Personal practice
> I use scripts to check for consistency when releasing a program:
> 
> * one script inspects all files and compares its modification time with the copyright notice (of course, I use tools which preserve modification times).
> * other scripts inspect files which are used to show the collective copyright dates, e.g., in version messages.
> 
> ### Restrictions
> Copyright law does not require a notice. If there is no no explicit license granting permission, fair use only is permitted, even for web pages. Fair use as such always requires attribution, and is limited to small portions of a document.
> 
> Just to be helpful, a couple of FAQS from my website (ncurses and xterm) have permissive licenses.
> 
> ### Abuses
> There are several issues to mention; examples are suitable in a few cases.
> 
> #### Backdating copyrights
> A copyright notice for an unpublished work is pointless. The notice is used for published works.
> 
> On occasion you may notice (pun intended) someone publishing a computer program and later adding copyright notices giving earlier dates than the first date it was made available. That of course is done to “prove” that the work was in development at some early date.
> 
> #### Updating the copyrights
> I have noticed some developers “updating the copyrights” on works which have not changed in a given year. Do not do that. A copyright notice tells someone when a work was published. If you did not change the work, there is no reason to keep changing the dates.

IOW:
* The overall copyright statement for OpenShot and/or libopenshot should list the aggregate years of active development, currently 2008-2020.
* The statements in the individual _files_ should only be updated for those which have actually been modified since January 1. (And I was wrong to bulk-update them to 2019 last year.)
* It is not expected that the individual files will all have the same statement, or that they will match the overall statement, and in fact it's both preferable and useful if they **don't**.

ferdnyc:
(The other takeaway from that Dickey page, as it applies to our source tree, is that the _start_ copyright date in each individual file should really be the year it was created, rather than "2008" across the board. Especially since few or possibly **none** of the current source files even existed in 2008. Most were created in 2014-2015, if not later.)

ChillerDragon:
yikes sounds like work

PR #482, opened 2020-03-26
Ignore ctags file

ChillerDragon:


ferdnyc:
Since there's no `ctags` file in the repo, it doesn't really make sense to include an ignore rule for it.

As I only learned myself a few months ago, you can incorporate any additional _local_ ignore rules you like (without throwing your branch out of sync with the repo) by adding them to `.git/info/exclude`.

ferdnyc:
> Since there's no `ctags` file in the repo

Er, and the repo also doesn't normally generate a `ctags` file, to complete that thought.

The best uses for a checked-in `.gitignore` are:
1. to protect checked-in files that get regenerated during development, but shouldn't be updated
1. to exclude files commonly created during the development process, to avoid them getting accidentally checked in by people in the habit of doing blanket `git add`s.

(Though CMake's out-of-source builds are a huge help in preventing both of those in the first place, since it's one of the guiding principles of CMake that an impenetrable wall be maintained  separating the source (input) dir from the build (output) dir.)

Still, a ctags file arguably falls under point 2, so it can't hurt. What the heck, merging.

ChillerDragon:
When I do ``git commit`` it opens vim which then creates a tag file that then gets tracked the next time I do ``git add .``

in the name of all ctags users thanks for merge <3

ferdnyc:
> then gets tracked the next time I do `git add .`

Well, that's kind of what I was getting at with the blanket adds. I always try to talk people out of doing mass-adds, since that's how unrelated changes sneak into commits and PRs.

Targeted adds on individual files are best, so each commit comprises only intentional, explicitly-selected changes. But failing that (or at times when it's unrealistic), using `git add -u` (for `--update`) is preferable to a repo-wide `git add .`, since `-u` will ignore untracked files.

Or you can also just jump straight to `git commit -a`, which will have the same effect of adding all known files, but ignoring untracked ones.

PR #483, opened 2020-03-27
Remove old command to fix Win64 builder

ferdnyc:
When merging #458 I somehow left an outdated command in the `.gitlab-ci.yml` Win64 command list that now references a nonexistent path, causing the build to fail.


ferdnyc:
Travis is suffering a [Major Outage](https://www.traviscistatus.com/) right now anyway, so things aren't exactly operating like clockwork.

ferdnyc:
This PR's build passed, but the Travis API was out to lunch so nobody knows about it. I'll close and refile to kickstart it.

PR #485, opened 2020-03-27
Fix Win64 Gitlab builds

ferdnyc:
When merging #458 I somehow left an outdated command in the `.gitlab-ci.yml` Win64 command list that now references a nonexistent path, causing the build to fail.


ferdnyc:
Ugh, FINALLY. Frickin' Travis.

PR #495, opened 2020-04-08
Set decoding and encoding color details

SuslikV:
1) This solves issue when imported files wrongly decoded as BT.601 videos ("greenish" result in the preview of OpenShot).

2) This solves issue when imported 4:2:0 files has "stairs" artifacts on colored objects (noticeble in Save Frame or image sequence Export of OpenShot).

**Edit:**
"Reference" input 4:2:0 mp4 file of BT.709 space, transfer, primaries, chroma origin is left, range is partial.
In mpv Player orig

What was wrong? Colors!
OpenShot's _Video Preview_. Just screenshot (I see it the same "greenish" on my screen). You can attempt to adjust colors but this will be mistake, because preview shouldn't change colors of the input but can has scaling artifacts (right edge has color spill over the grey about one pixel width - this is fast_bilinear optimizations of FFmpeg, so it's OK for preview)
Preview in OpenShot 251 of 420 file bt709 tv

OpenShot's _Save Frame_ or _Image Sequence Export_ colors are wrong. And these images can be used for production. This is no good. Also, some simplifications (no chroma scale down back to 1/2) makes the final output even worse - details is here: https://github.com/OpenShot/openshot-qt/pull/3335#issuecomment-610788981
OpenShot 251 SaveFrame or Image seq Export

Image import - movie Export (23 crf, mp4 720p, playback in mpv, screenshot - I see it the same "dark green" on my screen, file was not flagged for color space), So better to not import PNG images (RGB) into OpenShot or colors in export will be wrong
In mpv Player OpenShot 251 Export Image only file

------------------

I hope that next changes will improve quality for the _Export_ (both: scale and colors) and _Preview_ (only colors). I can say that only for OpenShot's preview there is enough _Point_ scaling, because it is only preview... (and any source can be scaled up by Qt transform that itself set to bilinear). But someone may want to use library somewhere else, and may not know about `HIGH_QUALITY_SCALING` switch in _libopenshot_, so I decided to not make changes to `scale_mode=SWS_FAST_BILINEAR` when lower quality is in use.

Some thoughts are inspired by Jason Summers ( entropymine.com/jason/ ) and Charles Poynton (if site even needed here?)

**info:** This is part of the: https://github.com/OpenShot/libopenshot/pull/219

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -1397,9 +1397,15 @@ void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
 			av_opt_set_int(sws_ctx, "srch", info.height, 0);
 			av_opt_set_int(sws_ctx, "dstw", width, 0);
 			av_opt_set_int(sws_ctx, "dsth", height, 0);
+
+#if LIBSWSCALE_VERSION_MAJOR >= 4 && LIBSWSCALE_VERSION_MINOR >= 6
+			//FFMPEG 3.3+
 			av_opt_set_pixel_fmt(sws_ctx, "src_format", pix_fmt, 0);
 			av_opt_set_pixel_fmt(sws_ctx, "dst_format", PIX_FMT_RGBA, 0);
-
+#else
+			av_opt_set_int(sws_ctx, "src_format", (int) pix_fmt, 0);
+			av_opt_set_int(sws_ctx, "dst_format", (int) PIX_FMT_RGBA, 0);
+#endif
 			// Assuming non-4:4:4 import
 			if (chrH != 0 or chrV != 0) {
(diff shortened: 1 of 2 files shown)


SuslikV:
For old FFmpeg the option `AV_OPT_TYPE_PIXEL_FMT` is actually `AV_OPT_TYPE_INT`. Next day I'll see if there simple workarounds.

And maybe few colored pictures needed here ^_^

SuslikV:
Current _develop_ was changed. I will update the code later.

SuslikV:
Updated on top of the current _develop_.

jonoomph:
Closing this unless someone wants to test all this code. I'm not sure the best way to test it, and I'm too scared to merge without testing it.

PR #496, opened 2020-04-13
Transform Improvements (Keyframe origin point)

jonoomph:
Adding keyframable origin point (for shear and rotation). Used in https://github.com/OpenShot/openshot-qt/pull/3366. Both rotation and shear use an origin point to control how those tools work. This allows the user to not only set a custom origin point (as a % of the original clip), but it allows the origin to be animated, which allows for some really great animations!

origin_x: **0.0** = left side of clip
origin_x: **1.0** = right side of clip

origin_y: **0.0** = top side of clip
origin_y: **1.0** = bottom side of clip

### DIFF
--- a/src/Timeline.cpp
+++ b/src/Timeline.cpp
@@ -677,6 +677,8 @@ void Timeline::add_layer(std::shared_ptr<Frame> new_frame, Clip* source_clip, in
 	y += (Settings::Instance()->MAX_HEIGHT * source_clip->location_y.GetValue(clip_frame_number)); // move in percentage of final height
 	float shear_x = source_clip->shear_x.GetValue(clip_frame_number);
 	float shear_y = source_clip->shear_y.GetValue(clip_frame_number);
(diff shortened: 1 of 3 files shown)


src/Timeline.cpp
```diff
@@ -677,28 +677,31 @@ void Timeline::add_layer(std::shared_ptr<Frame> new_frame, Clip* source_clip, in
...
+		transform.translate(x, y);
 		transformed = true;
 	}
 
-    if (!isEqual(x, 0) || !isEqual(y, 0)) {
-        // TRANSLATE/MOVE CLIP
-        transform.translate(x, y);
-        transformed = true;
-    }
+	if (!isEqual(r, 0) || !isEqual(shear_x, 0) || !isEqual(shear_y, 0)) {
```
SuslikV, line 695: This code has related issue:  https://github.com/OpenShot/openshot-qt/issues/2794

PR #497, opened 2020-04-15
Rename our BaseException class to ExceptionBase, for Python

ferdnyc:
`pylint` pointed out that `BaseException` is a Python standard library exception class (not quite a reserved word, but pretty close), so it's not a great idea to redefine that name in our bindings. 

This PR renames it to `ExceptionBase`, which is actually more in keeping with our standard class naming system anyway. (Compare `ReaderBase`, `EffectBase`, `CacheBase`, etc.)

### DIFF
--- a/include/Exceptions.h
+++ b/include/Exceptions.h
@@ -38,16 +38,16 @@ namespace openshot {
 	/**
 	 * @brief Base exception class with a custom message variable.
 	 *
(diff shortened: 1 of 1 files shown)


ferdnyc:
This is pretty minor, and it's best if we avoid name clashes in Python. Merging.

PR #498, opened 2020-04-16
Add all missing override markers on derived-class methods

ferdnyc:
Turns out, some Clang++ versions are _much_ pickier than GCC, and will complain mightily about missing `override` annotations in derived classes. Because this is a header-file issue, the same set of warnings tend to repeat over, and over, and _over_ throughout the build.

This PR silences them all, with appropriate application of `override` wherever necessary.

### DIFF
--- a/include/ChunkReader.h
+++ b/include/ChunkReader.h
@@ -132,7 +132,7 @@ namespace openshot
 		ChunkReader(std::string path, ChunkVersion chunk_version);
 
 		/// Close the reader
-		void Close();
+		void Close() override;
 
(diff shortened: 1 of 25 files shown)


ferdnyc:
I was mistaken, blaming Clang for this — I forgot that I set libopenshot-audio to build with `-Wall -Wpedantic` flags in Travis, and it's when applying those same flags to the libopenshot build that the compiler (any of them) gets mouthy about all the missing `override`s. Still, this PR silences those warnings, and lets us build with better code-quality checks enabled. Merging.

PR #499, opened 2020-04-16
Make debug handling match libopenshot-audio

ferdnyc:
This PR adds the debug-mode handling logic from the libopenshot-audio CMakeLists to the libopenshot build, ensuring that the two projects are build the same when built under the same circumstances.

(Juce's pickiness about having a `#define` for either `DEBUG` or `NDEBUG` in macOS builds also affects libopenshot, when the two libraries are being linked together, so it helps to have them both on the same page.)


ferdnyc:
Going to merge this, as it has implications for some macOS build experiments I'm doing.

PR #501, opened 2020-04-17
WIP: Add Mac & Win CI builds with Travis, AppVeyor

ferdnyc:
Once again, mirroring OpenShot/libopenshot-audio#96, _this_ PR adds not only Windows building in AppVeyor, but also introduces a rudimentary form of "pseudo-chained" macOS building as well.

More than the usual set of caveats begin to apply, however.
1. The unit tests are disabled under Windows, because (even after I finally managed to get UnitTest++ _building_ at all; there's no MSYS package for it) they're currently failing, for reasons that aren't completely clear to me.<sup>1</sup> You might say that the lack of unit tests makes doing CI builds under Windows sort of pointless, and I would be forced to agree. But, hopefully the issues with the tests can be solved.
1. I had to disable the Ruby bindings under both Windows and macOS, because they also failed to build if attempted. On the macOS side our old friend `isinfinite` decided to start breaking things again, and there were some other issues with the Ruby includes in XCode on Travis. In the end I just gave up on trying to figure out what the problem was, though.
1. Both the Windows and macOS builds are, as I said, "pseudo-chained". Basically how it works is, prior to building libopenshot itself, the CI run will first check out, build, and install libopenshot-audio. Completely, from scratch, from the `develop` branch. Every time.
    It's a slow process, though I may be able to achieve minor speedups using the AppVeyor and Travis cache feature(s) to store and re-use previously-built libopenshot-audio installs. I had that half-working, but ended up tearing it out for this PR because it was causing hard-to-diagnose build issues. I figured best to get it working, period, first, and then we can experiment with improvements to the process.
1. Building on Windows and macOS involves installing all of the dependencies for every build, using either MSYS or Homebrew respectively. This process _also_ takes an extremely long time, relative to installing Linux dependencies om Ubuntu via `apt`. Also, both packaging systems have to be updated internally before they can install anything else, which _really_ sucks time. Homebrew in particular takes something like _**5-6 minutes**_ just to sync its catalog. But, if the update isn't run — again, _every_ time — then the Travis Homebrew plugin will actually break and fail the build, since they update their macOS system images so infrequently that there are periods where it ends up becoming completely incompatible with the central repo until they update it again. We are currently in such a period.

The same vestigial NuGet packaging I included in the PR on libopenshot-audio is here in this repo as well, though it's just as non-functional and abandoned.

OH!! I forgot to mention an incredibly small, but very important change: Both of these PRs also finally pull the trigger on moving the library targets' `RUNTIME DESTINATION` from `<prefix>/lib` to `<prefix>/bin`, which is the "correct" location for DLL installs under Windows.

I _believe_  `FindOpenShotAudio.cmake` should already support detecting a libopenshot-audio install with the DLL in `bin/` without any changes. If it does need changes, they'll be minor and I can easily throw them in. However, this might affect @SuslikV and anyone else building libopenshot on Windows, so fair warning.

It'll also definitely have implications for the OpenShot packaging scripts, upon merge, but again the changes are ultimately minor and easily made.

#### Notes
1. Part of the unit test failure issue can be attributed to the Settings unit tests. The settings that get modified there will persist throughout the unit test run, and they affect the rest of the tests. (Singletons are notoriously hard to unit test for this reason.) Especially since one of the changes accidentally enables hardware accelerated decoding, on Windows. But even if I rip out the settings tests, there are failures.


ferdnyc:
Ugh, I hate stupid typos. Third time's the charm, hopefully? :crossed_fingers:

ferdnyc:
Third time _was_ the charm, though I think I need to rearrange the Travis matrix so macOS isn't last, now that our list of builds to run once again exceeds the concurrency max of 5 jobs. Right now macOS is the one that will be delayed until a previous job completes, which is bad because it also takes more than twice as long to run as any of the Linux builds.

SuslikV:
Affects me? ^_^ it will be much easier to drop it forever then. Than to follow the code changes... There was the rise, now there should be the fall. Thank you @ferdnyc for personal warning. It is first time I was warned that something affects me.

ferdnyc:
Withdrawing this, as it can now be done more easily using Github Actions CI.

Issue #471, opened 2020-03-20
Developers RFC: Includes reorganization


ferdnyc:
I've been thinking about a major reorg to the repo header file management, for a couple of reasons (which I'll explain). So I wanted to post my thoughts here and invite feedback/commentary from both the other libopenshot developers, _and_ our library consumers. (Though my current plans have considered the impact on the latter, and include pains taken to keep it to a minimum, at least short-term.) As well as, of course, anyone else who's interested and can offer any relevant insights or experience.

#### Motivations
My concerns regarding the current header setup are,
1. As I detail in #469, relative include paths can be problematic. And relative include paths that _leave_ the root of the source tree can be disastrous. Depending where that path is anchored, you could end up with a file from almost anywhere. <code>#include "../../../include/<var>anything</var>"</code> is just a bad idea and a recipe for problems.
1. _Without_ the relative paths, all we have are bare filenames, and that's a poor way to organize a set of API headers. `#include "Color.h"`? Do you know how many possible matches there could be for that name?<sup>**1**</sup>
1. Related to the previous item: The include paths used by library consumers look _nothing_ like the paths we use internally. That means there are aspects of our code that are simply not tested at any point during development. In fact, as internal developers we never even get a _glimpse_ of what it's like to code against libopenshot as an external consumer. That's a troubling hole in the development plan.
1. The whole `src/` vs. `include/` structure for a codebase is pretty dated, these days, and it shows in the hoops we have to jump through to make it work. The CMake/Qt `AUTOMOC`, for example, by default will scan for and process Qt classes in any implementation-header file pair it finds in a directory tree — but it expects to find both files together. It took manually adding all of the Qt headers as _source_ files of the library, in the CMake configs, to get `AUTOMOC` working with the current organization. (That setup _works_ fine, but it was an extra PITA that could've been avoided.)
1. On a personal note, I find it particularly hard to edit a split directory structure like we use. It's pretty rare that editing a particular `.cpp` file doesn't send me off to open its header file as well, or vice versa. Most editors don't make that very convenient — not nearly as convenient as having the files side-by-side would be, anyway.

#### Proposals
So, given all that, I'm actually considering a PR that would go one of two very _different_ routes. And I'll say up front, I vastly prefer the first option. But if it's objectionable for some reason, the second would be my compromise proposal. I'm advocating that we do **one of** the following...
[...]

PR #504, opened 2020-04-19
MAJOR CHANGE (but not really): Move headers to src/ directory

ferdnyc:
This implements the proposal I made in #471, that header files live in the `src/` directory alongside their `*.cpp` files, rather than being kept in an `include/` directory.

- Move all `/include/**/*.h` to `/src/...` (with same hierarchy)
- Move `/src/examples` to `/examples`
- Move `/src/Qt/demo` to `/examples/qt-demo`
- Move `/src/bindings/` to `/bindings`
- Adjust source files (far fewer relative includes):
  - `/src/**/*.cpp`
  - `/src/**/*.h`
  - `examples/**/*.cpp`
  - `tests/**/*.cpp`
- Change `#include "JuceHeader.h"` to `#include <JuceHeader.h>` everywhere, since that header isn't part of our local source for the purposes of a libopenshot build
- Adjust all `CMakeLists.txt` files to account for new include path
- Add an `examples/CMakeLists.txt`, and move the logic for compiling and installing the necessary components to that file
- Add logic to `tests/CMakeLists`, `examples/CMakeLists`, `/CMakeLists.txt` so that build components are imported locally when needed
- Update `Doxygen.in` file to build docs from new locations

This will be confusing, at first, when looking for moved files. (I've confused myself a couple of times already.) The relocated `bindings` directory is the one that throws me the most. But it really is _much_ simpler this way. Both in terms of humans editing the code, _and_ in terms of the contents of the code.

The reason `src/examples`, `src/Qt/demo`, and `src/bindings` got moved out of `src/` is in small part practical: If there are directories in `src/` without any installable headers, the `install(DIRECTORY...)` CMake command will end up creating empty directories in the output location. So, I wanted to avoid that.

But at the same time, the fact that that happens makes the point: _none_ of those directories' contents are part of the _libopenshot_ source code itself. So they really don't _belong_ in the **libopenshot** source dir. It makes more sense if they're separated.

The Doxygen docs mostly come out better, now. It never could follow the parent-relative `#includes`, so in the past _none_ of the source code listings' `#include`s linked to the header in question. Now almost all of them do, except for a few `#include "../foo.h"` statements in the `src/effects/` or `src/Qt` paths. (And in theory I could eliminate those using the include path, since `#include "ReaderBase.h"` should mean the same thing in `src/effects/Blur.h` that it means anywhere else. But I haven't tackled that for this PR, so it [still] uses `#include "../ReaderBase.h"`.)

I've tested this in CMake 3.2 and CMake 3.17 and everything worked, but I've only tested it on Linux. This branch is in the project repo, so the builders can chew on it in addition to Travis. (And if we merge #501, also AppVeyor.)


ferdnyc:
Oh, I forgot to mention: This _shouldn't_ affect anyone's builds or use of the library, since the installed output is exactly the same. It just takes a different path getting there. But it will affect developers, obviously, including anyone running OpenShot out of a build dir with `PYTHONPATH=...build/src/bindings/python` (that would now become `PYTHONPATH=...build/bindings/python`).

jonoomph:
@ferdnyc This one seems scary to me... and lots of conflicts. I might wait on this for now.

ferdnyc:
> @ferdnyc This one seems scary to me... and lots of conflicts. I might wait on this for now. I think this is more of an emotional response, vs a technical one. Technically, nothing is wrong with this at all.

@jonoomph 

Honestly, the only way to do something like this is to set it up and then commit it immediately, or pretty close. It touches **every** header, so conflicts are inevitable and they pile up fast... but it's mostly just a few quick `git mv` commands to actually set the changes up on a current branch. The file content edits are pretty much just a search-and-replace of all the `#include` lines, and then some `CMakeLists.txt` tweaks; they're way easier to redo than to try and reconcile with other changes.

If you're down with the _concept_, I'll re-create this as a new PR and we can get it merged quick before the conflicts pile up.

Issue #506, opened 2020-04-24
Pixelization = 1,0 does not pixelate at all


mkarg:
**Describe the bug:**
The "Pixelization" values is not linear. It starts with 0,0 which means "no pixelization at all", through 0,9" which means "heavy pixelization", but ends with "1,0" which **does not pixelate at all** (i. e. 0,0 and 1,0 behave *the same*). This makes it hard to understand for the user, and hard to prevent failures in usage.

**Steps to reproduce the behavior:**
1. Use Pixelation effect
2. Set Pixelization = 1,0

**Expected behavior:**
Strongest possible pixelization

**System Details:**
- OpenShot Version 2.5.1-dev2
- Operating System Windows 10

**Log Files:**
N/A

**Exception / Stacktrace:**
N/A

**Screenshots:** (Optional)
N/A

ferdnyc:
Not hard to see why this happens, looking at the code.
https://github.com/OpenShot/libopenshot/blob/15c3efb76be2a734bd3b73b6122048de85397959/src/effects/Pixelate.cpp#L70-L86

Guess that should've been
```c++
1.01 - std::min(fabs(pixelization.GetValue(frame_number)), 1.0);
```

ferdnyc:
> The "Pixelization" values is not linear.

_That's_ a separate issue, and one that could really use correcting as well. The same thing was true in the old code, the effect really doesn't even **begin** to do anything visibly until you get to like 0.75.

The problem is actually that it's TOO linear (1.0 excepted) — it could really do with a curve function rather than the current constant-slope adjustment, since it defines what's essentially a size multiplier. Reducing the size of something by even 50-75% and then blowing it back up again doesn't result in what most people would really consider to be pixelization, it just makes your image look slightly crappier.

PR #507, opened 2020-04-27
Improve Pixelate filter scale

SuslikV:
Fixes: https://github.com/OpenShot/libopenshot/issues/506

### DIFF
--- a/src/effects/Pixelate.cpp
+++ b/src/effects/Pixelate.cpp
@@ -33,7 +33,7 @@
 using namespace openshot;
 
 /// Blank constructor, useful when using Json to load the effect properties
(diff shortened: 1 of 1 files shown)


ferdnyc:
This looks pretty good to me, though I haven't tested it very thoroughly yet. 

Sorry, I got bogged down when I started testing with the discovery that I screwed up the implementation of the effect borders, which seem to be pretty much broken on two sides. (Top, for sure.)

I'm having a bear of a time getting it to work right, `QRect` and `QPainter` seem to be  having these weird, nondeterministic interactions. Decided to leave it alone for now and come back to it tomorrow.

Issue #511, opened 2020-05-15
gcc-10: multiple definition of `AV_GET_CODEC_CONTEXT'


steils:
See the full log: 
[build.log](https://github.com/OpenShot/libopenshot/files/4637447/build.log)

It does not happen with gcc-9.
See also: https://gcc.gnu.org/gcc-10/porting_to.html (Default to -fno-common).

So there are multiple definitions with `LIBAVFORMAT_VERSION_MAJOR >= 58` or `defined(IS_FFMPEG_3_2)`.

ferdnyc:
Hmm. Is this a new change in GCC 10.1? Because [it doesn't happen with GCC 10.0.1](http://koji.rpmfusion.org/kojifiles/packages/libopenshot/0.2.5/1.fc33/data/logs/x86_64/build.log).

I'm wondering if this is really a GCC 10 issue, or if it's caused by building with Ninja instead of Makefiles. (Which we don't officially support.)

steils:
I'm not sure if `-fno-common` was default in 10.0.1, I have only 9.3.0 and 10.1.0 installed. But this isssue is definitely not related to ninja/make build system. If I try to build it with make (without the patch) I see exaclty the same error.

ferdnyc:
All that being said, libopenshot builds fine _with_ this change too, so if it works the same in all cases and doesn't cause any issues, I'm fine with it.

While there _shouldn't_ be multiple definitions of `AV_GET_CODEC_CONTEXT` **within** the file (all of those blocks are `elif`'d, so only one of them will ever be used), _technically_ it is a variable declaration, so I suppose each compiled object could end up with its own `AV_GET_CODEC_CONTEXT`. Probably wouldn't be an issue if the header were defining a namespaced C++ class, but since it's all wrapped in `extern "C"` everything lands in the global namespace.

...All _that_ being said, this still smells like an out-of-date builddir problem to me, and I have a feeling doing a `make clean` (or completely deleting and regenerating the build environment) before running a fresh build might fix it. So far I haven't been able to reproduce.

But like I said, I'm not really opposed to the change even if I can't reproduce it.

ferdnyc:
(I did try adding `-fno-common` to the `g++` command line with GCC 9.3.1, no change. I don't (yet) have a local system with GCC 10 installed that I can easily test on, my laptop is just about to start its Fedora 32 upgrade run. So in a couple of hours or so, I'll at least have GCC 10.0.1 to play with.)

[...]

PR #512, opened 2020-05-15
FFmpegUtilities: replace variable definition with statement expression

steils:
It is needed to avoid multiple definitions of AV_GET_CODEC_CONTEXT,
which is considered as an error with '-fno-common' which is default
since gcc-10.

Fixes: #511


ferdnyc:
Huh, shows what I know. Confirmed with GCC 10.1, though it's not an issue with 10.0 weirdly enough.

Still, the fix seems fine, all unit tests still pass, and OpenShot works as expected.

And since Travis already tested Clang builds for us, I'd say this is ready to merge. Thanks!

PR #515, opened 2020-05-19
Enable building with a libopenshot-audio that's based on non-bundled JUCE

ferdnyc:
This is the companion PR to OpenShot/libopenshot-audio#97, and contains the fixes necessary in libopenshot's build to ensure that it works with a libopenshot-audio that wasn't built from our bundled JUCE sources.

(See that PR, as well as [Debian bug 925755](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=925755), for background.)

In addition:
* Added some overlooked `juce::` prefixing, in places the unit tests don't touch (so it's not technically critical)
* Replaced our one use of `juce::ScopedPointer` (which is [deprecated](https://juce.com/discover/stories/coding-standards#object-lifetime-and-ownership)) with the recommended `std::unique_ptr` replacement.

### DIFF
--- a/include/Qt/AudioPlaybackThread.h
+++ b/include/Qt/AudioPlaybackThread.h
@@ -75,7 +75,7 @@ namespace openshot
 		static AudioDeviceManagerSingleton * Instance();
 
 		/// Public device manager property
-		AudioDeviceManager audioDeviceManager;
+		juce::AudioDeviceManager audioDeviceManager;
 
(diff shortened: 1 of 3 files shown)


ferdnyc:
Merging this, as it's fine with the current libopenshot-audio source, but the build will start breaking if OpenShot/libopenshot-audio#97 is merged before it.

PR #516, opened 2020-05-19
Fix support for older CMake

ferdnyc:
I used `add_compile_definitions()` in #515, which is CMake 3.12+ only. (Oops.) Switch to `add_definitions()` which has been supported since the dawn of time.


ferdnyc:
Travis seems to be permanently out to lunch here, I think perhaps because I submitted this PR using the same branch as the one I'd **just** merged. Closing and trying with a new branch.

PR #517, opened 2020-05-19
Fix support for older CMake

ferdnyc:
I used `add_compile_definitions()` in #515, which is CMake 3.12+ only. (Oops.) Switch to `add_definitions()` which has been supported since the dawn of time.


ferdnyc:
\*sigh\* NOPE. It seems I broke Travis again, somehow. :angry:

ferdnyc:
@jonoomph I don't know what happened here, but if you're able to either unfuck Travis or forcibly merge this, that'd be a help. I'm guessing at least one of the Gitlab builders is currently breaking on the CMake 3.12 command I accidentally merged into the repo (and now can't fix).

ferdnyc:
@jonoomph This change, BTW, was made for the Debian folks, in hopes of getting a newer version of OpenShot finally packaged into their repos. See #515, OpenShot/libopenshot-audio#97, and most of all the Debian bug I linked to from those, for all the gory details.

PR #519, opened 2020-05-29
Don't link Python module with libpython

ferdnyc:
Seems we're not meant to be linking with the Python library (**Edit:** on Linux), when
building a module. (See [PEP 513].) Not doing so allows the module to be used with builds of Python other than the library version.

[PEP 513]: https://www.python.org/dev/peps/pep-0513/#id41


ferdnyc:
Withdrawing this, as I've come to question this advice.

In building the module, I've found that if I were to e.g. compile with the headers from `/usr/include/python3.9` but _not_ link to `libpython3.9`, then attempting to load the module into `python3.8` fails due to missing symbols. If I _do_ link with `libpython3.9`, though, then the module is usable in Python 3.8 despite the mismatching library versions.

I'm not sure what the Python developers _expect_ to happen in that scenario (and perhaps should take it up with them), but for now keeping the module linking that's been working for us so far seems like the safest course of action.

PR #520, opened 2020-06-01
Fix FPS setting on FFmpeg 4

jonoomph:
Fix FPS setting on FFmpeg 4, which currently is not setting a valid FPS. This causes all sorts of strange behavior on videos produced with libopenshot using FFmpeg 4. This patch does require a depreciated approach, and I'm not sure how we are supposed to be correctly setting the FPS parameter, so I'm open to suggestions.

Also, thanks to PeterM for troubleshooting this issue and finding a solution! Nice work!

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -1223,11 +1223,11 @@ AVStream *FFmpegWriter::add_video_stream() {
 	st->time_base.num = info.video_timebase.num;
 	st->time_base.den = info.video_timebase.den;
 #if (LIBAVFORMAT_VERSION_MAJOR >= 58)
-	_Pragma ("GCC diagnostic push");
-	_Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\""); \
+	#pragma GCC diagnostic push
+	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 	st->codec->time_base.num = info.video_timebase.num;
 	st->codec->time_base.den = info.video_timebase.den;
-	_Pragma ("GCC diagnostic pop");
+	#pragma GCC diagnostic pop
 #endif
 
 	c->gop_size = 12; /* TODO: add this to "info"... emit one intra frame every twelve frames at most */
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -1222,6 +1222,13 @@ AVStream *FFmpegWriter::add_video_stream() {
 	st->avg_frame_rate = av_inv_q(c->time_base);
 	st->time_base.num = info.video_timebase.num;
 	st->time_base.den = info.video_timebase.den;
+#if (LIBAVFORMAT_VERSION_MAJOR >= 58)
+	_Pragma ("GCC diagnostic push");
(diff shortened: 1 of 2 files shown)


jonoomph:
@ferdnyc Travis is having trouble with finding our FFmpeg PPA: https://launchpad.net/~jon-hedgerows/+archive/ubuntu/ffmpeg-backports. This all of a sudden does not seem to exist anymore. Ughhh.... :wink:

ferdnyc:
Oy, not again. I'm beginning to understand why some people say they won't ever use community PPAs in production builds.

ferdnyc:
I'll just set the 3.2 build as allowed-to-fail, for now, like I did with FFmpeg 4 during the "walkout". But it looks like that PPA was made private, I doubt it's coming back. And there was no Ubuntu LTS release that included FFmpeg 3.2, it got skipped over. We may just have to drop that version from our build matrix.

jonoomph:
I suppose we could always setup our own PPA builds, but that sounds like a huge amount of time to get things all automated and building constantly, and then we'll have the same problem that Jon has, a ton of other projects will see our PPA and start incorporating it into their production builds.

ferdnyc:
Fixed in your branch, and I'll open a PR for develop.

jonoomph:
> I'll just set the 3.2 build as allowed-to-fail, for now, like I did with FFmpeg 4 during the "walkout". But it looks like that PPA was made private, I doubt it's coming back. And there was no Ubuntu LTS release that included FFmpeg 3.2, it got skipped over. We may just have to drop that version from our build matrix.

Thanks, that makes sense!

ferdnyc:
OK, let's try that again.

ferdnyc:
There we go, now I see it's marked "allowed to fail'. I tried to mask it by environment variable like I did last time, but that didn't work. I think to do that I'd have to mask the **entire** environment config, and we have three variables now. So I masked it by name insead.

src/FFmpegWriter.cpp
```diff
@@ -1222,6 +1222,13 @@ AVStream *FFmpegWriter::add_video_stream() {
 	st->avg_frame_rate = av_inv_q(c->time_base);
 	st->time_base.num = info.video_timebase.num;
 	st->time_base.den = info.video_timebase.den;
+#if (LIBAVFORMAT_VERSION_MAJOR >= 58)
+	_Pragma ("GCC diagnostic push");
+	_Pragma ("GCC diagnostic ignored \"-Wdeprecated-declarations\""); \
+	st->codec->time_base.num = info.video_timebase.num;
+	st->codec->time_base.den = info.video_timebase.den;
+	_Pragma ("GCC diagnostic pop");
```
ferdnyc, line 1230:
```suggestion
	#pragma GCC diagnostic push
	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
	st->codec->time_base.num = info.video_timebase.num;
	st->codec->time_base.den = info.video_timebase.den;
	#pragma GCC diagnostic pop
```

The only reason we had to use the `_Pragma()` form in FFmpegUtilities is that you can't use preprocessor-macro `#pragma` _in_ a preprocessor macro. But it works fine here.

PR #521, opened 2020-06-01
Travis: allow_failures for FFmpeg 3.2

ferdnyc:
Temporarily set the FFmpeg 3.2 build as "allowed to fail", while we work out what to do about its suddenly-private PPA.


ferdnyc:
I also removed the (as it turns out) unused Qt PPAs from the configs, since they were pointless.

PR #435, opened 2020-02-10
Travis: Add FFmpeg 3.2 build

ferdnyc:
This adds an additional build type to the Travis matrix, building against FFmpeg 3.2 specifically (using a PPA which packages FFmpeg 3.2.6 for Ubuntu Xenial) — the existing FFmpeg 3 builds are relabeled "FFmpeg 3.4" to differentiate them.

PR #522, opened 2020-06-01
Travis: Switch to Ubuntu 20.04 Focal for FFmpeg 4 build

ferdnyc:
Travis appear to have quietly brought up Ubuntu Focal as an available build `dist` (still undocumented anywhere, AFAICT). Still, using an undocumented Travis environment is the lesser evil, vs. using FFmpeg packages from a community PPA that's already bailed on us once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,19 +57,17 @@ jobs:
           - lcov
           - binutils-common # For c++filt
 
-    - name: "FFmpeg 4 GCC (Ubuntu 18.04 Bionic)"
+    - name: "FFmpeg 4 GCC (Ubuntu 20.04 Focal)"
       env:
         - BUILD_VERSION=ffmpeg4
         - CMAKE_EXTRA_ARGS=""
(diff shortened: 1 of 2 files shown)


ferdnyc:
Merging this, as it allows us to eliminate _all_ community/unofficial PPA usage (except our own!) in the CI setup, other than the FFmpeg 3.2 problem-child which remains in limbo ATM.

Travis's Focal support isn't likely to go away now that they've let that genie out of its bottle, I'm sure it'll become official soon enough.

cc: @jonoomph

ferdnyc:
FFmpeg 3.2 will probably just end up getting dropped. Adding it as a test version in #435 was a "nice to have", after #433 revealed some issues with code that was _claimed_ to be FFmpeg 3.2 compatible but actually required 3.4... but it was never critical. Those are **both** outdated versions not worth expending resources on.

When adding 3.2 was simple, it made sense. No longer simple? No longer makes sense.

Issue #447, opened 2020-02-22
Frame class and it's API


musteresel:
I'm currently struggling a bit with `Frame`'s API.

https://openshot.org/files/libopenshot/classopenshot_1_1Frame.html

(I'll try to summarize the issues I have, but it may take a few edits over the next hours / days even :) )

We have constructors:
 - a default constructor
 - an image only constructor
 - an audio only constructor
 - image and audio construtor

We have public data members:
 - two flags for whether there's audio / image data
 - the frame number

There are member functions to "Add" to the frame:
 - AddAudio
 - AddAudioSilence
 - AddColor
 - AddImage
 - AddMagickImage

There are member functions to "query" (**sometimes not only** ...) stuff of the frame:
- ChannelsLayout
- CheckPixel
- GetAudioChannelsCount
- GetAudioSample
- GetAudioSampleBuffer
- GetAudioSamples
- GetAudioSamplesCount
- GetBytes
- GetWidth
- GetHeight
- GetPixels
- GetPixelRatio
- GetImage
- GetMagickImage
- GetPlanarAudioSamples
- GetInterleavedAudioSamples
- GetSamplesPerFrame
- GetWaveform
- GetWaveformPixels
- SampleRate

Then there are a few "modifying" member functions:
- SampleRate
- SetFrameNumber
- SetPixelRatio
- ResizeAudio
- ChannelsLayout
- ApplyGainRamp

Finally there are member functions which "do" things, e.g. for debugging:
- ClearWaveform
- DeepCopy
- Display
- DisplayWaveform
- Play
- Thumbnail

I hope I didn't forget a member function.

All in all this is ... *a lot* .. for a "container" of audio and image data.  Furthermore, some of these functions do **more** than they're advertising, e.g. `GetPixels()` creating image data if there's none, whereas `GetPixels(0) /* a row */` just crashes when there's no image data.

Which of these functions are actually used? Or rather: Where should I look for the "main" user of frames?  Which (undocumented) side effects are actually used? Is `Frame` used directly from OpenShot?

I think `Frame` would greatly benefit from a bit cleanup work :)

Related: https://github.com/OpenShot/libopenshot/pull/446

ferdnyc:
A few things I would characterize differently:

* DeepCopy is used by the assignment operator (and I think by the copy constructor? I forget, do we have a copy constructor?), so it's a (necessary) utility method, in my view, not a debugging tool. It _could_ be private, perhaps, arguably.
[...]

PR #527, opened 2020-06-06
Adding some new functionality and documentation to DummyReader

jonoomph:
Adding the ability to add test frames `WriteFrame()`, with fake image and audio data. This will can be used in unit-tests, and will soon be used to verify some new audio improvements (coming soon).

### DIFF
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -480,6 +480,11 @@ const unsigned char* Frame::GetPixels()
 // Get pixel data (for only a single scan-line)
 const unsigned char* Frame::GetPixels(int row)
 {
+	// Check for blank image
+	if (!image)
+		// Fill with black
+		AddColor(width, height, color);
+
 	// Return array of pixel packets
 	return image->constScanLine(row);
 }
--- a/tests/DummyReader_Tests.cpp
+++ b/tests/DummyReader_Tests.cpp
@@ -1,0 +1,102 @@
+/**
+ * @file
+ * @brief Unit tests for openshot::DummyReader
+ * @author Jonathan Thomas <jonathan@openshot.org>
+ *
+ * @ref License
+ */
+
+/* LICENSE
+ *
+ * Copyright (c) 2008-2019 OpenShot Studios, LLC
+ * <http://www.openshotstudios.com/>. This file is part of
+ * OpenShot Library (libopenshot), an open-source project dedicated to
+ * delivering high quality video editing and animation solutions to the
+ * world. For more information visit <http://www.openshot.org/>.
+ *
+ * OpenShot Library (libopenshot) is free software: you can redistribute it
+ * and/or modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * OpenShot Library (libopenshot) is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with OpenShot Library. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "UnitTest++.h"
+// Prevent name clashes with juce::UnitTest
+#define DONT_SET_USING_JUCE_NAMESPACE 1
+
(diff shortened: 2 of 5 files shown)


jonoomph:
@musteresel Please take a look and let me know what you think.

src/Frame.cpp
```diff
@@ -480,6 +480,11 @@ const unsigned char* Frame::GetPixels()
 // Get pixel data (for only a single scan-line)
 const unsigned char* Frame::GetPixels(int row)
 {
+	// Check for blank image
```
jonoomph, line 483: I know... more magic stuff happening in the Frame class :grimacing:

tests/DummyReader_Tests.cpp
```diff
@@ -0,0 +1,102 @@
...
+	CHECK_EQUAL(1, r.GetFrame(1)->GetPixels(700)[700] == 0); // black pixel
+	CHECK_EQUAL(1, r.GetFrame(1)->GetPixels(701)[701] == 0); // black pixel
+}
+
+TEST (DummyReader_Fake_Frame) {
+	// Create a default fraction (should be 1/1)
+	openshot::DummyReader r(openshot::Fraction(30, 1), 1920, 1080, 44100, 2, 30.0);
+	r.Open(); // Open the reader
+
+	// Let's create some test frames
```
jonoomph, line 72: Creating some test Frame objects, with custom audio data (it doesn't really need to make sense), and then recalling those frames from DummyReader, and verifying our custom audio data is still intact.

jonoomph:
This PR is loosely related to https://github.com/OpenShot/libopenshot/issues/447, and is related to some timeline audio merging improvements being worked on by @musteresel.

ferdnyc:
> Adding the ability to add test frames `WriteFrame()`, with fake image and audio data.

I have to say, I _really_ **hate** that idea. Adding a write-API to the dummy _reader_? That's... I don't see the point of unit-testing something if it has to be fitted with special APIs that make no sense in the rest of the codebase. What is that even testing?

Let me propose a slightly less invasive approach. Instead of adding `WriteFrame()` to the DummyReader just so the tests can populate its cache, why not just give it a constructor that takes a `CacheMemory` object? The unit tests can create the cache, call `testcache.Add(frame)` to their heart's content, and then call `new openshot::DummyReader(testcache)` to wrap a Reader around the completed cache. That at least makes it work roughly the same way any other reader works, where it gets wrapped around some data (whether it's a file path, HTML source, etc...) in its constructor.

That way, the "special" API is limited to a constructor overload, and the rest of it is just a normal reader.

jonoomph:
@ferdnyc :+1: I like that as well. Let me tweak things real quick.

src/Frame.cpp
ferdnyc: Not just magic stuff, _wrong_ stuff. Getters shouldn't be creating data! Heck, that should be a `const` method.

If there's no image in the Frame and something requests pixel data, isn't the right thing to either throw an exception or return `nullptr`? Why does the class do so much work to make it **impossible** to keep image data out of it? I mean, heck, why is there even a `has_image_data` info member, when it's never checked, and it's virtually **impossible** for it to remain `false` even if you TRY?

I feel like all this magical image-creation is trying to make the API more "friendly" by loading it with assumptions about what the caller wants, which is only even maybe-right when they're **correct** assumptions. In fact, even the idea that auto-creating data out of thin air makes the API more friendly is _itself_ an assumption.

If nothing else the comment is wrong. The image will be filled with whatever the value of `color` is, not necessarily black. (And `color` will be set to that argument, at the same time. Which makes me really nervous, now that I look at it.)

jonoomph:
We are going to be generating Frame objects with artificial audio data, such as an always incrementing number, and then build unittests which pass these frames through the Timeline with different FPS and sample rates. The idea is to verify audio data being truncated in the Timeline when combining Frames of different amounts of audio samples, and then correct the behavior once we can verify the failures. Hope that makes sense, haha.

ferdnyc:
Like I said in the inline comments, this part suddenly makes me nervous:

https://github.com/OpenShot/libopenshot/blob/7831cfe9128502ae894d85c14e033991cd9a3d1b/src/Frame.cpp#L737-L756

If calling `GetPixelData()` can trigger an `AddColor()`, then theoretically it could be called from two threads at the same time. There's no reason to mistrust a getter that returns `const`, normally. But that `color = new_color;` bit is outside all of the thread synchronization. Yeah, it's going to be the _same_ string, but because we pass everything by value the code doesn't actually KNOW that, and it's going to copy it over again anyway, potentially colliding on the writes.

PR #530, opened 2020-06-08
Move off deprecated OpenMP API, fix FFmpeg code paths

ferdnyc:
This was primarily inspired by Codecov flagging some code as unreachable, turned out it has a bit of history. But this corrects that, moves us off of a deprecated OpenMP API call, and makes some tweaks to FFmpegWriter/Reader along with Timeline, while I was touching them.

I'll explain some of the changes in inline comments.

### DIFF
--- a/include/FFmpegReader.h
+++ b/include/FFmpegReader.h
@@ -233,14 +233,13 @@ namespace openshot {
 		/// codecs have trouble seeking, and can introduce artifacts or blank images into the video.
 		bool enable_seek;
 
-		/// Constructor for FFmpegReader.  This automatically opens the media file and loads
-		/// frame 1, or it throws one of the following exceptions.
-		FFmpegReader(std::string path);
-
-		/// Constructor for FFmpegReader.  This only opens the media file to inspect its properties
-		/// if inspect_reader=true. When not inspecting the media file, it's much faster, and useful
-		/// when you are inflating the object using JSON after instantiating it.
-		FFmpegReader(std::string path, bool inspect_reader);
+		/// @brief Constructor for FFmpegReader.
+		///
+		/// Sets (and possibly opens) the media file path,
+		/// or throws an exception.
+		/// @param path  The filesystem location to load
+		/// @param inspect_reader  if true (the default), automatically open the media file and loads frame 1.
+		FFmpegReader(const std::string& path, bool inspect_reader=true);
 
 		/// Destructor
 		virtual ~FFmpegReader();
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -83,7 +83,7 @@ int hw_de_on = 0;
 	AVHWDeviceType hw_de_av_device_type_global = AV_HWDEVICE_TYPE_NONE;
 #endif
 
-FFmpegReader::FFmpegReader(std::string path)
+FFmpegReader::FFmpegReader(const std::string& path, bool inspect_reader)
 		: last_frame(0), is_seeking(0), seeking_pts(0), seeking_frame(0), seek_count(0),
 		  audio_pts_offset(99999), video_pts_offset(99999), path(path), is_video_seek(true), check_interlace(false),
 		  check_fps(false), enable_seek(true), is_open(false), seek_audio_frame_found(0), seek_video_frame_found(0),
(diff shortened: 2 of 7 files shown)


src/FFmpegWriter.cpp
```diff
@@ -691,15 +697,8 @@ void FFmpegWriter::WriteFrame(std::shared_ptr<Frame> frame) {
...
-			// Write frames to video file
-			write_queued_frames();
-
-		else {
-			// Write frames to video file
-			write_queued_frames();
-		}
+		// Write frames to video file
+		write_queued_frames();
 	}
```
ferdnyc, line 702: This was the bit Codecov flagged as unreachable. I should certainly say so! Turns out this block has a bit of history.

Once upon a time it looked basically exactly as it will after my change. Then, 8 years ago in 0f3758bc29d2470d0cf2fe589e89b21330aee0d9, it became this:
```c++
	if (spooled_video_frames.size() == cache_size || spooled_audio_frames.size() == cache_size)
	{
		// Is writer currently writing?
		if (!is_writing)
			// Write frames to video file
			write_queued_frames();

		else
		{
			// YES, WRITING... so wait until it finishes, before writing again
			while (is_writing)
				usleep(250 * 1000); // sleep for 250 milliseconds

			// Write frames to video file
			write_queued_frames();
		}

	}
```

It trundled along like that with only minor changes until 851ad86634c01a2c2c05bdffc3805466b3354c91, when (despite the commit message being "Changing some sleep() calls to usleep()", the sleep was _removed_ entirely, making the two branches accidentally equivalent.

Maybe the right change is to put the sleep call _back_, I don't know. But in its current state the code is nonsensical, the second branch can never be reached.

ferdnyc: (I might also point out that even _if_ we want to put the sleep-before-write back in, the cleanest form would be...)
```c++
if (spooled_video_frames.size() == cache_size || spooled_audio_frames.size() == cache_size) {
    // Wait until an in-progress write finishes
    while (is_writing) {
        std::this_thread::sleep_for(std::chrono::milliseconds(2500));
    }
    write_queued_frames();
}
```
There's no reason for the preceding `if()`, when `is_writing` is false execution will fall straight through the `while()` without a single iteration.

include/FFmpegReader.h
```diff
@@ -233,14 +233,13 @@ namespace openshot {
...
-		/// if inspect_reader=true. When not inspecting the media file, it's much faster, and useful
-		/// when you are inflating the object using JSON after instantiating it.
-		FFmpegReader(std::string path, bool inspect_reader);
+		/// @brief Constructor for FFmpegReader.
+		///
+		/// Sets (and possibly opens) the media file path,
+		/// or throws an exception.
+		/// @param path  The filesystem location to load
+		/// @param inspect_reader  if true (the default), automatically open the media file and loads frame 1.
+		FFmpegReader(const std::string& path, bool inspect_reader=true);
```
ferdnyc, line 242: Previously the two constructors for FFmpegReader differed only in that one required an `inspect_reader` argument, which turned off the `Open(); Close();` cycle that was always performed by the one-arg version. (IOW, there was no reason to use the 2-arg version unless `inspect_reader` was false.) By making it one constructor with an optional `inspect_reader` argument that defaults to `true`, I was able to eliminate an entire constructor's worth of duplicate code.

While I was touching these constructors, I changed their string args to take a const reference, so that we can try it out as a change. Especially for arguments that are simply copied to member variables in the initializer list, there should be no down side, only a slightly more efficient method that avoids one unnecessary redundant copy of a class that's just going to be copied **again** by the initializer list. I anticipate submitting a much larger PR that updates a significant porition of our class methods in the same manner, so consider this a trial run.

ferdnyc: The documentation for a bunch of these constructors also kept ending on, "...throws one of the following exceptions." with no list of exceptions after it. There's nothing worse than documentation that teases you and doesn't follow through, so I changed it to not over-promise.

src/FFmpegReader.cpp
```diff
@@ -898,11 +882,6 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
 	int minimum_packets = OPEN_MP_NUM_PROCESSORS;
 	int max_packets = 4096;
 
-	// Set the number of threads in OpenMP
-	omp_set_num_threads(OPEN_MP_NUM_PROCESSORS);
-	// Allow nested OpenMP sections
-	omp_set_nested(true);
```
ferdnyc, line 904: Last but not least, I noticed there were a few places scattered around the code that kept making these calls. They were always in very ad-hoc locations, generally right before a nested `#pragma` section was actually used. But that meant (a) they weren't necessarily in effect _before_ that point, and (b) they'd keep getting called every time that block of code was run.

I have no idea how expensive `omp_set_num_threads()` and `omp_set_nested()` are, but they can't be _FREE_ and there's no reason to call them more than once. So, I moved this initialization into the class constructors where it can get done once and never needs to be fiddled wih again.

There's also the fact that `omp_set_nested()` was deprecated in OpenMP 4.5 due to its lack of granularity. OpenMP has always (since 1.0) supported `omp_set_max_active_levels()`, which allows control over the _amount_ of nesting possible in the code, rather than just offering a boolean switch. So the code I moved to the constructors uses that, instead.

jonoomph:
Good catch! I think this implementation is fine for now. I'll do some local testing as well.

Issue #532, opened 2020-06-12
DummyReader fails when given an almost empty cache


musteresel:
~~~~c++
CacheMemory cache;
shared_ptr<Frame> first_frame(new Frame(1, 3, 1));
// populate frame
cache.Add(first_frame);
shared_ptr<Frame> second_frame(new Frame(2, 4, 1));
// populate frame
cache.Add(second_frame);
DummyReader reader(Fraction(30, 1), 640, 480, 3, 1, 10, &cache);
Clip clip(&reader);
//clip.Open();
//clip.GetFrame(3);
Timeline timeline(640, 480, Fraction(30, 1), 3, 1, LAYOUT_MONO);
timeline.AddClip(&clip);
timeline.Open();
timeline.GetFrame(1);
~~~~

The "expected" behavior here IMO would be to get the `first_frame`.  However I get an exception *"Requested frame not found. You can only access Frame numbers that exist in the Cache object."*

This is due to buffering / pre-caching behavior of `Timeline`, which requests more frames by default to populate caches:

https://github.com/OpenShot/libopenshot/blob/c29174f18b10f01b5bc17ed6edd87dcc29d00274/src/Timeline.cpp#L887-L922

I'm not sure if this should be considered a bug ... or if we should just add a note to the DummyReader example?

*side note:* `clip.GetFrame(3)` will work (return an empty shared pointer of course) *unless* `clip.Open()` was called before. That's a bit confusing.

musteresel:
The exception comes from here: https://github.com/OpenShot/libopenshot/blob/c29174f18b10f01b5bc17ed6edd87dcc29d00274/src/DummyReader.cpp#L149

Is a reader allowed to return a "null" frame (empty shared pointer)?

PR #533, opened 2020-06-12
Add a note about the buffering behavior of Timeline to DummyReader

musteresel:
Using DummyReader with a custom cache with a Timeline can lead to
strange exceptions due to Timeline's buffering.  Therefore adding a
note to the DummyReader documentation about that.

See #532

### DIFF
--- a/include/DummyReader.h
+++ b/include/DummyReader.h
@@ -53,6 +53,12 @@ namespace openshot
 	 * A dummy reader can be created with any framerate or samplerate. This is useful in unit
 	 * tests that need to test different framerates or samplerates.
 	 *
+	 * @note Timeline does buffering by requesting more frames than it
+	 * strictly needs. Thus if you use this DummyReader with a custom
+	 * cache in a Timeline, make sure it has enough
+	 * frames. Specifically you need some frames after the last frame
+	 * you plan to access through the Timeline.
+	 *
 	 * @code
 	 * 	// Create cache object to store fake Frame objects
 	 * 	CacheMemory cache;


include/DummyReader.h
```diff
@@ -53,6 +53,12 @@ namespace openshot
 	 * A dummy reader can be created with any framerate or samplerate. This is useful in unit
 	 * tests that need to test different framerates or samplerates.
 	 *
+	 * @note Timeline does buffering by requesting more frames than it
+	 * strictly needs. Thus if you use this DummyReader with a custom
+	 * cache in a Timeline, make sure it has enough
+	 * frames. Specifically you need some frames after the last frame
+	 * you plan to access through the Timeline.
+	 *
```
ferdnyc, line 61: Makes sense. The exact number is runtime-dependent, basically it'll be OPEN_MP_NUM_PROCESSORS which is set to the smaller of:

* `num_procs()` (the actual CPU core count of the system)
* `openshot::Settings::Instance()->OMP_THREADS` (as long as that value is ≥ 2)

So, by setting `OMP_THREADS` the caller can limit the required readahead cache size to just 1 extra frame.

It might be good to add a more in-depth discussion of how all that works to the API docs for either the Settings or Timeline class, and link to that from here.

([openshot::Settings](http://openshot.org/files/libopenshot/classopenshot_1_1Settings.html#a24a29a1ca3dfaaf529f119b8fe0660c6) could really stand to be better documented in general. It doesn't help that a lot of the settings are only ever used in `#define` macros, which means they don't get indexed by Doxygen and _appear_ to be completely unreferenced in the code.)

But for the time being, this is definitely an improvement. Thanks!

PR #534, opened 2020-06-17
Set the field order when interlace is on

eisneinechse:
The field order of the interlaced frames is set. This is only a very rude version.
More work needs to be done


jonoomph:
LGTM

PR #539, opened 2020-07-03
Bitratecorrect

eisneinechse:
The values for qmax and qmin were set for all codec but only mpeg2 needed them and as a result the bitrates of the exported clips could be wrong. Now only when the export is with mpeg2 the default values are changed.


jonoomph:
LGTM. Not sure why codecov has an issue though.

PR #540, opened 2020-07-07
MacOS Catalina Support + _CXX11 + GCC 8.4

jonoomph:
Big changes to our Mac builder, including an upgrade to Catalina (10.15). This change was required due to new Gatekeeper changes on newer MacOS versions. Apple now requires all apps to be notarized, which involves many things, including a hardened runtime, more strict code signing, and entitlements. It also required a newer compiler than GCC 4.8 (our previous one), so I upgraded to GCC 8.4.

The libopenshot changes are mostly related to GCC 8.4 + Catalina changes to folder locations.
 - I had to manually include the XCode header files (since Apple removed /usr/include/, yup)
 - Updated CMake to use GCC 8.4
 - Disabled _GLIBCXX_USE_CXX11_ABI, due to breakage caused by symbol differences
 - Disabled find_package(Threads REQUIRED), since this breaks on MacOS now (and all OSes seem to be able to compile fine without it).
 - Related PR for openshot-qt changes: https://github.com/OpenShot/openshot-qt/pull/3600


src/CMakeLists.txt
```diff
@@ -303,8 +305,6 @@ endforeach()
 # Threading library -- uses IMPORTED target Threads::Threads (since CMake 3.1)
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 set(THREADS_PREFER_PTHREAD_FLAG TRUE)
-find_package(Threads REQUIRED)
```
jonoomph, line 306: @ferdnyc MacOS Catalina and GCC 8.4 did not like this at all. The only way I could get a passing build was to remove this line. However, it still appears that Linux and Windows build find with this missing. I still need to test those builds, and make sure they actually work, but the compilation passes. Any thoughts on this one?

.gitlab-ci.yml
```diff
@@ -43,7 +43,7 @@ mac-builder:
...
+    - cmake -DCMAKE_CXX_FLAGS=-I\ /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -D"CMAKE_INSTALL_PREFIX:PATH=$CI_PROJECT_DIR/build/install-x64" -DCMAKE_CXX_COMPILER=/usr/local/opt/gcc@8/bin/g++-8 -DCMAKE_C_COMPILER=/usr/local/opt/gcc@8/bin/gcc-8 -DCMAKE_PREFIX_PATH=/usr/local/qt5/5.5/clang_64 -DPYTHON_INCLUDE_DIR=/Library/Frameworks/Python.framework/Versions/3.6/include/python3.6m -DPYTHON_LIBRARY=/Library/Frameworks/Python.framework/Versions/3.6/lib/libpython3.6.dylib -DPYTHON_MODULE_PATH=python -DPython_FRAMEWORKS=/Library/Frameworks/Python.framework/ -D"CMAKE_BUILD_TYPE:STRING=Release" -D"CMAKE_OSX_SYSROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk" -D"CMAKE_OSX_DEPLOYMENT_TARGET=10.9" -D"CMAKE_INSTALL_RPATH_USE_LINK_PATH=1" -D"ENABLE_RUBY=0" ../
```
jonoomph, line 46: @ferdnyc I had to manually add `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include`, since Apple removed `/usr/include/`. What the heck? :confused:

ferdnyc: Hmm. Weird. Probably some stupid attempt at going all-in on Frameworks as their primary shared resource system, as opposed to "legacy" shared libraries / includes.

ferdnyc: What I find weird, though, is that CMake wouldn't take care of that automatically, if that's the case. We may need to update a config somewhere.

src/CMakeLists.txt
ferdnyc: @jonoomph Only just saw this, sorry. The threading library configs were only ever there for OpenMP, since apparently not linking with pthreads [could lead to link-time failures due to missing symbols](https://gitlab.inria.fr/openmp/libkomp/commit/cf0276b9ab4dba5d77a131fcf4061e5e2172d8eb). But I just ran a build with the current `CMakeLists.txt` and `/usr/lib64/libpthread.so` is right there next to `/usr/lib/gcc/x86_64-redhat-linux/10/libgomp.so` in the final linker command line, so clearly `OpenMP::OpenMP_CXX` is taking care of its dependencies now, the way it's supposed to.

In fact I just tried again after ripping out the variable-settings just above, and everything _still_ built exactly the way it was supposed to, so we can probably dump those as well just to avoid confusion. Might as well let OpenMP handle everything for itself.

PR #510, opened 2020-05-07
Add version/feature #defines to OpenShotVersion.h

ferdnyc:
The CMake process will pick up Qt's `Qt5Core_VERSION_STRING` variable and stick it in the CMake cache (with forced updating each run). A new `#define` in the header exports that as `QT_VERSION_STR` to both the library and the Python bindings. That way it can be queried at runtime, same as PyQt5's versioning:
```python3
>>> import openshot
>>> print(openshot.QT_VERSION_STR)
5.13.2
>>> from PyQt5 import Qt
>>> print(Qt.QT_VERSION_STR)
5.13.2
>>> print(Qt.PYQT_VERSION_STR)
5.13.2
```

**Edit:** I added a few more:
```python3
>>> openshot.AVUTIL_VERSION_STR
'56.31.100'
>>> openshot.AVCODEC_VERSION_STR
'58.54.100'
>>> openshot.AVFORMAT_VERSION_STR
'58.29.100'
>>> openshot.HAVE_IMAGEMAGICK
1
>>> openshot.HAVE_RESVG
1
```

**Edit2:** One more:
```python3
>>> openshot.APPIMAGE_BUILD
0
```
It's implemented as a CMake option, default `OFF` (hence the `0`). The Gitlab Linux runner will enable it (`cmake -DAPPIMAGE_BUILD=1`) when it generates the build environment.

I'll probably merge this soon, as it shouldn't affect any existing code and I'm going to have a need for that last one in an OpenShot PR shortly.

Inspired by https://github.com/OpenShot/openshot-qt/issues/3442#issuecomment-625206538

PR #542, opened 2020-07-08
Enable IN_LIST in older CMake

ferdnyc:
Should fix #510 breakage on gitlab builders.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -29,6 +29,11 @@ include(FeatureSummary)
 
 include(GNUInstallDirs)
 
+# Enable IN_LIST in older CMake
+if (POLICY CMP0057)
+  cmake_policy(SET CMP0057 NEW)
+endif()
+
 ################ WINDOWS ##################
 # Set some compiler options for Windows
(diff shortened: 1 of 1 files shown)


jonoomph:
Linux builder is happy now! Nice!

ferdnyc:
Yeah, I guess _technically_ this increases our minimum CMake version to 3.3 (which is where `IN_LIST` was introduced) — I could've also fixed this by changing the version range to `3.3...3.17` here:

https://github.com/OpenShot/libopenshot/blob/85b6a967254af5493f1130631da5aa99559675e3/CMakeLists.txt#L27

Which would've made `CMP0057` default to `NEW` behavior automatically. I guess in CMake 3.7 (version on the Linux builder) it was still defaulting to `OLD`, though they obviously dropped that at some point after.

Trying to build libopenshot with CMake 3.2 will likely now break when it hits the `IN_LIST`, but really who's going to try to do that?  `¯\_(ツ)_/¯`

jonoomph:
> ...but really who's going to try to do that? `¯\_(ツ)_/¯`

I guess we'll find out! haha

PR #545, opened 2020-07-12
CMake: Update Resvg find module

ferdnyc:
#### FindResvg

In addition, this PR also includes some (unrelated, I admit ­— apologies) enhancements to the Resvg find module, as I made them in the course of testing this:
1. I renamed the module to `FindResvg.cmake`, and the targets and variables it creates now start with `Resvg` instead of `RESVG`.
1. I added the ability to discover and publish _static_ library configs as well, in a separate target `Resvg::resvg_static`. The `Resvg::resvg` target is still the preferred and correct choice for linking with the `libresvg.so` shared library.
1. It now explicitly consumes `${Resvg_ROOT}` as the initial discovery path for the library, if that variable is set. CMake will automatically use the path set there, but only when building with version 3.12 or higher. Passing any paths set there in to the find commands explicitly means it'll work with any CMake version.

#### Withdrawn elements
<strike>
I find myself setting the same options over and over again every time I build the library — assuming I remember, of course.

There's my `CMAKE_INSTALL_PREFIX`, my preferred `CMAKE_BUILD_TYPE`, the `Resvg_ROOT` path to pick up that dependency from, etc. They're options that can't be incorporated into the project defaults because they're inherently _local_, but they're still things I'd rather not have to deal with every single time.

This PR adds configuration to take advantage of a CMake feature (only present in CMake 3.15+, I'm afraid) that will inject a source file into the beginning of the build, right before the `project()` command that kicks off the build environment generation. That file can be used to set up local defaults for the build which will be used each time, so they don't have to be passed on the command line.

If there is a file in the repository named `cmake/UserDefaults.cmake`, if the build is running under CMake 3.15, and if the option `IGNORE_DEFAULTS` has **not** been set on the command line, then CMake will source the `cmake/UserDefaults.cmake` file prior to running its own build scripts. Any settings made in that file will influence the subsequent build.

Setting things properly can be a bit tricky. Some things, like find module parameters, only need to be set as CMake variables:
```cmake
set(Package_ROOT /path/to/package/dir)
```

Other settings require more work to properly configure. In order to be effective both during and after the generation process, something like `CMAKE_INSTALL_PREFIX` has to be set as a cache variable, which requires the full signature:
```cmake
set(CMAKE_INSTALL_PREFIX "install-x64" CACHE PATH "Doc string")
```
But, with a little persistence it's possible to avoid the burden of specifying so much on the command line each time. 

The `cmake/UserDefaults.cmake` project is added to the checked-in `.gitignore` file, so there's no danger of anyone accidentally committing their local settings.

If anyone thinks this could be made better or more useful, I'm open to suggestions — including on the name. I thought about possibly calling it `cmake/LocalDefaults.cmake`, since that's perhaps closer to what it represents., but I went with "User" instead because I wanted to emphasize that the file is a place to store _personal_ build options and preferences, things that simply aren't appropriate to manage directly in the project's shared configuration.

OpenShot/libopenshot-audio#99 is my companion PR which adds the same functionality there.
</strike>


ferdnyc:
I'm reconsidering this. I like the idea in _theory_, but that trickiness I mention in the initial comment makes it difficult to use in practice, and I'm not sure it's really the right approach.

The FindResvg rewrite is still useful, though, so I guess I'll just pare this down to that part.

cmake/Modules/FindResvg.cmake
```diff
@@ -0,0 +1,142 @@
+# vim: ts=2 sw=2
+#[=======================================================================[.rst:
+FindResvg
+---------
+Try to find the shared-library build of resvg, the Rust SVG library
```
ferdnyc, line 5:
```suggestion
Try to find C-api library builds of Resvg, the Rust SVG library
```

cmake/Modules/FindResvg.cmake
```diff
@@ -0,0 +1,142 @@
...
+
+Control variables
+^^^^^^^^^^^^^^^^^
+
+The following variables can be used to provide path hints to the module:
+
+Resvg_ROOT    - (preferred) CMake 3.12+ standard path hint variable
+RESVGDIR      - Set in the calling CMakeLists.txt or on the command line
+ENV{RESVGDIR} - An environment variable in the cmake process context
+
```
ferdnyc, line 43:
```suggestion

With CMake 3.4+ (only), RESVGDIR can be a relative path
anchored at the PROJECT_SOURCE_DIR.

```

cmake/Modules/FindResvg.cmake
```diff
@@ -0,0 +1,142 @@
...
+Copyright (c) 2020, FeRD (Frank Dana) <ferdnyc@gmail.com>
+#]=======================================================================]
+include(FindPackageHandleStandardArgs)
+
+# CMake 3.4+ only: Convert relative paths to absolute
+if(DEFINED RESVGDIR AND CMAKE_VERSION VERSION_GREATER 3.4)
+  get_filename_component(RESVGDIR "${RESVGDIR}" ABSOLUTE
+    BASE_DIR ${CMAKE_CURRENT_BINARY_DIR})
```
ferdnyc, line 51:
```suggestion
    BASE_DIR ${PROJECT_SOURCE_DIR})
```

jonoomph:
Hi @ferdnyc. I'm a bit confused by this PR. You have a few suggestions you have made to yourself? Do we need to commit these suggestions before merging? Thx!

ferdnyc:
@jonoomph 

Heh. I ended up re-doing these changes alongside the header file relocation into the `src/` directory. (A new PR for which I'm about to open, as the branch has passed Travis and builds successfully on all platform by Gitlab.)

That change necessitated a re-working of the build system, so I took the opportunity to upgrade both `FindResvg.cmake` and `FindOpenShotAudio.cmake`. (Now with added `OpenShot::Audio` target! It's still not quite the same as having the libopenshot-audio build _automatically_ export that target with all of its configs, the way it _should_ be doing... but it's the next closest thing.)

I also modified `/src/CMakeLists.txt` to create an `ImageMagick::Magick++` target that holds all of the ImageMagick dependency configs. Using targets allows the dependency configs to automatically propagate even across separate directory trees. That way, everything is carried over from `/src/` into the separate `/bindings/` directory (as well as the existing `/tests/`) without having to be repeated.

I do plan to submit a _separate_ PR for Resvg, hopefully next week (just have to write it first), that will make our code compatible with the current Resvg 0.11. RazrFalcon has made pretty sweeping changes that have left it almost completely unrecognizable — not to mention API-incompatible — to the version we're using. (Which I think is 0.5?)

On the plus side, the new API allows us to add font file(s) to the font database directly, so we can inject the bundled Ubuntu (or any other font) the same way we do in Qt. Which means we may be able to count on at least one known-good font to use in our templates — except, of course, we still can't count on even _that_ in all of the Linux distribution packages where the font gets _**un**_-bundled. 

...But that'll require source code changes as well as build system changes, and it's a totally separate thing. 

Point is, _this_ PR has now become completely superfluous. Closing.

PR #549, opened 2020-08-07
Improved Support for Clang and new Qt Version: 5.15 (on Mac)

jonoomph:
Adding Qt 5.15 support for Mac builds, and misc fixes to allow clang++ and clang to correctly build libopenshot. This bumps the minimum Mac OS X version to 10.13 (due to our pre-compiled Qt library). In the future, we can compile that ourselves, and drop the minimum back to 10.9 (if there is a demand for it).

Also, changes GitLab builder to new Linux builder,

Related to PRs:
https://github.com/OpenShot/libopenshot-audio/pull/100
https://github.com/OpenShot/openshot-qt/pull/3604


jonoomph:
I'm not sure I understand the codecov/patch failure...

ferdnyc:
@jonoomph The explanation is shown in the detail view, though it still takes a bit of unpacking. Specifically, at this point in the diff: 
https://github.com/OpenShot/libopenshot/pull/549/files#diff-76f89b751d681404c52704a8e267f8f0

Codecov flags an "added" (really edited, but they count it as added) line of code as not being covered by tests. That line is the _only_ line of program source in the entire PR (all the rest are `#includes` or other non-program-code lines).

One of Codecov's criteria is what percentage of the diff adds lines that have unit test coverage. I think the default settings expect at least 50% coverage for code in the diff. (Obviously with only one qualifying line, the only real possibilities are 100% or 0%.)

PR #550, opened 2020-08-08
Opencv integration

BrennoCaldato:
Integrated OpenCV with OpenShot. Created Tracker, Stabilizer, and ObjectDetection Effects.
Already merged with the new-webengine-support branch.

The new OpenCV related effects work as follows:

1- For effects that cannot be processed in real-time a threaded preprocessing step is applied by calling the ClipProcessingJobs class.
2- The ClipProcessingJobs communicates between libopenshot and openshot-qt using Json.
3- After the preprocessing is done, the results are saved to disk in a serialized protobuf message.
4- The protobuf file path is passed to the effect class by openshot-qt.


jonoomph:
Merging into opencv branch so we can play with our build servers

PR #555, opened 2020-08-20
Use Qt::endl with QTextStream

ferdnyc:
Recent Qt releases complain about `endl` being used to terminate `QTextStream` lines, and recommend using `Qt::endl` instead. So, let's.

### DIFF
--- a/include/CacheDisk.h
+++ b/include/CacheDisk.h
@@ -38,8 +38,6 @@
 #include "Frame.h"
 #include "Exceptions.h"
 #include <QDir>
-#include <QString>
-#include <QTextStream>
 
 namespace openshot {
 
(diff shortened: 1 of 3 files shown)


ferdnyc:
Well, turns out `Qt::endl` wasn't a thing before Qt 5.14.0, so long story short, now we have a `QtUtilities.h` to go with our other compatibility headers. :stuck_out_tongue:

ferdnyc:
This is incredibly minor, since it only impacts one source file (the only place we use `QTextStream` for _output_). We should perhaps be more consistent and use it in more places, but for now it's only CacheDisk. Point is, it passes Travis and should be fine. Merging.

PR #557, opened 2020-08-21
Tests: Don't enable HW accel as side-effect

ferdnyc:
The Settings test attempts to test the class by changing settings variables and reading them back again. Problem is, that affects the REST of the unit tests.

So instead of changing `HARDWARE_ACCEL` (accidentally enabling HW accel and causing test failures), we'll diddle something more innocuous, like `OMP_THREADS`.


ferdnyc:
This is actively preventing libopenshot from being packaged for Fedora 33, so I'm planning to merge this ASAP so I can apply the commit to our 0.2.5 build tree.

PR #558, opened 2020-08-27
JUCE 5.4.7 changes

ferdnyc:
The only changes in this PR are to the Gitlab build script, but they're necessary to build libopenshot with the MacOS 10.11 SDK, which is required for JUCE 5.4.7.

So, in order to merge OpenShot/libopenshot-audio#88, we would need to merge this here as well.


ferdnyc:
One of these Travis jobs is checking the _wrong_ PR. I think I confused it when I first opened this. Let me push another commit to see if it sorts itself out.

ferdnyc:
(This PR only changes one line in `.gitlab-ci.yml` — it cannot _possibly_ fail a Travis run!

ferdnyc:
I seem to have successfully unbroken Travis. Now we wait...

ferdnyc:
Passed, merging.

(@jonoomph, like I said this only changes one line in the GitLab builder configuration, to use the tools from the `MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk` directory rather than the `MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk` directory. But it still sets `CMAKE_OSX_DEPLOYMENT_TARGET=10.9`, so that should be fine.)

jonoomph:
@ferdnyc Was `MacOSX10.11.sdk` a necessary change? Just curious, although we still have Qt which requires 10.11 anyway, so this isn't a big deal. But I thought using a different SDK folder would throw errors with our 10.9 target. If not, no worries.

ferdnyc:
@jonoomph It was necessary, 10.11 SDK is the minimum for JUCE 5.4.7 -- one of the only reasons we've been holding at 5.4.3 for so long. (The other was GCC 4.8 on Linux, as JUCE 5.4.7 now requires **full** C++11 compliance which GCC < 5 doesn't provide. But the bionic image solves that.) 

Now I can merge OpenShot/libopenshot-audio#88.

ferdnyc:
It certainly shouldn't throw errors, the whole point of the deployment target is that builds with newer SDKs can still target older releases. There'd be no point in having that variable if it had to be the same as the SDK. The errors were probably due to libstdc++.

ferdnyc:
At least in theory, we COULD be building with the 10.15 SDK (or 10.16 if that's out yet), and by targeting 10.9 the binaries would still be compatible with any 10.9+ Mac. (In _theory_.)

ferdnyc:
One thing I do still need to figure out is why _this_ is there, specifically the last two lines inside the `if()`:

https://github.com/OpenShot/libopenshot/blob/ac8966ad7d630dba45424b7624d5a9a0b1f06278/src/CMakeLists.txt#L55-L64

...And more importantly, really: whether it can be removed. I'm _guessing_ yes, because it's **NOT** part of the libopenshot-audio CMakeLists.txt, and I can't fathom why any part of libopenshot would need to have a C++-features macro removed from its symbols when compiling for MacOS.

ferdnyc:
Oh. Hmm. @jonoomph added that in 6e6b60fa22f774a0717423893b798ba1cf6be8c3 just two months ago — I suspect that was probably before we sorted out the switch from libstdc++ to libc++ as well, though? If that's the case, I have a feeling `libc++` makes it unnecessary (if it even does anything when libstdc++ isn't part of the build)... unless it's necessary _because of_ libc++?

PR #563, opened 2020-09-02
Timeline: Add clip/effect lookup api, GetMaxFrame/GetMaxTime method (w/ unit tests)

ferdnyc:
This PR adds two things to Timeline, primarily intended for use by OpenShot:

1. A lookup API to retrieve Clips, timeline Effects (transitions), and Clip Effects singly by `Id` string lookup
1. A `GetMaxTime()` that computes the time the last clip/effect ends and returns it as a `double` whole-and-partial-seconds value, and a `GetMaxFrame()` that returns the same thing accounting for frame rate as an `int64_t` frame number.

Full unit tests are included for `Timeline::GetMaxFrame()/Timeline::GetMaxTime()`, `Timeline::GetClip()`, `Timeline::GetTimelineEffect()`, and `Timeline::GetClipEffect()`. The `Clip::GetEffect()` implementation isn't unit-tested directly, but it is used by `Timeline::GetClipEffect()` so it's tested indirectly.

Along the way, I also deprecated `ReaderBase::GetClip` and `ReaderBase::SetClip`, because I wanted to reuse one of the names and anyway those names are _awful_ and confusing. They're now called `GetParentClip` and `SetParentClip`. The few calls in the libopenshot source are updated. The old names _are_ still available as deprecated aliases, for now.

#### Motivation

I noticed that there are a lot of places OpenShot does stuff like this, from `video_widget.py`:
```python
        # Get new clip for transform
        if clip_id:
            self.transforming_clip = Clip.get(id=clip_id)

            if self.transforming_clip:
                self.transforming_clip_object = None
                clips = get_app().window.timeline_sync.timeline.Clips()
                for clip in clips:
                    if clip.Id() == self.transforming_clip.id:
                        self.transforming_clip_object = clip
                        need_refresh = True
                        break
```

Which just seemed inefficient. _Sometimes_ OpenShot needs all of the clips, but when it only needs one, why should it search through them all when the library can do it much faster and  just return what's needed? So, now, that can be:
```python
        # Get new clip for transform
        if clip_id:
            self.transforming_clip = Clip.get(id=clip_id)
            self.transforming_clip_object =
                get_app().window.timeline_sync.timeline.GetClip(clip_id)
            if self.transforming_clip and self.transforming_clip_object:
                need_refresh = True
```

(`GetClip(id)` returns `nullptr` if it can't find a match, which in Python automatically translates to `None`. So there's no need to pre-set it to `None` or treat the result as fragile.)

And then, of course, there are the max-frame computations that get performed every time the preview is going to start playing, or needs to seek to the end of the timeline, etc. From `actionJumpEnd_trigger` in `main_window.py`:
```python
        # Determine max frame (based on clips)
        timeline_length = 0.0
        fps = get_app().window.timeline_sync.timeline.info.fps.ToFloat()
        clips = get_app().window.timeline_sync.timeline.Clips()
        for clip in clips:
            clip_last_frame = clip.Position() + clip.Duration()
            if clip_last_frame > timeline_length:
                # Set max length of timeline
                timeline_length = clip_last_frame
        # Convert to int and round
        timeline_length_int = round(timeline_length * fps) + 1
```

Now, **all** of that's just:
```python
timeline_length_int = get_app().window.timeline_sync.timeline.GetMaxFrame()
```
If the max time is needed (I'm sure somewhere we do that, too), it's just
```python
timeline_end = get_app().window.timeline_sync.timeline.GetMaxTime()
```

### DIFF
--- a/include/Timeline.h
+++ b/include/Timeline.h
@@ -61,7 +61,7 @@ namespace openshot {
 	/// from lowest layer to top layer (since that is the sequence they need to be combined), and then
 	/// by position (left to right).
 	struct CompareClips{
-		bool operator()( Clip* lhs, Clip* rhs){
+		bool operator()( openshot::Clip* lhs, openshot::Clip* rhs){
 			if( lhs->Layer() < rhs->Layer() ) return true;
 			if( lhs->Layer() == rhs->Layer() && lhs->Position() <= rhs->Position() ) return true;
 			return false;
@@ -71,7 +71,7 @@ namespace openshot {
 	/// from lowest layer to top layer (since that is sequence clips are combined), and then by
 	/// position, and then by effect order.
 	struct CompareEffects{
-		bool operator()( EffectBase* lhs, EffectBase* rhs){
+		bool operator()( openshot::EffectBase* lhs, openshot::EffectBase* rhs){
 			if( lhs->Layer() < rhs->Layer() ) return true;
 			if( lhs->Layer() == rhs->Layer() && lhs->Position() < rhs->Position() ) return true;
 			if( lhs->Layer() == rhs->Layer() && lhs->Position() == rhs->Position() && lhs->Order() > rhs->Order() ) return true;
@@ -146,33 +146,33 @@ namespace openshot {
 	 * t.Close();
 	 * @endcode
 	 */
-	class Timeline : public ReaderBase {
+	class Timeline : public openshot::ReaderBase {
 	private:
 		bool is_open; ///<Is Timeline Open?
 		bool auto_map_clips; ///< Auto map framerates and sample rates to all clips
-		std::list<Clip*> clips; ///<List of clips on this timeline
-		std::list<Clip*> closing_clips; ///<List of clips that need to be closed
-		std::map<Clip*, Clip*> open_clips; ///<List of 'opened' clips on this timeline
-		std::list<EffectBase*> effects; ///<List of clips on this timeline
-		CacheBase *final_cache; ///<Final cache of timeline frames
-		std::set<FrameMapper*> allocated_frame_mappers; ///< all the frame mappers we allocated and must free
+		std::list<openshot::Clip*> clips; ///<List of clips on this timeline
+		std::list<openshot::Clip*> closing_clips; ///<List of clips that need to be closed
+		std::map<openshot::Clip*, openshot::Clip*> open_clips; ///<List of 'opened' clips on this timeline
+		std::list<openshot::EffectBase*> effects; ///<List of clips on this timeline
+		openshot::CacheBase *final_cache; ///<Final cache of timeline frames
+		std::set<openshot::FrameMapper*> allocated_frame_mappers; ///< all the frame mappers we allocated and must free
 		bool managed_cache; ///< Does this timeline instance manage the cache object
 		std::string path; ///< Optional path of loaded UTF-8 OpenShot JSON project file
 
 		/// Process a new layer of video or audio
-		void add_layer(std::shared_ptr<Frame> new_frame, Clip* source_clip, int64_t clip_frame_number, int64_t timeline_frame_number, bool is_top_clip, float max_volume);
+		void add_layer(std::shared_ptr<openshot::Frame> new_frame, openshot::Clip* source_clip, int64_t clip_frame_number, int64_t timeline_frame_number, bool is_top_clip, float max_volume);
 
 		/// Apply a FrameMapper to a clip which matches the settings of this timeline
-		void apply_mapper_to_clip(Clip* clip);
+		void apply_mapper_to_clip(openshot::Clip* clip);
 
 		/// Apply JSON Diffs to various objects contained in this timeline
 		void apply_json_to_clips(Json::Value change); ///<Apply JSON diff to clips
 		void apply_json_to_effects(Json::Value change); ///< Apply JSON diff to effects
-		void apply_json_to_effects(Json::Value change, EffectBase* existing_effect); ///<Apply JSON diff to a specific effect
+		void apply_json_to_effects(Json::Value change, openshot::EffectBase* existing_effect); ///<Apply JSON diff to a specific effect
 		void apply_json_to_timeline(Json::Value change); ///<Apply JSON diff to timeline properties
 
 		/// Calculate time of a frame number, based on a framerate
-		double calculate_time(int64_t number, Fraction rate);
+		double calculate_time(int64_t number, openshot::Fraction rate);
 
 		/// Find intersecting (or non-intersecting) openshot::Clip objects
 		///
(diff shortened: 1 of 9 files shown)


ferdnyc:
#### Results
The savings from `GetMaxTime()` and `GetMaxFrame()` are fairly minor. They're meant more as a convenience than anything else. The _difference_ in time elapsed vs. the Python version appears actually even more significant than the lookup methods, but that's misleading because it takes so little time in either version that the differences are meaningless. (When I show the results of my benchmarks, keep in mind that the GetMaxFrame test is running **50,000** iterations, vs. only 2,000 for the `GetClip()` tests.)

The impact of `GetClip()` and friends, though, has the potential to be meaningful. I benchmarked it a bit, and here's the test setup I used. (Assume the requisite imports and etc.):
```python
t = openshot.Timeline(
    1280, 720, openshot.Fraction(30, 1), 48000, 2, openshot.LAYOUT_STEREO)
num_clips = 20

def lookup_locally(t, id):
    clips = t.Clips()
    for c in clips:
        if c.Id() == id:
            return c
def request_by_id(t, id):
    c = t.GetClip(id)
    return c
def maxframe_locally(t):
    clips = t.Clips()
    timeline_length = 0.0
    for clip in clips:
        clip_last_frame = clip.Position() + clip.Duration()
        if clip_last_frame > timeline_length:
            # Set max length of timeline
            timeline_length = clip_last_frame
    # Convert to int and round
    timeline_length_int = round(timeline_length * 30) + 1
    return timeline_length_int
def request_maxframe(t):
    n = t.GetMaxFrame()
    return n
def updatePositions(t):
    upd_clips = t.Clips()
    [t.RemoveClip(c) for c in upd_clips]
    for i, c in enumerate(upd_clips):
        p = c.Position()
        c.Position((p * 13333 * i) % 1599)
    shuffle(upd_clips)
    [t.AddClip(c) for c in upd_clips]
def main():
    test_clips = [openshot.Clip(TEST_MEDIA_FILE) for i in range(num_clips)]
    layers = [l for l in range(50, 50+num_clips)]
    starts = [s for s in range(num_clips)]
    positions = [p for p in range(20, 20+num_clips*2, 2)]
    shuffle(layers)
    shuffle(starts)
    shuffle(positions)

    for i, c in enumerate(test_clips):
        id = f"CLIP_NUMBER_{i}"
        c.Id(id)
        c.Layer(layers[i])
        c.Position(positions[i])
        c.Start(starts[i])
    shuffle(test_clips)
    [t.AddClip(c) for c in test_clips]
    t.Open()
```

So, it creates a somewhat-randomized timeline populated with 20 clips that all have shuffled Layer, Position, and Start values. `updatePositions()` will re-shuffle the clips and move their Positions, when called.

I ran best-of-5 trials with the `timeit` module on each of the four test functions (I'll attach the full source file, for anyone curious), and these were the results:
```
Running best-of-5 test for local Clip search...
  Fastest time: 0.9091/2000 runs (0.00045 per iteration)

Running best-of-5 test for GetClip() lookup by ID...
  Fastest time: 0.0978/2000 runs (0.00005 per iteration)

Running best-of-5 test for local max-frame search...
  Fastest time: 2.7240/50000 runs (0.000054 per iteration)

Running best-of-5 test for GetMaxFrame() request...
  Fastest time: 0.1839/50000 runs (0.000004 per iteration)
```
Like I said, the GetMaxFrame() results are almost too tiny to measure. The test also isn't entirely valid, as both versions run `updatePositions()` at intervals so a lot of the time logged is actually spent doing that. But the `GetClip()` tests only run `updatePositions()` _once_ at the start, and the rest of the test is them doing this:
```python
# Locally (the Clips()-and-search version)
x = [lookup_locally(t, f"CLIP_NUMBER_{i}") for i in range(20)];
[x_c.Id() for x_c in x]
# Remotely (using t.GetClip(id))
y = [request_by_id(t, f"CLIP_NUMBER_{i}") for i in range(20)];
[y_c.Id() for y_c in y]
```
Now, obviously _that's_ a bit contrived since it's unfair to the local version. If it _actually_ needed all 20 `Clip.Id()`s, it would make _sense_ to use `Clips()` to get the whole list, not request them one-at-a-time via `t.GetClip(id)`.

But in all the places we _are_ grabbing just a single `Id`, this helps a fair bit.

Benchmark code: [Bench_GetClip.py.txt](https://github.com/OpenShot/libopenshot/files/5160324/Bench_GetClip.py.txt)

include/Timeline.h
```diff
@@ -71,13 +71,28 @@ namespace openshot {
...
+		bool operator()(const openshot::Clip* lhs, const openshot::Clip* rhs) {
+			return (lhs->Position() + lhs->Duration())
+			    <= (rhs->Position() + rhs->Duration());
+	}};
+
+	/// Like CompareClipEndFrames, but for effects
+	struct CompareEffectEndFrames {
+		bool operator()(const openshot::EffectBase* lhs, const openshot::EffectBase* rhs) {
+			return (lhs->Position() + lhs->Duration())
+				<= (rhs->Position() + rhs->Duration());
```
ferdnyc, line 93: Clearly my cunning plan here was, because I couldn't decide on which formatting I preferred, I'd just use _BOTH_. Because that makes sense! :rage:

src/Timeline.cpp
```diff
@@ -263,6 +263,67 @@ void Timeline::RemoveClip(Clip* clip)
...
+		}
+	}
+	return nullptr;
+}
+
+openshot::EffectBase* Timeline::GetClipEffect(const std::string& id)
+{
+	// Search all clips for matching effect ID
+	for (const auto& clip : clips) {
+		auto e = clip->GetEffect(id);
```
ferdnyc, line 294:
```suggestion
		const auto e = clip->GetEffect(id);
```

ferdnyc:
Heh. You can see the exact point in the unit tests where I just got tired of writing code comments, apparently.

jonoomph:
@ferdnyc Nice work! I love it. I feel like these methods should have been written a long time ago, lol. This might create some merge issues for me (I'm refactoring the Timeline and Clip classes), but no big deal. Let's get this merged! Thanks!

include/Timeline.h
```diff
@@ -71,13 +71,28 @@ namespace openshot {
...
 			if( lhs->Layer() == rhs->Layer() && lhs->Position() == rhs->Position() && lhs->Order() > rhs->Order() ) return true;
 			return false;
 	}};
 
+	/// Comparison method for finding the far end of the timeline, by locating
+	/// the Clip with the highest end-frame number using std::max_element
+	struct CompareClipEndFrames {
+		bool operator()(const openshot::Clip* lhs, const openshot::Clip* rhs) {
+			return (lhs->Position() + lhs->Duration())
+			    <= (rhs->Position() + rhs->Duration());
```
ferdnyc, line 86:
```suggestion
			return (lhs->Position() + lhs->Duration())
			       <= (rhs->Position() + rhs->Duration());
```

ferdnyc:
@jonoomph 

>  I feel like these methods should have been written a long time ago

I was a bit surprised that there were no API calls to do these things. Especially since it causes OpenShot itself to jump through a lot of hoops doing them for itself.

> Let's get this merged!

Great, thanks! I just applied all of my self-review suggestions, as soon as Travis finishes I'll send it through. (Then I have to start on the Python-side changes to take advantage of these new calls.)

ferdnyc:
Oh, fun, seems Travis is a little backed up right now. Took almost 10 minutes for the first job to even _start_. (Hmm... although, maybe it was still finishing up our previous run, since I put through two quick edits in separate commits. Gotta remember to use that batch button.)

Eh. Regardless, it'll get there eventually I'm sure. Just a reminder that I shouldn't be hanging around waiting on Travis anyway.

jonoomph:
Also, I temp disabled Traivs to work on some of the older PRs...

PR #565, opened 2020-09-11
Fix the PTS offset logic error when first reading a file with FFmpegReader

jonoomph:
Use the calculated `0 - PTS`, unless it is too large (more than 1 second off from zero). This is related to a previous PR (with much discussion): https://github.com/OpenShot/libopenshot/pull/311. This PR fixes the regression related to limiting the offset, and I think makes more sense than previous suggestions. 

The PTS offset was always designed as a quick way to "zero" out the video PTS and audio PTS values, so we could calculate frame numbers, based on the starting PTS values, etc... But in my testing, some videos are encoded with negative timestamps, especially common with audio tracks, where the audio starts before the 1st video frame. OpenShot is not really designed to deal with this, and it causes us to search for matching packets which never come. This could/does introduce some amount of audio drift, just depending on how negative the audio timestamps are. Also, in my testing, I've come across videos that have garbage timestamps (usually negative values), for example -9999999999999999, and then the rest of the packets have normal PTS values. So, this is the background on why I want to limit the offset to some reasonable amount (-1 to +1 seconds in timebase units).

I've done some testing on this with my crazy test videos, and this seems to work great.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -1875,6 +1875,8 @@ void FFmpegReader::UpdatePTSOffset(bool is_video) {
 			// Also, determine if PTS is invalid (too far away from zero)
 			// We compare the PTS to the timebase value equal to 1 second (which means the PTS
 			// must be within the -1 second to +1 second of zero, otherwise we ignore it)
+			// TODO: Please see https://github.com/OpenShot/libopenshot/pull/565#issuecomment-690985272
+			// for ideas to improve this logic.
 			int64_t max_offset = info.video_timebase.Reciprocal().ToFloat();
 			if (video_pts_offset < -max_offset || video_pts_offset > max_offset) {
 				// Ignore PTS, it seems invalid
@@ -1892,6 +1894,8 @@ void FFmpegReader::UpdatePTSOffset(bool is_video) {
 			// Also, determine if PTS is invalid (too far away from zero)
 			// We compare the PTS to the timebase value equal to 1 second (which means the PTS
 			// must be within the -1 second to +1 second of zero, otherwise we ignore it)
+			// TODO: Please see https://github.com/OpenShot/libopenshot/pull/565#issuecomment-690985272
+			// for ideas to improve this logic.
 			audio_pts_offset = 0 - packet->pts;
 			int64_t max_offset = info.audio_timebase.Reciprocal().ToFloat();
 			if (audio_pts_offset < -max_offset || audio_pts_offset > max_offset) {
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -1868,8 +1868,18 @@ void FFmpegReader::UpdatePTSOffset(bool is_video) {
 		// VIDEO PACKET
 		if (video_pts_offset == 99999) // Has the offset been set yet?
 		{
-			// Find the difference between PTS and frame number (no more than 10 timebase units allowed)
-			video_pts_offset = 0 - std::max(GetVideoPTS(), (int64_t) info.video_timebase.ToInt() * 10);
+			// Find the difference between PTS and frame number
+			video_pts_offset = 0 - GetVideoPTS();
+
+			// Find the difference between PTS and frame number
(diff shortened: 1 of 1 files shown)


jonoomph:
@ferdnyc Please take a look, and let me know that I'm not crazy, haha. I think this is much "saner" than before, and saner than not filtering out crazy values.

ferdnyc:
> @ferdnyc Please take a look, and let me know that I'm not crazy, haha. I think this is much "saner" than before, and saner than not filtering out crazy values.

I'll take a look! I'm in my phone now, so I only managed a quick perusal with too little context. One question, though: Why is there a GetVideoPTS() method, but then the audio computation just uses `packet->pts`? Struck me as odd that they'd be different, but I don't yet know what the contents of GetVideoPTS() are.

jonoomph:
@ferdnyc That method does a little bit of checking before deciding which timestamp to return (PTS or DTS in some situations)
```
	int64_t current_pts = 0;
	if (packet->dts != AV_NOPTS_VALUE)
		current_pts = packet->dts;

```

ferdnyc:
> @ferdnyc That method does a little bit of checking before deciding which timestamp to return (PTS or DTS in some situations)
> 
> ```
> 	int64_t current_pts = 0;
> 	if (packet->dts != AV_NOPTS_VALUE)
> 		current_pts = packet->dts;
> ```

In every situation where it's set, in fact. ...That feels off to me.

I'm worried that I'm re-hashing #311 here, but... DTS is the _decode_ timestamp, so by unilaterally overriding the _presentation_ timestamp with it, isn't the code saying, "Always store these frames in the order they were decoded, even if we _know_ they're out of order from how they should be shown."? That's how I read the code above — which sounds off to me.

It's also very different from what `ffplay` does. Here's how their [`decoder_decode_frame`](https://ffmpeg.org/doxygen/trunk/ffplay_8c.html#a09fd32cf1173473a29fd8a3d9eb3a785) handles frame timestamps ([`ffplay.c@L602:626`](https://ffmpeg.org/doxygen/trunk/ffplay_8c_source.html#l00602))

```c
switch (d->avctx->codec_type) {
    case AVMEDIA_TYPE_VIDEO:
        ret = avcodec_receive_frame(d->avctx, frame);
        if (ret >= 0) {
            if (decoder_reorder_pts == -1) {
                frame->pts = frame->best_effort_timestamp;
            } else if (!decoder_reorder_pts) {
                frame->pts = frame->pkt_dts;
            }
        }
        break;
    case AVMEDIA_TYPE_AUDIO:
        ret = avcodec_receive_frame(d->avctx, frame);
        if (ret >= 0) {
            AVRational tb = (AVRational){1, frame->sample_rate};
            if (frame->pts != AV_NOPTS_VALUE)
                frame->pts = av_rescale_q(frame->pts, d->avctx->pkt_timebase, tb);
            else if (d->next_pts != AV_NOPTS_VALUE)
                frame->pts = av_rescale_q(d->next_pts, d->next_pts_tb, tb);
            if (frame->pts != AV_NOPTS_VALUE) {
                d->next_pts = frame->pts + frame->nb_samples;
                d->next_pts_tb = tb;
            }
        }
        break;
```

They don't even _touch_ the packet PTS, ever, if they're not prevented from reordering frames. Instead, the library handles the packets, and it cooks the available timestamp data down into a `best_effort_timestamp` provided in the **AVFrame** structure. 

That makes sense to me, as trying to reorder packets _before_ they're decoded into frames feels like rearranging deck chairs on a SCUD missile. You _know_ the thing's going to be thrown away when it's decoded into a more usable format, so why even mess with it? If we just hand every packet to `avcodec_receive_frame()`, and **then** deal with timing only after we get the AVFrame back, is there a chance that could work? Possibly even work better?

jonoomph:
> That makes sense to me, as trying to reorder packets _before_ they're decoded into frames feels like rearranging deck chairs on a SCUD missile.

I love it! I think your concerns and suggestions are worth investigating for sure! However, for this PR, I'm fixing the really broken logic back to what I initially intended. I'll add a TODO above this code with some links back to your suggestion, as I'm sure we'll revisit this again. Thx!

PR #567, opened 2020-09-13
Change of name of library

eisneinechse:
With the official inclusion of SVT-AV1 the library changed from svt_av1 to svtav1. This takes this into account


jonoomph:
Another Travis-ci hang for some reason. Merged!

PR #571, opened 2020-09-20
WIP: Clip Refactor (move Timeline Drawing code to Clip)

jonoomph:
This will likely be absorbed into another PR before I'm done. But I wanted to add some visibility to the work.


ferdnyc:
Still cleaning up the conflict-resolution here, I should have this back to passing within the hour.

ferdnyc:
Hrm. This could become an issue, if anyone ever decides to actually use the Ruby bindings...

> ```cmake
> [ 79%] Swig compile openshot.i for ruby
> src/Clip.h:96: Warning 802: Warning for openshot::Clip, base openshot::ReaderBase
> ignored. Multiple inheritance is not supported in Ruby.
> src/Timeline.h:166: Warning 802: Warning for openshot::Timeline,
> base openshot::ReaderBase ignored. Multiple inheritance is not supported in Ruby.
> [ 79%] Built target rbopenshot_swig_compilation
> ```

jonoomph:
> ```cmake
>  Multiple inheritance is not supported in Ruby.
> ```

I did notice this. Hmmm, I wonder how this will actually manifest at runtime. It apparently still compiles the swig module for Ruby, but somehow with missing multiple inheritance? Sounds like it fail to build the Ruby binding I'm missing something.

src/Clip.cpp
```diff
@@ -1053,7 +1116,250 @@ std::shared_ptr<Frame> Clip::apply_effects(std::shared_ptr<Frame> frame)
...
+			ZmqLogger::Instance()->AppendDebugMethod("Clip::apply_keyframes (Scale: SCALE_CROP)", "frame->number", frame->number, "source_width", source_size.width(), "source_height", source_size.height());
+			break;
+		}
+		case (SCALE_NONE): {
+			// Calculate ratio of source size to project size
+			// Even with no scaling, previews need to be adjusted correctly
+			// (otherwise NONE scaling draws the frame image outside of the preview)
+			float source_width_ratio = source_size.width() / float(width);
+			float source_height_ratio = source_size.height() / float(height);
+			source_size.scale(width * source_width_ratio, height * source_height_ratio, Qt::KeepAspectRatio);
```
ferdnyc, line 1214: This is exactly equivalent to:
```cpp
source_size.scale(source_size.width(), source_size.height(), Qt::KeepAspectRatio);
```
The two `width`s and `height`s in the ratios and in the arguments cancel each other out. So, it's a complete no-op.

ferdnyc: (By canceling, I mean:)

`width` * `source_width_ratio`
= <s>`width`</s> * ( `source_size.width()` / <s>`float(width)`</s> )
= `source_size.width()`

src/FrameMapper.cpp
```diff
@@ -116,6 +117,15 @@ void FrameMapper::Init()
...
 
+	// Get clip position from parent clip (if any)
+	float clipPosition = 0.0;
+	float clipStart = 0.0;
+	Clip *parent = (Clip *) ParentClip();
+	if (parent) {
+		clipPosition = parent->Position();
+		clipStart = parent->Start();
+	}
+
```
ferdnyc, line 128: Near as I can tell, this block of code (which is duplicated enough to send Codacy into conniptions) is only ever used to collect numbers to feed into `AdjustFrameNumber()`, yes?

So, why not just make **`AdjustFrameNumber()`** do the work of checking those values, by passing it the pointer?

```cpp
// In FrameMapper::Init() and friends
int remaining_samples = Frame::GetSamplesPerFrame(
    AdjustFrameNumber(frame_number, ParentClip()),
    target,
    reader->info.sample_rate,
    reader->info.channels
);

// Declared in FrameMapper.h
int64_t FrameMapper::AdjustFrameNumber(int64_t frame_number, ClipBase* parent);

// Defined in FrameMapper.cpp
int64_t FrameMapper::AdjustFrameNumber(int64_t frame_number, ClipBase* parent) {
    float position = parent ? parent->Position() : 0.0;
    float start = parent > parent->Start() : 0.0;

    // ...rest of code...
```
That way:
1. If `ParentClip()` returns `nullptr` that's fine, because it'll just get passed in to `AdjustFrameNumber` and you'll check it there.
1. You don't need any of this endlessly-duplicated setup code in the caller.

...Though — come to think of it, if `ParentClip()` is a method of the _FrameMapper_ instance, seems like you might not even need to pass **that** into `AdjustFrameNumber()`. Couldn't _it_ call `ParentClip()`?

ferdnyc:
> > ```cmake
> >  Multiple inheritance is not supported in Ruby.
> > ```
> 
> I did notice this. Hmmm, I wonder how this will actually manifest at runtime. It apparently still compiles the swig module for Ruby, but somehow with missing multiple inheritance? Sounds like it fail to build the Ruby binding I'm missing something.

Oh, it builds just fine, but what comes out... ain't right. For instance, the ruby `Openshot::Clip` class _**doesn't have**_ an `info` member anymore, or a `DisplayInfo()` method, because it no longer inherits from `ReaderBase`. It has the methods defined in `ClipBase`, like `Position()` and `Start()` and etc., but that's it:

```sh
$ cd build/bindings/ruby
$ irb -I . -ropenshot
irb(main):002:0> c = Openshot::Clip.new('/var/tmp/TaylorLachte.mp4')
irb(main):003:0> c.Open()
=> nil
irb(main):004:0> c.IsOpen()
=> true
irb(main):005:0> c.Position()
=> 0.0
irb(main):006:0> c.Duration()
=> 3.3366665840148926
irb(main):007:0> c.DisplayInfo()
        4: from /usr/bin/irb:23:in `<main>'
        3: from /usr/bin/irb:23:in `load'
        2: from /home/ferd/.gem/ruby/gems/irb-1.2.4/exe/irb:11:in `<top (required)>'
        1: from (irb):7
NoMethodError (undefined method `DisplayInfo' for #<Openshot::Clip:0x00005618a4aafee8>)
Did you mean?  display
irb(main):008:0> c.info
        5: from /usr/bin/irb:23:in `<main>'
        4: from /usr/bin/irb:23:in `load'
        3: from /home/ferd/.gem/ruby/gems/irb-1.2.4/exe/irb:11:in `<top (required)>'
        2: from (irb):7
        1: from (irb):8:in `rescue in irb_binding'
NoMethodError (undefined method `info' for #<Openshot::Clip:0x00005618a4aafee8>)
irb(main):009:0> 
```

src/Clip.cpp
```diff
@@ -308,46 +320,71 @@ float Clip::End() const
...
 }
 
-// Get an openshot::Frame object for a specific frame number of this reader.
-std::shared_ptr<Frame> Clip::GetFrame(int64_t requested_frame)
+// Create an openshot::Frame object for a specific frame number of this reader.
+std::shared_ptr<Frame> Clip::GetFrame(int64_t frame_number)
+{
+	// Check for open reader (or throw exception)
+	if (!is_open)
+		throw ReaderClosed("The Clip is closed.  Call Open() before calling this method", "N/A");
```
ferdnyc, line 328:
```suggestion
		throw ReaderClosed("The Clip is closed.  Call Open() before calling this method.");
```
The `file_path` argument to `ReaderClosed` is optional.

src/Clip.cpp
```diff
@@ -308,46 +320,71 @@ float Clip::End() const
...
+		// Throw error if reader not initialized
+		throw ReaderClosed("No Reader has been initialized for this Clip.  Call Reader(*reader) before calling this method.");
+}
+
+// Use an existing openshot::Frame object and draw this Clip's frame onto it
+std::shared_ptr<Frame> Clip::GetFrame(std::shared_ptr<openshot::Frame> frame, int64_t frame_number)
 {
+	// Check for open reader (or throw exception)
+	if (!is_open)
+		throw ReaderClosed("The Clip is closed.  Call Open() before calling this method", "N/A");
```
ferdnyc, line 349:
```suggestion
		throw ReaderClosed("The Clip is closed.  Call Open() before calling this method.");
```

src/Clip.cpp
```diff
@@ -633,21 +693,26 @@ int64_t Clip::adjust_frame_number_minimum(int64_t frame_number)
...
-	int samples_in_frame = Frame::GetSamplesPerFrame(number, reader->info.fps, reader->info.sample_rate, reader->info.channels);
-
 	try {
 		// Debug output
-		ZmqLogger::Instance()->AppendDebugMethod("Clip::GetOrCreateFrame (from reader)", "number", number, "samples_in_frame", samples_in_frame);
+		ZmqLogger::Instance()->AppendDebugMethod("Clip::GetOrCreateFrame (from reader)", "number", number);
 
 		// Attempt to get a frame (but this could fail if a reader has just been closed)
-		new_frame = reader->GetFrame(number);
+		std::shared_ptr<Frame> reader_frame = reader->GetFrame(number);
```
ferdnyc, line 701:
```suggestion
		auto reader_frame = reader->GetFrame(number);
```

src/Clip.cpp
```diff
@@ -633,21 +693,26 @@ int64_t Clip::adjust_frame_number_minimum(int64_t frame_number)
...
+			// Create a new copy of reader frame
+			// This allows a clip to modify the pixels and audio of this frame without
+			// changing the underlying reader's frame data
+			//std::shared_ptr<Frame> reader_copy(new Frame(number, 1, 1, "#000000", reader_frame->GetAudioSamplesCount(), reader_frame->GetAudioChannelsCount()));
+			std::shared_ptr<Frame> reader_copy(new Frame(*reader_frame.get()));
+			{
+				reader_copy->SampleRate(reader_frame->SampleRate());
+				reader_copy->ChannelsLayout(reader_frame->ChannelsLayout());
+			}
+			return reader_copy;
```
ferdnyc, line 714:
```suggestion
			auto reader_copy = std::make_shared<Frame>(*reader_frame.get());
			reader_copy->SampleRate(reader_frame->SampleRate());
			reader_copy->ChannelsLayout(reader_frame->ChannelsLayout());
			return reader_copy;
```

src/Clip.cpp
```diff
@@ -657,14 +722,17 @@ std::shared_ptr<Frame> Clip::GetOrCreateFrame(int64_t number)
...
+	// Estimate # of samples needed for this frame
+	int estimated_samples_in_frame = Frame::GetSamplesPerFrame(number, reader->info.fps, reader->info.sample_rate, reader->info.channels);
+
 	// Debug output
-	ZmqLogger::Instance()->AppendDebugMethod("Clip::GetOrCreateFrame (create blank)", "number", number, "samples_in_frame", samples_in_frame);
+	ZmqLogger::Instance()->AppendDebugMethod("Clip::GetOrCreateFrame (create blank)", "number", number, "estimated_samples_in_frame", estimated_samples_in_frame);
 
 	// Create blank frame
-	new_frame = std::make_shared<Frame>(number, reader->info.width, reader->info.height, "#000000", samples_in_frame, reader->info.channels);
+	std::shared_ptr<Frame> new_frame = std::make_shared<Frame>(number, reader->info.width, reader->info.height, "#000000", estimated_samples_in_frame, reader->info.channels);
```
ferdnyc, line 732:
```suggestion
	auto new_frame = std::make_shared<Frame>(
		number, reader->info.width, reader->info.height,
		"#000000", estimated_samples_in_frame, reader->info.channels);
```

src/Clip.cpp
```diff
@@ -1053,7 +1116,250 @@ std::shared_ptr<Frame> Clip::apply_effects(std::shared_ptr<Frame> frame)
...
+	QTransform transform;
+
+	// Transform source image (if needed)
+	ZmqLogger::Instance()->AppendDebugMethod("Clip::apply_keyframes (Build QTransform - if needed)", "frame->number", frame->number, "x", x, "y", y, "r", r, "sx", sx, "sy", sy);
+
+	if (!isEqual(x, 0) || !isEqual(y, 0)) {
+		// TRANSLATE/MOVE CLIP
+		transform.translate(x, y);
+		transformed = true;
+	}
```
ferdnyc, line 1289: The `transformed` boolean isn't really that necessary here. QTransform has an `isIdentity()` method that serves the same purpose. If `transform.isIdentity()` is `false`, some sort of transform was applied. If it's `true`, then the transform hasn't been modified — _**OR**_ all of the modifications canceled each other out, and the _end result_ is a no-op. That's a potential optimization that an external boolean flag can't give you.

(Then again, Qt is most likely efficient enough that if you _do_ pass an identity transform to `QPainter::setTransform()`, it'll just be ignored without affecting either performance or output, so even checking `isIdentity()` is probably unnecessary. I'd be surprised if there's any measurable impact from applying an "unnecessary" identity transform.)

src/Clip.cpp
```diff
@@ -1053,7 +1116,250 @@ std::shared_ptr<Frame> Clip::apply_effects(std::shared_ptr<Frame> frame)
...
+		transform.scale(source_width_scale, source_height_scale);
+		transformed = true;
+	}
+
+	// Debug output
+	ZmqLogger::Instance()->AppendDebugMethod("Clip::apply_keyframes (Transform: Composite Image Layer: Prepare)", "frame->number", frame->number, "transformed", transformed);
+
+	/* COMPOSITE SOURCE IMAGE (LAYER) ONTO FINAL IMAGE */
+	std::shared_ptr<QImage> new_image;
+	new_image = std::shared_ptr<QImage>(new QImage(QSize(width, height), source_image->format()));
```
ferdnyc, line 1316:
```suggestion
	auto new_image = std::make_shared<QImage>(QSize(width, height), source_image->format());
```

src/ClipBase.h
```diff
@@ -33,10 +33,13 @@
...
 #include <sstream>
+#include "CacheMemory.h"
 #include "Exceptions.h"
+#include "Frame.h"
 #include "Point.h"
 #include "KeyFrame.h"
 #include "Json.h"
+#include "TimelineBase.h"
 
 namespace openshot {
```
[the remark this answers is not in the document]
ferdnyc: (I can't add the corresponding suggestion for `ClipBase.cpp` since it's not part of the PR changes.)

src/Clip.cpp
```diff
@@ -1053,7 +1116,250 @@ std::shared_ptr<Frame> Clip::apply_effects(std::shared_ptr<Frame> frame)
...
+		}
+		case (SCALE_CROP): {
+			QSize width_size(width, round(width / (float(source_size.width()) / float(source_size.height()))));
+			QSize height_size(round(height / (float(source_size.height()) / float(source_size.width()))), height);
+
+			// respect aspect ratio
+			if (width_size.width() >= width && width_size.height() >= height)
+				source_size.scale(width_size.width(), width_size.height(), Qt::KeepAspectRatio);
+			else
+				source_size.scale(height_size.width(), height_size.height(), Qt::KeepAspectRatio);
```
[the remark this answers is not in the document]
jonoomph: OMG

src/FrameMapper.cpp
jonoomph: Yes, ParentClip() could be directly called in AdjustFrameNumber. I think my idea was to prevent calling ParentClip() over and over again... but the cost is actually trivial... so I like this idea.

src/Clip.cpp
jonoomph: Thanks, I've removed the boolean check

src/FrameMapper.cpp
jonoomph: I've moved this duplicated logic into AdjustFrameNumber. Thanks.

jonoomph:
@ferdnyc Regarding Ruby, I don't have an easy solution for this one. I tried a ton of different approaches to access the parent timeline of a clip. But since the Timeline class depends on the Clip, the Clip interface can't also depend on the Timeline. Creating a base class of the Timeline to expose a few of the parameters needed seemed to make the most sense. However, simple forward declarations don't work, since we are also access some members of the class. I had to revert the one suggestion because it fails to compile.

In summary, I've tried a bunch of things, and what we have now seems the simplest (and only) solution I could come up with, but I'm open to ideas.

ferdnyc:
@jonoomph 

> However, simple forward declarations don't work, since we are also access some members of the class. I had to revert the one suggestion because it fails to compile.

Hm, really? In the _header_? I missed that, somehow.

(In the implementation, sure. Part of what I was suggesting was that we **add** `#include "TimelineBase.h"` to the `ClipBase.cpp` file, _instead of_ the header. That way it would have access to the class definition. I just couldn't actually submit that as a suggestion via the Github review, because `ClipBase.cpp` wasn't part of this PR yet.)

Well, anyway, that's largely housekeeping. But on this...

> In summary, I've tried a bunch of things, and what we have now seems the simplest (and only) solution I could come up with, but I'm open to ideas.

Yeah, I've got nothin' either, when it comes to the multiple inheritance problem. (Which the forward declaration suggestion wouldn't solve anyway, that was just about the source-file dependency tree.)

I did have one awful thought last night, but I hated myself immediately for it so I didn't bring it up:

...What if both TimelineBase and ClipBase were derived from ReaderBase?

It _could_ possibly solve the Ruby problem (at the expense of, you know, sanity), as it would eliminate the multiple inheritance. (It seems like it would, anyway. Not 100% sure. Like I said, I felt dirty as soon as I thought it, so I kind of pushed it way down and tried not to think about it.)

jonoomph:
> ...What if both TimelineBase and ClipBase were derived from ReaderBase?

That is not a terrible idea. In effect, both Clip and Timeline are readers now. We would just need to move some things around. But I'll keep that a separate idea from this PR. But I do agree it would solve the problem! Nice

ferdnyc:
@jonoomph 

Heh. libopenshot 0.3.0, the "It's Readers all the way down" edition. :laughing:

jonoomph:
Finally merging this branch, as I have other branches which depend on this one. If this causes any huge issues, we can open up another PR. :crossed_fingers:

jonoomph:
Fingers crossed emoji is right next to the middle finger one... lol. That is dangerous!

jonoomph:
@ferdnyc You might be right about the reverse playback. I will look into that. In theory, the direction of the player should drive the direction of the clip caching. Yes, it is a bit scorched earth (and was somewhat experimental), but it also solves some problems (as scorching the earth tends to do). We can always revert it, if the problems outweigh the fixes. 

This was really designed to solve 1 thing:
- Jumping the playhead around creating dozens of pockets of cached frames... which have audio alignment issues, since we have cleared internal buffers and Seeked around... causing tiny mis-alignments that are unique to the preview experience.

But in my testing (including the changes to video caching), I found things much more responsive and continuous in OpenShot. So I thought, dang, maybe this is better. But maybe it's not. :grimacing:

ferdnyc:
> @ferdnyc You might be right about the reverse playback. I will look into that. In theory, the direction of the player should drive the direction of the clip caching. 

Yeah, but sorta-not-really. You can't escape the basic reality that encoded video is tuned to be read primarily in one direction — except for a few specialty transport formats or full-frame raw video streams (@ gigabytes-per-second of bandwidth), video wants to be read "in, like... forwards".

I think what the reverse player would have to do, to cache most effectively, is figure out the "frame chunk size" that's going to be produced by the reader (based on the processor count, etc.) and then keep jumping that many frames "ahead of" the "end" of the cache (really, jumping _before_ the beginning of the cache), stuffing N new frames onto the front in one go, and then jumping again. Like a video-frame-hoarding Scott Bakula.

> But in my testing (including the changes to video caching), I found things much more responsive and continuous in OpenShot. So I thought, dang, maybe this is better. But maybe it's not.

I wouldn't be surprised if it _is_, for the most degenerate case where OpenShot is easily able to keep up with the input. Those are the times when you really **don't** need a cache, and a small readahead buffer is plenty. But as soon as the readahead starts falling behind and can't keep up with the playhead, then everything goes to hell fast. That's true even _with_ the cache, to some extent (unfortunately), but at least with a cache you can give it a run to fill the buffer, then back up and play from the cache at reasonable speed.

ferdnyc:
> I think what the reverse player would have to do, to cache most effectively, is figure out the "frame chunk size" that's going to be produced by the reader (based on the processor count, etc.) and then keep jumping that many frames "ahead of" the "end" of the cache (really, jumping _before_ the beginning of the cache), stuffing N new frames onto the front in one go, and then jumping again. Like a video-frame-hoarding Scott Bakula.

...And even doing that, there are going to be a _ton_ of situations where it takes the next leap and ends up smack in the middle of <a>the Kennedy assassination</s. — sorry, no, that's _Quantum Leap_ again — smack in the middle of a bunch of _i_-frames that really, really don't _want_ to be read starting from the middle, and it's going to have to (very quickly!) adjust in one direction or the other and then figure out how to deal with the break in its preferred stride. That kind of thing is where timestamping and frame ordering (a) become total hell, but also (b) are absolutely critical.

Issue #573, opened 2020-10-03
Cannot build openshot-player. 


meksor:
Greetings!
After including libopenshot in my cmake project via:

```
ExternalProject_Add(libopenshot
  GIT_REPOSITORY "https://github.com/OpenShot/libopenshot"
  GIT_TAG "origin/develop"
  PREFIX ${CMAKE_BINARY_DIR}/libopenshot
  INSTALL_COMMAND ""
)
```
(I'm using the develop branch because of an ffmpeg incompatibility on tag "v0.2.5", as documented here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=967016)

...and trying to compile, I get the following linker error:

```
[...]
[ 22%] Performing build step for 'libopenshot'
[  1%] Automatic MOC for target openshot
[  1%] Built target openshot_autogen
[ 84%] Built target openshot
[ 85%] Linking CXX executable openshot-player
/usr/bin/ld: libopenshot.so.0.2.5: undefined reference to `Magick::Image::columns() const'
/usr/bin/ld: libopenshot.so.0.2.5: undefined reference to `Magick::Image::rows() const'
/usr/bin/ld: libopenshot.so.0.2.5: undefined reference to `Magick::DrawableBase::DrawableBase()'
/usr/bin/ld: libopenshot.so.0.2.5: undefined reference to `Magick::Geometry::height() const'
/usr/bin/ld: libopenshot.so.0.2.5: undefined reference to `Magick::Geometry::width() const'
/usr/bin/ld: libopenshot.so.0.2.5: undefined reference to `Magick::Image::draw(std::vector<Magick::Drawable, std::allocator<Magick::Drawable> > const&)'
/usr/bin/ld: libopenshot.so.0.2.5: undefined reference to `Magick::Geometry::Geometry(unsigned long, unsigned long, long, long)'
/usr/bin/ld: libopenshot.so.0.2.5: undefined reference to `Magick::Geometry::aspect(bool)'
/usr/bin/ld: libopenshot.so.0.2.5: undefined reference to `Magick::Image::alpha(bool)'
collect2: error: ld returned 1 exit status
make[5]: *** [src/CMakeFiles/openshot-player.dir/build.make:123: src/openshot-player] Error 1
make[4]: *** [CMakeFiles/Makefile2:258: src/CMakeFiles/openshot-player.dir/all] Error 2
make[3]: *** [Makefile:171: all] Error 2
[...]

PR #574, opened 2020-10-04
Add ENABLE_MAGICK CMake option (default ON)

ferdnyc:
This will allow building with ImageMagick support to be switched off (using `-DENABLE_MAGICK=0` on the `cmake` command line), even if it's detected on the system (which was previously not possible). 

Also: Change the default build type (when not specified at all) from Debug to Release, which is more typically what's expected.


ferdnyc:
I'm going to merge this immediately since it's fairly trivial, and I'm hoping it might help solve some issues seen in #573 .

Issue #576, opened 2020-10-13
Crash inside sws_scale with certain inputs


ckirmse:
It's easily reproducible to get both invalid reads and writes (causing a crash; valgrind shows more detail of what's wrong) by using an input video whose width is not a multiple of 4. The underlying issue is that the alignment passed to a bunch of libav functions is 1, but for sws_scale to work in general it is apparently supposed to be 32 (the answer here helped me a lot: https://stackoverflow.com/questions/31253485/how-do-you-resize-an-avframe/31270501#31270501 ).

@ferdnyc this seemed up your alley--I have a rough patch that fixes the problem. Would you be interested in me polishing it up and submitting it? I'd love a second set of eyes because this is tricky stuff.

ferdnyc:
@ckirmse 
> 
> @ferdnyc this seemed up your alley

Oh, my no. Well, not by _choice_, and not in the sense that I have any particular knowledge of that code — @eisneinechse and @jonoomph would be the go-tos/primaries. But in the sense that I'd like to see this _fixed_ and can help turn the wheels on getting a PR merged, yes definitely.

> --I have a rough patch that fixes the problem. Would you be interested in me polishing it up and submitting it? I'd love a second set of eyes because this is tricky stuff.

Absolutely! Heck, I'd be interested in you submitting it _unpolished_, or polishing it along the way. If it's truly in an un-_finished_ state, just throw `WIP:` at the start of the PR title.

It's always easiest to discuss code changes in the context of _the actual code_, and the simplest path to that is an open PR. While it's _possible_ to view branches in forked repos, comment on commits, and discuss in-development changes that way, it's just not as convenient. (For starters, it makes it **much** harder for any third parties to find or jump into those conversations.)

> It's easily reproducible to get both invalid reads and writes (causing a crash; valgrind shows more detail of what's wrong) by using an input video whose width is not a multiple of 4. The underlying issue is that the alignment passed to a bunch of libav functions is 1, but for sws_scale to work in general it is apparently supposed to be 32 (the answer here helped me a lot: https://stackoverflow.com/questions/31253485/how-do-you-resize-an-avframe/31270501#31270501 ).

Mmm, I see where we have `alignment` just hardcoded to `1`:
https://github.com/OpenShot/libopenshot/blob/414a2cdaa3e0b3a009f335670d65179d8366b470/include/FFmpegUtilities.h#L162
...in _multiple_ places, actually:
[...]

PR #577, opened 2020-10-13
WIP: fix crashes with video widths not multiple of 4

ckirmse:
- need to align 32 everywhere for sws_scale to not crash
- older ffmpegs don't have a nice function (thanks for telling me about it @ferdnyc , I didn't know about av_cpu_max_align), so 32 hard-coded in a few places
- needed to tell QImage about the change too.
- fixes issue #576


include/Frame.h
```diff
@@ -156,7 +156,7 @@ namespace openshot
 		void AddColor(int new_width, int new_height, std::string new_color);
 
 		/// Add (or replace) pixel data to the frame
-		void AddImage(int new_width, int new_height, int bytes_per_pixel, QImage::Format type, const unsigned char *pixels_);
+		void AddImage(int new_width, int new_height, int bytes_per_pixel, int bytes_per_line, QImage::Format type, const unsigned char *pixels_);
```
ferdnyc, line 159: Perhaps it'd be better to add this as an additional overload, rather than redefining the existing `Frame::AddImage()` semantics? There may be external API consumers who rely on the previous signature.

To make overload resolution slightly less harrowing, there's _some_ argument for tacking `bytes_per_line` on at the very end. Arguably it's a detail specific to the **buffer**, not a parameter describing the image (like all of the other parameters surrounding it), so it should (optionally) follow the buffer it describes — and only in the cases where it's necessary to specify it. (IOW, only when `bytes_per_line != width * bytes_per_pixel`.)

src/Frame.cpp
```diff
@@ -759,11 +759,11 @@ void Frame::AddColor(int new_width, int new_height, std::string new_color)
 }
 
 // Add (or replace) pixel data to the frame
-void Frame::AddImage(int new_width, int new_height, int bytes_per_pixel, QImage::Format type, const unsigned char *pixels_)
+void Frame::AddImage(int new_width, int new_height, int bytes_per_pixel, int bytes_per_line, QImage::Format type, const unsigned char *pixels_)
```
ferdnyc, line 762: The old signature could then just be turned into:
```c++
void Frame::AddImage(
    int width, int height, int bytes_per_pixel,
    QImage::Format type, const unsigned char* pixels_)
{
       AddImage(width, height, bytes_per_pixel, type, *pixels_, (width * bytes_per_pixel));
}
```

src/Frame.cpp
```diff
@@ -759,11 +759,11 @@ void Frame::AddColor(int new_width, int new_height, std::string new_color)
...
 {
 	// Create new buffer
 	const GenericScopedLock<juce::CriticalSection> lock(addingImageSection);
-	int buffer_size = new_width * new_height * bytes_per_pixel;
+	int buffer_size = bytes_per_line * new_height;
```
ferdnyc, line 766: Alternatively, if you _are_ gonna change the signature of `Frame::AddImage()`, it makes sense to _**drop**_ `bytes_per_pixel` (and replace it with `bytes_per_line`), since it no longer has **any** purpose.

(Plus it's totally implied by `QImage::Format`, so it was never necessary to pass it in anyway.)

src/Frame.cpp
```diff
@@ -772,7 +772,7 @@ void Frame::AddImage(int new_width, int new_height, int bytes_per_pixel, QImage:
 	// Create new image object, and fill with pixel data
 	#pragma omp critical (AddImage)
 	{
-		image = std::shared_ptr<QImage>(new QImage(qbuffer, new_width, new_height, new_width * bytes_per_pixel, type, (QImageCleanupFunction) &openshot::Frame::cleanUpBuffer, (void*) qbuffer));
+		image = std::shared_ptr<QImage>(new QImage(qbuffer, new_width, new_height, bytes_per_line, type, (QImageCleanupFunction) &openshot::Frame::cleanUpBuffer, (void*) qbuffer));
```
ferdnyc, line 775: If `openshot::Frame`s are going to contain `QImage`s with possible line-padding, I think this is actually going to necessitate an audit of _all_ of our image-processing code. Because, for instance, I can't see how this **wouldn't**  break this loop:

https://github.com/OpenShot/libopenshot/blob/414a2cdaa3e0b3a009f335670d65179d8366b470/src/effects/Shift.cpp#L79-L84

And potentially dozens more just like it, hiding scattered around the code. We are unfortunately **not** rigorous about using `QImage::[const]scanLine`. Not at _all_.

ferdnyc (commented):
Looks good, for the most part. I brought up a couple of points about `AddImage()`, but nothing critical or imperative.

My real concerns are about effects on the rest of the code, which is unfortunately not in a state where it's resilient enough to easily weather a change like this. We're probably going to have to shore up a bunch of bad algorithms elsewhere.

src/Frame.cpp
ckirmse: Very good catch. Now, using QImage but NOT using bytesPerLine(), in my opinion, is very poor code hygiene and we should fix all of these cases, since we should not assume only certain types of QImages get here.

However, fixing all of those effects to use bytesPerLine() properly is costly. Here's another option: we could leave the input to AddFrame with this bytes_per_line, but we could copy row-by-row into the QImage _without_ any extra padding, so the QImages remain "packed" and all those effects will work unchanged. This would be slightly lower performance (hundreds of row-by-row copies instead of one bit batch copy), but that may not be meaningful, and it wouldn't change assumptions out from underneath all our users of QImage.

As I said above, I'm open to any/all of these thoughts. Let me know what you think and I'll revise the pull request based on what we decide!

include/Frame.h
ckirmse: A lot of good questions there, and I'm open to any solution. Here's my thoughts:

- I don't consider the Frame class part of the externally exposed API of libopenshot, so since there are no other uses of this function, I felt safe changing it. If we consider the Frame class as something external callers can use, then I agree it should be a separate overload.

- I see your point about bytes_per_line at the end; it doesn't matter to me, and I see the argument that it's a "property" of buffer.

PR #580, opened 2020-10-17
Tests: Increase coverage for ImageReader, 100% for ImageWriter

ferdnyc:
`ImageWriter.cpp` was literally _one line_ of code away from full coverage, so I wrote a `CHECK_THROW()` test for that one line (an exception).

I also added `CHECK_THROW()`s for two exceptions in `ImageReader.cpp`, and performed the same constructor consolidation trick on it that I used for FFmpegReader (make `inspect_reader` an optional arg with a default value of true, instead of having two overloads, one with and one without). That should complete the coverage of every method in ImageReader _except_ for the `Json::Value`-form setter and getter.

### DIFF
--- a/include/ImageReader.h
+++ b/include/ImageReader.h
@@ -76,15 +76,15 @@ namespace openshot
 		bool is_open;
 
 	public:
-
-		/// Constructor for ImageReader.  This automatically opens the media file and loads
-		/// frame 1, or it throws one of the following exceptions.
-		ImageReader(std::string path);
-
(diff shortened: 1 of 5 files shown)


ferdnyc:
SH*T!!! We're _still_ getting those duplicate Travis checks. They [thought they'd fixed that](https://travis-ci.community/t/duplicate-job-attempts-for-pr-builds-causing-queued-build-created-problems/9776/23).

PR #583, opened 2020-10-27
New "Caption" Effect (Supports VTT and Subrip formats)

jonoomph:
This is a new effect, which allows for VTT or Subrip caption data to be displayed over a clip. Many different keyframes are available, from font-size, color, background color, alpha, corner radius, padding, font family, stroke, etc... Also, word-wrap is supported, and custom margins for left, top, and right (didn't include bottom because I didn't want to crop the text off). You can also set/animate a custom fade-in / fade-out for each caption, which gives it a nice animated look.

This **not** yet implemented:
- Light formatting: `<b><i><u><font color=""></font></u></i></b>`
- Custom regions (i.e. 2 dialogs happening, and displayed in different areas)

Pretty much we parse the basic timestamp and caption text out, and display when it's the correct time... augmented by all the animation keyframes that are set.

CaptionsOpenShot


jonoomph:
Merging this now

PR #585, opened 2020-10-28
3 New OpenCV Effects (Tracker, Stabilization, and Object Detection)

BrennoCaldato:
- Added Opencv and Protobuf dependencies to libopenshot.
- Created Opencv related effects: Tracker, Video stabilization, and Object detection
- Created a class to handle effects pre-processing on a separated thread and communicates with openshot-qt

The idea is that processing frame by frame data using algorithms or AI is slow on most consumer hardware, and would be unusably slow in OpenShot's real-time preview. To counter this, we pre-process the image data before the effect is dropped onto the clip (for example, determine the amount of stabilization required per frame, or determine where an object moves on each frame). The user get's a progress bar, and it takes as long as it takes. Once done, we save the data using Protobuf files (for compression and speed), and our Effect reads the data from protobuf in real-time. Thus, the real-time previews are super responsive and very usable.

Related UI for this code: https://github.com/OpenShot/openshot-qt/pull/3806

**Tracker:**
Tracker

**Object Detection:**
Screenshot from 2020-10-29 03-08-10

**Stabilization:**
Screenshot from 2020-10-29 03-09-21


src/CMakeLists.txt
```diff
@@ -34,6 +34,132 @@ if (POLICY CMP0057)
   cmake_policy(SET CMP0057 NEW)
 endif()
 
+# Juce requires either DEBUG or NDEBUG to be defined on MacOS.
```
jonoomph, line 37: We might need to rebase this file off of develop, since it's changed quite a bit, and I think this PR will revert some of those changes....

src/CVObjectDetection.h
```diff
@@ -0,0 +1,123 @@
+/**
+ * @file
+ * @brief Header file for CVObjectDetection class
```
jonoomph, line 3: Can we add a more descriptive title/description for all 3 of the effects?

src/Clip.h
```diff
@@ -245,6 +266,9 @@ namespace openshot {
 		bool Waveform() { return waveform; } ///< Get the waveform property of this clip
 		void Waveform(bool value) { waveform = value; } ///< Set the waveform property of this clip
 
+		/// Stabilize the clip using opencv and opticalflow
```
jonoomph, line 269: Let's remove these commented lines

src/Frame.cpp
```diff
@@ -935,6 +938,55 @@ std::shared_ptr<QImage> Frame::GetImage()
...
+	// if (imagecv.empty())
+	// Convert Qimage to Mat
+	imagecv = Qimage2mat(image);
+
+	return imagecv;
+}
+
+std::shared_ptr<QImage> Frame::Mat2Qimage(cv::Mat img){
+	cv::cvtColor(img, img, cv::COLOR_BGR2RGB);
+	QImage qimg((uchar*) img.data, img.cols, img.rows, img.step, QImage::Format_RGB888);
```
jonoomph, line 971: We are using `QImage::Format_RGBA8888_Premultiplied` as our primary color format now, let's change that here

src/Frame.h
```diff
@@ -98,7 +107,7 @@ namespace openshot
 	class Frame
 	{
 	private:
-		std::shared_ptr<QImage> image;
+		std::shared_ptr<QImage> image; ///< RGBA Format
```
jonoomph, line 110: Remove this comment (and any similar ones), since our format is QImage::Format_RGBA8888_Premultiplied, it's a little misleading I think

tests/CMakeLists.txt
```diff
@@ -56,6 +56,24 @@ if(ENABLE_BLACKMAGIC)
 	endif()
 endif()
 
+####################### OPENCV ##########################
```
jonoomph, line 59: Same suggestion on all CmakeLists.txt files... let's rebase this file on 'develop', and then add back in these changes.... so we aren't reverting any updates.

jonoomph:
@BrennoCaldato Also, the `opencv` branch is failing on mac and windows builders (not sure if you already saw that). Thx!

BrennoCaldato:
> @BrennoCaldato Also, the `opencv` branch is failing on mac and windows builders (not sure if you already saw that). Thx!

@jonoomph Thanks for the comments. I'll review it and check the builders

tests/CVTracker_Tests.cpp
```diff
@@ -0,0 +1,125 @@
...
+        // Create a video clip
+        std::stringstream path;
+        path << TEST_MEDIA_PATH << "test.avi";
+
+        // Open clip
+        openshot::Clip c1(path.str());
+        c1.Open();
+
+        // Create tracker
+        CVTracker kcfTracker("{\"protobuf_data_path\": \"\", \"tracker_type\": \"KCF\", \"bbox\": {\"x\": 294, \"y\": 102, \"w\": 180, \"h\": 166}, \"first_frame\": 0}", processingController);
```
ferdnyc, line 58: As a general suggestion, [raw string literals](https://en.cppreference.com/w/cpp/language/string_literal) (the ugly-but-handy <code>R"<var>delim</var>( ... )<var>delim</var>"</code> format strings) are your friend in these situations. See the JSON tests I just added to a few classes in #586, like:

https://github.com/OpenShot/libopenshot/blob/db738cecea98ae47e24703670d9cc15ede9a33c2/tests/Color_Tests.cpp#L165-L176

tests/CMakeLists.txt
```diff
@@ -74,17 +92,35 @@ set(OPENSHOT_TEST_FILES
...
   Settings_Tests.cpp
-  Timeline_Tests.cpp )
+  Timeline_Tests.cpp)
+
+##########  SET OPENCV RELATED TEST FILES  ###############
+set(OPENSHOT_CV_TEST_FILES
+  CVTracker_Tests.cpp
+  CVStabilizer_Tests.cpp
+  # CVObjectDetection_Tests.cpp
+  )
```
ferdnyc, line 102:
```suggestion
##########  SET OPENCV RELATED TEST FILES  ###############
if(OpenCV_FOUND)
  set(OPENSHOT_CV_TEST_FILES
    CVTracker_Tests.cpp
    CVStabilizer_Tests.cpp
    # CVObjectDetection_Tests.cpp
    )
  set(OPENSHOT_CV_LIBRARIES
    ${OpenCV_LIBS}
    ${PROTOBUF_LIBRARY}
  )
endif()
```
Then you can just use those variables unconditionally in the executable definition. CMake will expand any undefined variable to an empty string.

tests/CMakeLists.txt
```diff
@@ -74,17 +92,35 @@ set(OPENSHOT_TEST_FILES
...
 # Create unit test executable (openshot-test)
 message (STATUS "Tests enabled, test executable will be built as tests/openshot-test")
-add_executable(openshot-test
-  tests.cpp
-  ${OPENSHOT_TEST_FILES} )
+if (OpenCV_FOUND)
+  add_executable(openshot-test
+    tests.cpp
+    ${OPENSHOT_TEST_FILES} 
+    ${OPENSHOT_CV_TEST_FILES})
```
ferdnyc, line 111:
```suggestion
add_executable(openshot-test
  tests.cpp
  ${OPENSHOT_TEST_FILES} 
  ${OPENSHOT_CV_TEST_FILES}
  )
```
With above.

tests/CMakeLists.txt
```diff
@@ -74,17 +92,35 @@ set(OPENSHOT_TEST_FILES
...
+  
+  # Link libraries to the new executable
+  target_link_libraries(openshot-test openshot ${UnitTest++_LIBRARIES} ${OpenCV_LIBS} ${PROTOBUF_LIBRARY})
```
ferdnyc, line 114:
```suggestion
# Link libraries to the new executable
target_link_libraries(openshot-test
  openshot
  ${UnitTest++_LIBRARIES}
  ${OPENSHOT_CV_LIBRARIES}
  )
```
Again, unconditionally. Really, I'll be wanting to get you set up with targets for all of this stuff (for one thing, that'll make it exportable), but for now this will serve.

tests/CMakeLists.txt
```diff
@@ -74,17 +92,35 @@ set(OPENSHOT_TEST_FILES
...
+else()
+  add_executable(openshot-test
+    tests.cpp
+    ${OPENSHOT_TEST_FILES} )
 
-# Link libraries to the new executable
-target_link_libraries(openshot-test openshot ${UnitTest++_LIBRARIES})
+  # Link libraries to the new executable
+  target_link_libraries(openshot-test openshot ${UnitTest++_LIBRARIES})
+endif()
```
ferdnyc, line 123: And then all of this can be dropped.

tests/CMakeLists.txt
ferdnyc: The other option is to do the configuration parts in multiple stages, like so:
```cmake
add_executable(openshot-test
  tests.cpp
  ${OPENSHOT_TEST_FILES}
  )
target_link_libraries(openshot-test openshot)

if(OpenCV_FOUND)
  target_sources(openshot-test
    CVTracker_Tests.cpp
    CVStabilizer_Tests.cpp
    # CVObjectDetection_Tests.cpp
    )
  target_link_libraries(openshot-test PUBLIC
    ${OpenCV_LIBS}
    ${PROTOBUF_LIBRARY}
    )
endif()
```
That also works perfectly fine, in fact it's slightly cleaner IMHO.

src/CMakeLists.txt
```diff
@@ -375,6 +395,59 @@ if (ENABLE_BLACKMAGIC)
   endif()
 endif()
 
+################## OPENCV ###################
+
+find_package( OpenCV 4 )
+if (OpenCV_FOUND)
+  message("\nCOMPILING WITH OPENCV\n")
+  set(CMAKE_SWIG_FLAGS "-DUSE_OPENCV=1")
```
ferdnyc, line 403: Nope.
```suggestion
  list(APPEND CMAKE_SWIG_FLAGS -DUSE_OPENCV=1)
```
That way you don't blow away all the _other_ SWIG flags.

src/CMakeLists.txt
```diff
@@ -375,6 +395,59 @@ if (ENABLE_BLACKMAGIC)
...
+  add_definitions( -DUSE_OPENCV=1 )
```
ferdnyc, line 404:
```suggestion
  target_compile_definitions(openshot PUBLIC USE_OPENCV=1)
```

src/CMakeLists.txt
```diff
@@ -375,6 +395,59 @@ if (ENABLE_BLACKMAGIC)
...
+  foreach(PROTO_CC ${PROTO_CCS})
+    execute_process(COMMAND ${CMAKE_COMMAND} -E copy ${PROTO_CC} ${CMAKE_CURRENT_SOURCE_DIR})
+    file(REMOVE ${PROTO_CC})
+  endforeach()
+  
+  FILE(GLOB PROTO_HS "${PROTOBUF_MESSAGES_FOLDER}/*.pb.h")
+  foreach(PROTO_H ${PROTO_HS})
+    execute_process(COMMAND ${CMAKE_COMMAND} -E copy ${PROTO_H} ${CMAKE_CURRENT_SOURCE_DIR})
+    file(REMOVE ${PROTO_H})
+  endforeach()
```
ferdnyc, line 439: Modifying files in the _source_ dir is **not** OK during a build.

The CMake Find module for Protobuf has an example of how to use the macros it defines to generate source files:
(From `/usr/share/cmake/Modules/FindProtobuf.cmake` on my machine...)
```cmake
  find_package(Protobuf REQUIRED)
  include_directories(${Protobuf_INCLUDE_DIRS})
  include_directories(${CMAKE_CURRENT_BINARY_DIR})
  protobuf_generate_cpp(PROTO_SRCS PROTO_HDRS foo.proto)
  protobuf_generate_cpp(PROTO_SRCS PROTO_HDRS EXPORT_MACRO DLL_EXPORT foo.proto)
  protobuf_generate_cpp(PROTO_SRCS PROTO_HDRS DESCRIPTORS PROTO_DESCS foo.proto)
  protobuf_generate_python(PROTO_PY foo.proto)
  add_executable(bar bar.cc ${PROTO_SRCS} ${PROTO_HDRS})
  target_link_libraries(bar ${Protobuf_LIBRARIES})

.. note::
  The ``protobuf_generate_cpp`` and ``protobuf_generate_python``
  functions and :command:`add_executable` or :command:`add_library`
  calls only work properly within the same directory.
```
You don't actually need the third line as that's already automatic, and obviously the last two lines should be replaced with configuration on the library instead. I'd also prefer `target_include_directories(openshot PUBLIC ${Protobuf_INCLUDE_DIRS})` to the directory-wide `include_dirs()`.

Using their macros will generate files into the _build_ directory and reference them from there. It'll also add the files to CMake's list of `GENERATED` files to delete during a `make clean`. If the example isn't adaptable to what you need to do, let me know what the issues are and we can figure something out that still maintains the immutability of the source tree.

I'll most likely still want to encapsulate all of this into an `INTERFACE` target, that way it'll automatically be carried with the `openshot` target across directories (you won't have to do anything in `tests/` but add the test source files), and it'll become part of the `EXPORTED` library configs automatically. But I can tackle that part later.

tests/CVTracker_Tests.cpp
ferdnyc, line 58: Specifically, this could be...
```suggestion
        CVTracker kcfTracker(R"proto(
                { "protobuf_data_path": "",
                  "tracker_type": "KCF",
                  "bbox": {"x": 294, "y": 102, "w": 180, "h": 166},
                  "first_frame": 0
                } )proto", processingController);
```
Or, since the line breaks are kinda ugly in an argument,
```suggestion
        const std::string proto_json = R"proto(
                { "protobuf_data_path": "",
                  "tracker_type": "KCF",
                  "bbox": {"x": 294, "y": 102, "w": 180, "h": 166},
                  "first_frame": 0
                } )proto";
        CVTracker kcfTracker(proto_json, processingController);
```

tests/Frame_Tests.cpp
```diff
@@ -147,4 +147,28 @@ TEST(Copy_Constructor)
...
+TEST(Convert_Image)
+{
+	// Create a video clip
+	std::stringstream path;
+	path << TEST_MEDIA_PATH << "sintel_trailer-720p.mp4";
+	Clip c1(path.str());
+	c1.Open();
+
+	// Get first frame
+	std::shared_ptr<Frame> f1 = c1.GetFrame(1);
```
ferdnyc, line 160:
```suggestion
	auto f1 = c1.GetFrame(1);
```
Personally I'm a big proponent of letting the compiler "type" all that crap. :wink:

.gitignore
```diff
@@ -8,3 +8,5 @@
 /.metadata/
 tags
 *~
+
+.vscode/
```
ferdnyc, line 12: You can ignore files/directories locally by adding them to the `.git/info/exclude` file, its contents are merged with the repo `.gitignore` when building the ignore list. There's no reason to add directories to the repo `.gitignore` unless they're likely to be created by something _in_ the repo.

(I mean, it doesn't _hurt_ anything, sure. Just that there's an easy local workaround, only problem is that it's so poorly-communicated it's practically an easter egg. I only learned about it this year, I think.)

ferdnyc: (There's also a "global" (per-user) `$HOME/.config/git/ignore` file, for anything you want to always ignore in all repos forever.)

tests/Frame_Tests.cpp
BrennoCaldato: :)

BrennoCaldato:
@ferdnyc Thanks for the very detailed review. I'll incorporate these and test it on my side :) 
Feel free to report if you find any other problem.

src/CMakeLists.txt
BrennoCaldato: Yeah this was far from ideal. At the time I was having problems including the message headers when I compiled the messages on build folder. I'll take another look at this and follow your suggestion. Thanks.

.gitignore
BrennoCaldato: Good to know about this easter egg. Some git tools include it on the main .gitignore by default

BrennoCaldato:
@ferdnyc I noticed opencv and develop branches for openshot-qt are failing at all build servers.
Do you have any intuition about this?

ferdnyc:
@BrennoCaldato 

Hmm. OK, I see now. The _builds_ are succeeding, it's the artifact upload at the end that's failing.

I don't have the right access to those systems to be sure, only @jonoomph does, but my Sysadmin's Intuition makes me suspect a disk is full, or a filespace quota exceeded... something like that.

```text
b"['/tmp/.mount_J9NeDj/package', 
'/home/gitlab-runner/builds/d0022447/0/OpenShot/openshot-qt/build/OpenShot.AppDir', 
'/home/gitlab-runner/builds/d0022447/0/OpenShot/openshot-qt/build/OpenShot-v2.5.1-dev2-1604625533-2d181430-2cb4eeff-x86_64.AppImage']\n"
Successfully completed build-server script!
Uploading artifacts for successful job
Uploading artifacts...
Runtime platform                                    arch=amd64 os=linux pid=30973 r
evision=6fbc7474 version=13.1.1
build/*.AppImage: found 1 matching files and directories 
WARNING: build/*.torrent: no matching files        
build/install-x64/share/: found 10 matching files and directories 
build/build-server.log: found 1 matching files and directories 
WARNING: Uploading artifacts to coordinator... failed  id=20865 responseStatus=500 
Internal Server Error status=500 Internal Server Error token=ZMEk8H-h
WARNING: Retrying...                                context=artifacts-uploader 
error=invalid argument
WARNING: Uploading artifacts to coordinator... failed  id=20865 responseStatus=500 
Internal Server Error status=500 Internal Server Error token=ZMEk8H-h
WARNING: Retrying...                                context=artifacts-uploader 
error=invalid argument
WARNING: Uploading artifacts to coordinator... failed  id=20865 responseStatus=500 
Internal Server Error status=500 Internal Server Error token=ZMEk8H-h
FATAL: invalid argument                            
ERROR: Job failed: exit status 1
```

ferdnyc:
Just shot @jonoomph an email.

src/CMakeLists.txt
ferdnyc: @BrennoCaldato 

Anytime you add include files to a build, you need to jump through a bunch of hoops. Doubly so if they become public headers, which of course they do if they get referenced in any _other_ public headers. And if those includes are **generated** files, then it gets really complex. 

1. There needs to be a step to generate the files, which has to be done before the build step for anything that includes them. (If the files can be generated _during_ the configure run, that's a lot easier because it eliminates the need to add build-step dependencies.)

1. The generated files need to be installed after the build is done. Which means you need to create a list of all of the files that _will be_ generated, and where they _will be_ on the disk, so the correct install scripts can be generated. You can't (ever) glob for _generated_ files in `CMakeLists.txt` scripting, because those output files **don't exist yet**. 

    In CMake's processing, generated files get produced _after_ the configuration is generated. CMake processes all of the `CMakeLists.txt` files, writes out the control files and scripting for everything specified at the various levels, and **then** executes the file-generation scripts. So at the time the glob is actually performed, there's nothing to find. (That's true whether the files in question are in the source tree or the build tree, and it's why there's never any reason (or need) to generate files into the build tree — it won't change the situation in any way that matters. Using build-tree files during the build is just a matter of adding the right configuration, same as source-tree files.)

1. Because new headers are being created and installed, their _location_ changes between the build context and the install context, so any references to them are contextual. The location of the install tree can also change, so _that_ context also **must** use only relative paths so they can be rooted at an as-yet-undetermined `CMAKE_INSTALL_PREFIX`. Otherwise, you end up with a build that succeeds, but then installs broken headers or configuration.

It can be a pain in the ass to get right, but that's why CMake includes a set of macros in their Protobuf configuration (the Find module) that should take care of most of the tedium automatically.

ferdnyc: > and it's why there's never any reason (or need) to generate files into the build tree

Er, I meant into the \*source tree.

ferdnyc: One of the reasons Kitware (the CMake developers) recommend so strongly against globbing for files _period_ in `CMakeLists.txt` scripts is, it's easy to trick yourself into thinking something works, and sending yourself down wrong paths based on that apparent success.

If a `CMakeLists.txt` script uses `file(GLOB...` commands, it'll work fine if at some point you managed to get the necessary files generated beforehand. Maybe a previous run partly succeeded, maybe you generated them by hand, whatever. Regardless how, they're there, so now CMake has something to find when it's _**re**-generating_ the configuration. It'll succeed as long as it's run in in a "dirty" tree. But _only_ in a dirty tree. Later, you go to test your configuration in a _clean_ tree, and surprise! Suddenly it's failing again. So then you start adding more complexity to try to "fix" the globs.

Stop me if any of this is sounding familiar. :wink:  Been there, done that, plenty of times. Finally learned my lesson, but only after far too much time wandering down those wrong paths.

The reality is that there's no way to "fix" the globs no matter how hard we try. They can't ever work that way, and the only real fix is to find a way to do things that doesn't require them.

BrennoCaldato: @ferdnyc 

I forgot to answer this message, sorry. 
The protobuf generated files issue was fixed some commits ago. I created a library for the proto messages and the files are built outside of src now.

Thanks again for the detailed explanation. Aways very informative.

ferdnyc:
@BrennoCaldato 

I got a notification about this PR, but — the magic of GitHub at work — I can't seem to find what actually _triggered_ the notification! It just dumps me at the top of the page when I click it from my notifications list. Stupid thing.

Anyway, though, I looked through your build configs here, and reworked a bunch of things to modernize the CMake and use build targets to avoid duplicated code — what I have now builds successfully, at least on my system, and runs the unit tests, including the new OpenCV ones. Those output a few "Error:" messages, and then pass anyway... bit disconcerting, that.

Anyway, the changes are substantial enough that I'm going to open a new PR _against_ the `opencv` branch (so, a PR on the PR)... we can pick this up from over there if need be, once we have the code in front of us. Gimme like 2-3 minutes...

src/effects/Stabilizer.h
```diff
@@ -0,0 +1,137 @@
...
+	private:
+		/// Init effect settings
+		void init_effect_details();
+		std::string protobuf_data_path;
+		Keyframe zoom;
+
+	public:
+	std::string teste;
+        std::map <size_t,EffectCamTrajectory> trajectoryData; // Save camera trajectory data
+    	std::map <size_t,EffectTransformParam> transformationData; // Save transormation data
```
ferdnyc, line 100: You've got a bunch of mixed-spaces-and-tabs damage here. The existing code in this file mostly uses tab indentation currently, personally I prefer four spaces (2 spaces for things like `CMakeLists.txt`, `.js` and `.cs` code, etc.) but it's best to follow the surrounding format when editing existing source.

Since this is a new file, I'm guessing it started as a copy-and-update of one of the other effects headers? In which case, I'm fine with either reformatting the entire thing with space indents, or continuing to use tabs everywhere. But whichever way you go, just pick a lane. :wink:

src/effects/Stabilizer.h
```diff
@@ -0,0 +1,137 @@
...
+		///
+		/// The frame object is passed into this method, and a frame_number is passed in which
+		/// tells the effect which settings to use from its keyframes (starting at 1).
+		///
+		/// @returns The modified openshot::Frame object
+		/// @param frame The frame object that needs the effect applied to it
+		/// @param frame_number The frame number (starting at 1) of the effect on the timeline.
+		std::shared_ptr<Frame> GetFrame(std::shared_ptr<Frame> frame, int64_t frame_number) override;
+
+		std::shared_ptr<openshot::Frame> GetFrame(int64_t frame_number) override { return GetFrame(std::shared_ptr<Frame> (new Frame()), frame_number); }
```
ferdnyc, line 119:
```suggestion
		std::shared_ptr<openshot::Frame> GetFrame(int64_t frame_number) override {
			return GetFrame(std::make_shared<openshot::Frame>(), frame_number);
		};
```

It's preferable to use `std::make_shared` when allocating new shared pointers. (And it's more concise, as a bonus.) 

Also, ultra-long lines make FeRD cry. :sob:  My rule of thumb: anytime GitHub starts horizontally scrolling, you done gone too far to the right.

src/effects/Stabilizer.h
```diff
@@ -0,0 +1,137 @@
...
+
+		// Load protobuf data file
+        bool LoadStabilizedData(std::string inputFilePath);
```
ferdnyc, line 122:
```suggestion
		bool LoadStabilizedData(std::string inputFilePath);
```
More tabs-vs-spaces.

src/effects/Stabilizer.h
ferdnyc, line 100:
```suggestion
		std::string teste;
		std::map <size_t,EffectCamTrajectory> trajectoryData; // Save camera trajectory data
		std::map <size_t,EffectTransformParam> transformationData; // Save transormation data 
```

ferdnyc: The new `struct`s above are all consistently space-indented. Because they are consistent already, it's fine to leave those as they are, even if the rest of the file continues to be tab-indented. As long as each block is consistent within itself, I don't actually have any objections to mixing within the same file. (Others may disagree, and take a more hard-line approach.)

src/effects/Tracker.h
```diff
@@ -0,0 +1,134 @@
...
+	class Tracker : public EffectBase
+	{
+	private:
+		/// Init effect settings
+		void init_effect_details();
+		std::string protobuf_data_path;
+
+	public:
+
+        std::map<int, EffectFrameData> trackedDataById; // Save object tracking box data
```
ferdnyc, line 95:
```suggestion
		std::map<int, EffectFrameData> trackedDataById; // Save object tracking box data
```

src/effects/Tracker.h
```diff
@@ -0,0 +1,134 @@
...
+		/// @param frame The frame object that needs the effect applied to it
+		/// @param frame_number The frame number (starting at 1) of the effect on the timeline.
+		std::shared_ptr<Frame> GetFrame(std::shared_ptr<Frame> frame, int64_t frame_number) override;
+		std::shared_ptr<openshot::Frame> GetFrame(int64_t frame_number) override { return GetFrame(std::shared_ptr<Frame> (new Frame()), frame_number); }
+		
+		// Load protobuf data file
+        bool LoadTrackedData(std::string inputFilePath);
+        
+		// Get tracker info for the desired frame 
+        EffectFrameData GetTrackedData(size_t frameId);
```
ferdnyc, line 119:
```suggestion
		// Load protobuf data file
		bool LoadTrackedData(std::string inputFilePath);
        
		// Get tracker info for the desired frame 
		EffectFrameData GetTrackedData(size_t frameId);
```

src/sort_filter/KalmanTracker.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+	measurement = Mat::zeros(measureNum, 1, CV_32F);
+
+	kf.transitionMatrix = (Mat_<float>(stateNum, stateNum) << 1, 0, 0, 0, 1, 0, 0,
+						   0, 1, 0, 0, 0, 1, 0,
+						   0, 0, 1, 0, 0, 0, 1,
+						   0, 0, 0, 1, 0, 0, 0,
+						   0, 0, 0, 0, 1, 0, 0,
+						   0, 0, 0, 0, 0, 1, 0,
+						   0, 0, 0, 0, 0, 0, 1);
```
ferdnyc, line 28: And, this is _why_ I personally hate tab indents. They only work if you always switch to spaces for **alignment** indentation.

Then again, I also hate blocks of code aligned way to the right side of the screen. IMHO this should be:
```suggestion
	kf.transitionMatrix = (Mat_<float>(stateNum, stateNum)
		<< 1, 0, 0, 0, 1, 0, 0,
		   0, 1, 0, 0, 0, 1, 0,
		   0, 0, 1, 0, 0, 0, 1,
		   0, 0, 0, 1, 0, 0, 0,
		   0, 0, 0, 0, 1, 0, 0,
		   0, 0, 0, 0, 0, 1, 0,
		   0, 0, 0, 0, 0, 0, 1);
```
And then it doesn't actually matter if you use spaces or tabs to indent.

ferdnyc, line 28: Other alternative:

```suggestion
	kf.transitionMatrix = (
		Mat_<float>(stateNum, stateNum)
		<< 1, 0, 0, 0, 1, 0, 0,
		   0, 1, 0, 0, 0, 1, 0,
		   0, 0, 1, 0, 0, 0, 1,
		   0, 0, 0, 1, 0, 0, 0,
		   0, 0, 0, 0, 1, 0, 0,
		   0, 0, 0, 0, 0, 1, 0,
		   0, 0, 0, 0, 0, 0, 1
	);
```

...I don't have any strong opinions on which one is more readable.

src/sort_filter/KalmanTracker.cpp
```diff
@@ -0,0 +1,117 @@
...
+void KalmanTracker::init_kf(
+	StateType stateMat)
+{
+	int stateNum = 7;
+	int measureNum = 4;
+	kf = KalmanFilter(stateNum, measureNum, 0);
+
+	measurement = Mat::zeros(measureNum, 1, CV_32F);
+
+	kf.transitionMatrix = (Mat_<float>(stateNum, stateNum) << 1, 0, 0, 0, 1, 0, 0,
```
ferdnyc, line 22: Code-resiliency question here: If you're hardcoding the initialization of the matrix, shouldn't you hardcode the dimension as well? If someone changes `stateNum` to, say, `5` this will break, right? Because it's hardcoded to be `7x7`. I think if you're going to hardcode the matrix for a specific size, it's actually cleaner to hardcode that size as well:
```suggestion
	kf.transitionMatrix = (Mat_<float>(7, 7) << 1, 0, 0, 0, 1, 0, 0,

```
...Otherwise, best to have the code generate a properly-dimensioned matrix programmatically, based on the value of `stateNum`. Maintaining a variable separate from the hardcoded initialization based on the (assumed) value of that variable feels like a maintenance headache.

ferdnyc: (`stateNum` can be set to `7` _after_ the hardcoded matrix initialization, of course. As long as there's a comment noting that its value needs to be equal to the size of either dimension of the square `transitionMatrix`.)

src/Frame.h
```diff
@@ -278,6 +291,18 @@ namespace openshot
 
 		/// Play audio samples for this frame
 		void Play();
+
+#ifdef USE_OPENCV
+		/// Convert Qimage to Mat
+		cv::Mat Qimage2mat( std::shared_ptr<QImage>& qimage);
+		std::shared_ptr<QImage> Mat2Qimage(cv::Mat img);
```
ferdnyc, line 298: This method should have a doc string too, since it'll be used in the generated API docs.

```suggestion

		/// Convert OpenCV Mat to QImage
		std::shared_ptr<QImage> Mat2Qimage(cv::Mat img);
```

src/Frame.h
```diff
@@ -114,6 +123,10 @@ namespace openshot
 		std::string color;
 		int64_t max_audio_sample; ///< The max audio sample count added to this frame
 
+#ifdef USE_OPENCV
+		cv::Mat imagecv; ///< OpenCV image. It will be always on BGR format
```
ferdnyc, line 127: Super-minor tweak:

```suggestion
		cv::Mat imagecv; ///< OpenCV image. It will always be in BGR format
```

src/effects/ObjectDetection.cpp
```diff
@@ -0,0 +1,285 @@
...
+	info.has_video = true;
+}
+
+// This method is required for all derived classes of EffectBase, and returns a
+// modified openshot::Frame object
+std::shared_ptr<Frame> ObjectDetection::GetFrame(std::shared_ptr<Frame> frame, int64_t frame_number)
+{
+    // Get the frame's image
+	cv::Mat cv_image = frame->GetImageCV();
+    std::cout<<"Frame number: "<<frame_number<<"\n\n";
```
ferdnyc, line 75: This should be taken out, before merging.

src/effects/ObjectDetection.cpp
```diff
@@ -0,0 +1,285 @@
...
+                cv::Rect_<float> bb_nrml = detections.boxes.at(i);
+                cv::Rect2d box((int)(bb_nrml.x*fw),
+                               (int)(bb_nrml.y*fh),
+                               (int)(bb_nrml.width*fw),
+                               (int)(bb_nrml.height*fh));
+                drawPred(detections.classIds.at(i), detections.confidences.at(i),
+                         box, cv_image);
+            }
+        }
+    }
```
ferdnyc, line 95: You can early-return here, instead.
```suggestion
    if(cv_image.empty()){
        return frame;
    }

    // Check if track data exists for the requested frame
    if (detectionsData.find(frame_number) != detectionsData.end()) {
        float fw = cv_image.size().width;
        float fh = cv_image.size().height;

        DetectionData detections = detectionsData[frame_number];
        for(int i = 0; i<detections.boxes.size(); i++){
            cv::Rect_<float> bb_nrml = detections.boxes.at(i);
            cv::Rect2d box((int)(bb_nrml.x*fw),
                           (int)(bb_nrml.y*fh),
                           (int)(bb_nrml.width*fw),
                           (int)(bb_nrml.height*fh));
           drawPred(detections.classIds.at(i), detections.confidences.at(i),
                     box, cv_image);
        }
    }
```

src/effects/ObjectDetection.cpp
```diff
@@ -0,0 +1,285 @@
...
+    // Iterate over all frames of the saved message
+    for (size_t i = 0; i < objMessage.frame_size(); i++) {
+        // Create protobuf message reader
+        const libopenshotobjdetect::Frame& pbFrameData = objMessage.frame(i);
+
+        // Get frame Id
+        size_t id = pbFrameData.id();
+
+        // Load bounding box data
+        const google::protobuf::RepeatedPtrField<libopenshotobjdetect::Frame_Box > &pBox = pbFrameData.bounding_box();
```
ferdnyc, line 166: I've gotta say, these names are kind of hard on the eyes. What do you think of just `pb_objdetect`, `pb_tracker`, and `pb_stabilize`? 

(Commit da02180c79148ffcae27cfdddcfae10e10093799 in my fork actually already does that across the tree, so you don't have to go through all that if you decide to make the change. If not, totally fine by me. I just wanted to make sure there weren't any problems before I suggested it.)

src/effects/Tracker.h
```diff
@@ -0,0 +1,133 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with OpenShot Library. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef OPENSHOT_TRACKER_EFFECT_H
+#define OPENSHOT_TRACKER_EFFECT_H
+
+#include "../EffectBase.h"
+
+#include <google/protobuf/util/time_util.h>
```
ferdnyc, line 36: Could this be moved to the `Tracker.cpp` file (along with the `using google::protobuf::util::TimeUtil;` farther down)? Same for the other headers that include it.

I don't see any actual use of those symbols in the headers (unless I'm missing it), and SWIG is complaining because it doesn't know what `google::protobuf::util::TimeUtil` is. (Since it doesn't get those headers.)

tests/CVTracker_Tests.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+        // Track clip for frames 0-20
+        kcfTracker.trackClip(c1, 0, 20, true);
+        // Get tracked data
+        FrameData fd = kcfTracker.GetTrackedData(20);
+        float x = fd.x1;
+        float y = fd.y1;
+        float width = fd.x2 - x;
+        float height = fd.y2 - y;
+        std::cout<<"\n\n Error: "<< processingController.GetErrorMessage() <<"\n";
```
ferdnyc, line 78:
```suggestion

```

tests/CVTracker_Tests.cpp
```diff
@@ -0,0 +1,151 @@
...
+        kcfTracker_2._LoadTrackedData();
+
+        // Get tracked data
+        FrameData fd_2 = kcfTracker_2.GetTrackedData(20);
+
+        float x_2 = fd_2.x1;
+        float y_2 = fd_2.y1;
+        float width_2 = fd_2.x2 - x_2;
+        float height_2 = fd_2.y2 - y_2;
+        std::cout<<"\n\n Error: "<< processingController.GetErrorMessage() <<"\n";
```
ferdnyc, line 143:
```suggestion

```

ferdnyc (changes requested):
@BrennoCaldato 
This'll get rid of my `Error:` messages. (Still not sure how you're not seeing those!)

ferdnyc:
@BrennoCaldato 

If you have a chance you may want to look over the [Codacy scan report](https://app.codacy.com/gh/OpenShot/libopenshot/pullRequest?prid=6412440) (that's its current URL, it may update with changes to the PR), it tends to flag a lot of stuff that's not fatal, and easy to overlook, but worth addressing if possible.

ferdnyc:
Re: Codacy, I see it's still flagging some stuff in `examples/Example_opencv.cpp` that isn't necessarily a problem there. I could mask the `examples/` directory from its checks, as the same coding standards aren't entirely applicable (or even desirable) when it comes to example code.

BrennoCaldato:
Hi @ferdnyc 

I opted for leaving the `examples/Example_opencv.cpp` the way it is for now just because it's easier to test a couple of things. But I can later use arg parser to set the parameters.

If you prefer to remove this file from codacy it's also fine for me.

src/sort_filter/Hungarian.cpp
```diff
@@ -0,0 +1,454 @@
...
+	int nOfRows,
+	int nOfColumns)
+{
+	double *distMatrix, *distMatrixTemp, *distMatrixEnd, *columnEnd, value, minValue;
+	bool *coveredColumns, *coveredRows, *starMatrix, *newStarMatrix, *primeMatrix;
+	int nOfElements, minDim, row, col;
+
+	/* initialization */
+	*cost = 0;
+	for (row = 0; row < nOfRows; row++)
```
ferdnyc, line 69: @BrennoCaldato Regarding Codacy's feedback on things like the scope of loop variables — I see it as a purely stylistic choice, ultimately, and in practice there will be no appreciable difference in performance between the two. However, I tend to side with the voices in [this discussion](http://www32.cplusplus.com/forum/general/116308/) who say that, _if anything_, multiple for loops that each use a separate, locally-scoped variable may run slightly _faster_ than reusing the same variable allocated once at the beginning.

That's mostly because, if we're talking about a POD variable (like a loop counter), the compiler can, and will, just store that counter in a register. Which means it won't require _any_ allocations at all. It's probably smart enough that it's going to do that **regardless** when optimization is turned on, but the version with locally-scoped variables gives it a better chance of optimizing away the allocations even at lower optimization levels.

See also Matt Godbolt's talk from CppCon 2017, where he shows off Compiler Explorer and uses it to examine how various styles of C++ loop code get translated into assembly instructions at different optimization levels: 

https://youtu.be/bSkpMdDe4g4

(The entire talk is really worth watching if you have time, it's all amazing. At around 35 minutes in, there's a segment where he demonstrates some of the tricks the compiler uses to optimize adds and multiplies, as well as to _avoid_ divide operations — that really blew me away.)

So, long story short, take these two bits of otherwise-identical code:

```c++
// Version 1
int row;
for (row = 0; row < ...; ++row) {
  ...
}
for (row = 0; ...) {
  ...
}
for (row = 0; ...) {
  ...
}
```
```c++
// Version 2
for (int row = 0; row < ...; ++row) {
  ...
}
for (int row = 0; ...) {
  ...
}
for (int row = 0; ...) {
  ...
}
```
In practice, all other things being equal:
* The difference in terms of real-world performance will be too small to measure, in basically all scenarios.
* With optimization turned on, both versions will most likely require _zero_ memory allocations for `row`.
* Therefore there's no practical advantage to Version 1 over the (IMHO) cleaner and more readable Version 2
* Codacy is therefore (again IMHO) correct to recommend rewriting instances of Version 1 as Version 2, where it can catch them.
* The same holds for instances that Codacy _can't_ catch. (It recommended reducing the scope of your `col`, but not `row`, even though `row` is never used outside of the loops either.)

But, as I said, ultimately I consider it a stylistic choice and I'm not hung up on imposing one or the other in the libopenshot code.

jonoomph:
@BrennoCaldato @ferdnyc I'm getting ready to merge this into libopenshot. I think we have the basic integration very well done at this point, and it opens up a tremendous amount of future effects and features we can based off of OpenCV functions. I still think we have a few things to tweak on the openshot-qt side... UI wise, and I might hide the "Object Detector" effect initially, since it is mostly useless until future Keyframe improvements. But this is a great start!!

src/sort_filter/Hungarian.cpp
BrennoCaldato: @ferdnyc This is very interesting. I didn't have time to watch the video yet but I'll definitely do. Thanks for sending it.

jonoomph:
libopenshot welcomes OpenCV and all the potential it brings!

PR #586, opened 2020-11-03
Various unit tests: Increase coverage

ferdnyc:
**Update**: I'm an idiot. The failure was due to the short duration, and the fact that there were only 120 frames _available_ in the reader at the specified frame rate. (Though, interestingly, that _doesn't_ throw an `OutOfBoundsFrame` it seems.)

<s>As it says, this PR is a work-in-progress. Not because it's unfinished, but because a test is failing.</s>

I decided to replace the first test (which had a `catch` in it that was screwing up the test coverage) with a simple no-op test: Create a mapping from 24fps to 24fps with no pulldown. Then read its `Name()` (adds an extra line of coverage), and finally read one frame from it. So, it just gets the same frame back. (When you use the right duration.)

<s>Unfortunately, **the test _fails_**, and gets back frame `120` for original frame `125`.
That seems bad.</s>

**Edit**: Added a JSON test that just reads the `Json()` string from the mapper, then `SetJson()`s it back again (and spot-checks that things haven't changed.) That _really_ raised the test coverage; I may do that for the other classes as well.

### DIFF
--- a/tests/FrameMapper_Tests.cpp
+++ b/tests/FrameMapper_Tests.cpp
@@ -31,33 +31,35 @@
 #include "UnitTest++.h"
 // Prevent name clashes with juce::UnitTest
 #define DONT_SET_USING_JUCE_NAMESPACE 1
-#include "OpenShot.h"
+#include "CacheMemory.h"
+#include "Clip.h"
+#include "DummyReader.h"
+#include "FFmpegReader.h"
+#include "Fraction.h"
+#include "Frame.h"
+#include "FrameMapper.h"
+#include "Timeline.h"
 
 using namespace std;
 using namespace openshot;
 
-TEST(FrameMapper_Get_Valid_Frame)
+SUITE(FrameMapper) {
+TEST(NoOp_GetMappedFrame)
 {
 	// Create a reader
 	DummyReader r(Fraction(24,1), 720, 480, 22000, 2, 5.0);
 
-	// Create mapping between 24 fps and 29.97 fps using classic pulldown
-	FrameMapper mapping(&r, Fraction(30000, 1001), PULLDOWN_CLASSIC, 22000, 2, LAYOUT_STEREO);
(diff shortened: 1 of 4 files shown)


tests/FrameMapper_Tests.cpp
```diff
@@ -31,33 +31,35 @@
...
-	}
-	catch (OutOfBoundsFrame &e)
-	{
-		// Unexpected failure to find frame
-		CHECK(false);
-	}
+	// Should find this frame
+	MappedFrame f = mapping.GetMappedFrame(125);
+	CHECK_EQUAL(125, f.Odd.Frame);
+	CHECK_EQUAL(125, f.Even.Frame);
```
ferdnyc, line 59: I'm an idiot. The problem is that at 24fps, there are only 120 frames in the reader.

```suggestion
	MappedFrame f = mapping.GetMappedFrame(100);
	CHECK_EQUAL(100, f.Odd.Frame);
	CHECK_EQUAL(100, f.Even.Frame);
```

jonoomph:
LGTM!

jonoomph:
Merge when ready!

ferdnyc:
I think I've run out of steam on this one for now. Just made a change to `codecov.yml` to hopefully get `/examples` out of the coverage report, and assuming that passes I'll merge. Coverage up from 49.90% (possibly higher, albeit artificially, if that config change takes), and newly 100% in three different files. I'd say that's a reasonable bump for one sortie.

ferdnyc:
(My attempt to exclude `examples/` was... well, honestly, I'm not _sure_ if it was successful.)

The two files that get built into targets by CMake still show up in the list, but [Codecov doesn't seem to be counting any of their source lines](https://codecov.io/gh/OpenShot/libopenshot/tree/develop/examples) for coverage purposes, which I guess is what really matters. Was hoping to get them off the **list** entirely, but meh...

PR #592, opened 2020-11-23
WIP: Increase Clip test coverage

ferdnyc:
This is on hold pending the merge of #635 , since the test changes here will have to be ported to Catch2.

In case anyone was doubting the value of unit tests, I've found three bugs in the Clip code so far.

### 1: Nullability error

The first, I found while writing tests to cover more failure cases (exceptions thrown, etc.). I added a set of tests, and every time they tried to run, the test program crashed before completion. Turned out, it was impossible to null a clip's Reader using the setter:
```c++
openshot::Clip c("video.mp4");
c.Open();
c.Reader(nullptr);
```
The third line would crash with an unhandled exception, because the code for `Reader(openshot::ReaderBase* new_reader)` tried to make calls to the reader immediately after setting the member variable to the new pointer value, without first checking whether `reader` was valid. Fixed.

### 2: Logic error

The second one, I found just by reading the code on Codecov's website, because it's in the only one of four conditional paths that we _don't_ currently have a test for:

image

Do you see it? The last condition should be
```c++
else if (enabled_video == -1 && reader && !reader->info.has_video)
    enabled_video = 0;
```

As a result, it was impossible to use auto-enabled video on a video file with no audio, or the video would auto-disable itself. Fixed.

### 3: Confusing order of operations

The third one isn't really a _bug_ per se, but it cleans up a very confusing logical flow in what should be a simple function. The code for `Clip::End` used to read:
```c++
float Clip::End() const
{
        // if a time curve is present, use its length
        if (time.GetCount() > 1)
        {
                // Determine the FPS fo this clip
                float fps = 24.0;
                if (reader)
                        // file reader
                        fps = reader->info.fps.ToFloat();
                else
                        // Throw error if reader not initialized
                        throw ReaderClosed("No Reader has been initialized for this Clip.  Call Reader(*reader) before calling this method.");

                return float(time.GetLength()) / fps;
        }
        else
                // just use the duration (as detected by the reader)
                return end;
}
```

Setting aside the unnecessary `else` conditions that can be streamlined with early returns, the innermost if/else also has a `throw` in the `else`, which means it'll never get to the code after the else when the `if` condition is false. Which in turn means the condition doesn't need to be there at all, the `throw` serves as an early return.

All in all, after spinning everything the right way around, `Clip::End` is now reduced to:
```c++
// Get end position of clip (trim end of video), which can be affected by the time curve.
float Clip::End() const
{
        if (!reader) {
                // Throw error if reader not initialized
                throw ReaderClosed(
                        "No Reader has been initialized for this Clip. "
                        "Call Reader(*reader) before calling this method.");
        }

        if (time.GetCount() <= 1) {
                // if no time curve, just use the duration (as detected by the reader)
                return end;
        }

        float fps = reader->info.fps.ToFloat();
        return float(time.GetLength()) / fps;
}
```

(This third one actually has nothing to do with unit tests, it's just something I spotted while I was in the code. One nice thing about writing tests is that it makes you _read_ the code, instead of _write_ it, which is a much better way of evaluating it objectively.)

I also used early returns to unspool a bunch of other functions that looked like this:
```c++
void Clip::DoSomething() {
    if (reader)
    {
        // Tons and TONS of code
    
        // Like, more than would fit on one screen
    
        // No matter how big your screen is

    }
}

/* That should of course read: */
void Clip::DoSomething() {
    if (!reader)
        return;
    // All the other stuff
}
```

I'm starting to develop this weird theory that 16:9 aspect screens are causing programmers to be more cavalier about deeply nested functions, because their screens  have _more_  width than height at this point.  It certainly seem like the art of the early return, and the preference for many separate, succinct functions over functions containing long chains of conditional logic, have been falling by the wayside in recent years.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@
 # along with OpenShot Library. If not, see <http://www.gnu.org/licenses/>.
 ################################################################################
 
-cmake_minimum_required(VERSION 3.2...3.14 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.12...3.20 FATAL_ERROR)
 
 message("\
 -----------------------------------------------------------------


src/Clip.cpp
```diff
@@ -245,193 +250,199 @@ Clip::~Clip()
...
-		if (reader)
-			// file reader
-			fps = reader->info.fps.ToFloat();
-		else
-			// Throw error if reader not initialized
-			throw ReaderClosed("No Reader has been initialized for this Clip.  Call Reader(*reader) before calling this method.");
-
-		return float(time.GetLength()) / fps;
+	if (!reader) {
+		// Throw error if reader not initialized
```
ferdnyc, line 311:
```suggestion
```

Most of these offensively-redundant comments, I took out if I was touching the code around them, but I left this one in for some reason. Does someone think we can't _read_ the text directly below it!?

src/Clip.cpp
```diff
@@ -245,193 +250,199 @@ Clip::~Clip()
...
-		throw ReaderClosed("No Reader has been initialized for this Clip.  Call Reader(*reader) before calling this method.");
+
+	// Apply keyframe / transforms
+	apply_keyframes(frame, width, height);
+
+	// Cache frame
+	cache.Add(frame);
+
+	// Return processed 'frame'
+	return frame;
```
ferdnyc, line 445:
```suggestion
	cache.Add(frame);
	return frame;
```

No idea why I left these, either.

ferdnyc:
> All in all, after spinning everything the right way around, `Clip::End` is now reduced to:

...The wrong thing, since it now strikes me that the `time.GetLength()` check needs to come _before_ the `reader` check. I'll fix that.

ferdnyc:
Quit being a pushy jerk, Codecov, I'm _**working** on it_! (Sheesh! :wink: )

jonoomph:
> Hah! But the new version of the function does break several of the existing tests, which are clearly assuming that they _can_ get `End()` for a clip without a reader. I'll have to figure out if that's a reasonable assumption to be making or not, maybe the comment is wrong.

I was going to mention that the previous function always return the End() value regardless of the reader. I think that might be more of a convenience for unit tests.. vs a real solution to anything.

ferdnyc:
@jonoomph Yeah, I ended up just surrendering and changing `::End()` to:
```c++
// Get end position of clip (trim end of video), which can be affected by the time curve.
float Clip::End() const
{
        // if no time curve, just use the current duration
        if (time.GetCount() <= 1)
                return end;

        if (!reader) {
                throw ReaderClosed(
                        "No Reader has been initialized for this Clip. "
                        "Call Reader(*reader) before calling this method.");
        }

        float fps = reader->info.fps.ToFloat();
        return float(time.GetLength()) / fps;
}
```

Which doesn't make quite as much _sense_ to me (still odd to have an end-frame without a reader), but it exactly mirrors the old behavior and fixes the unit tests.

jonoomph:
This LGTM! Nice work as usual!

ferdnyc:
@jonoomph Thanks, I think I'm pretty close to merging it. I just want to add a couple more tests around time curves and frame remapping.

(I don't think I'm going to delve into audio resampling, at least not on this round. Nearly 3% coverage increase is pretty good for one push, no need to go nuts.)

For the remapping, I just have to figure out the best way to do that. It may be time to revive my frame-counter test file. (25 frames, each showing the frame number in a huge font that practically fills the frame, atop an ugly pinkish background on frames 1-24, and a yellow background on frame 25.) But it could be even more useful with one extra twist: If I give _each_ frame a **uniquely** colored background, then it's easy to use `CheckPixel()` to work out the original frame# of the frame you just got handed by a call to `GetFrame()`.

ferdnyc:
I have _no idea_ what I just did (well, I do, I screwed  up a `git rebase`), but apparently I wiped out all of the changes here! Now I get to dig around in the git reflog to try and recover everything. Ugh. **#FML**

PR #593, opened 2020-11-27
WIP: New stream-based logging, macros for ZmqLogger

ferdnyc:
## Update

Somehow I'm an _idiot_ and totally missed that OpenShot needs the ZmqLogger class visible in the public API, because it's how `logger_libopenshot.py` sets up debug logging. So, on hold pending a solution to that issue. (And because there are other, more pressing things to focus on right now.)

(Actually, I should really break out the Fraction, Point, Coordinate, etc. stream operator templates, which _could_ still go in, and I think are probably worth having independent of logging changes.)

## Original PR

`ZmqLogger` is great, but `AppendDebugMethod` is not. Its use is, to say the least, clunky.

This PR adds some convenience macros to make using the logger more convenient, and adds a new stream-based logging alternative.

### The `ZmqLogger`  macros

* `DebugLog()` is a macro to call `AppendDebugLog()`. It takes care of the ZmqLogger instancing.
* `DEBUGVAR()` is an argument-expansion macro, for the common case of wanting to log a variable name followed by the value.

Used together, they allow cleaner and DRY-er code:
```c++
// Instead of writing this
ZmqLogger::Instance()->AppendDebugMethod("AudioReaderSource::getNextAudioBlock", 
    "number_to_copy", number_to_copy,
    "buffer_samples", buffer_samples,
    "buffer_channels", buffer_channels,
    "info.numSamples", info.numSamples,
    "speed", speed,
    "position", position);

// You can write this (it expands to the exact same thing)
DebugLog("AudioReaderSource::getNextAudioBlock",
    DEBUGVAR(number_to_copy),
    DEBUGVAR(buffer_samples),
    DEBUGVAR(buffer_channels),
    DEBUGVAR(info.numSamples),
    DEBUGVAR(speed),
    DEBUGVAR(position));
```

### The `StreamLog` class and macros

An even more convenient alternative is the `StreamLog`, which allows stream-based logging to the `ZmqLogger` instance. It has a `LOGVAR()` macro which acts like `DEBUGVAR()` but for stream output, and a `zmqLog()` macro to access the stream. Use them like this:
```c++
zmqLog() << "AudioReaderSource::getNextAudioBlock " << LOGVAR(number_to_copy)
        << ", " << LOGVAR(buffer_samples) << ", " << LOGVAR(buffer_channels)
        << ", " << LOGVAR(info.numSamples) << ", " << LOGVAR(speed)
        << ", " << LOGVAR(position);
```
That might output something like:
```text
AudioReaderSource::getNextAudioBlock number_to_copy = 1, buffer_samples = 120, 
buffer_channels = 2, info.numSamples = 1500, speed = 0, position = 120
```
Unlike AppendDebugMethod, formatting of values is _not_ forced to four decimal places, instead the default stream formatting is used. It can be adjusted with the usual stream manipulators if desired.

Commas and spacing are also not automatic, though perhaps something like the automatic space insertion from `qDebug()` would be useful.

### Templated `operator<<` for Coordinate, Fraction, and Point
To facilitate use of stream logging, several classes have been enhanced with `<<` operator templates for any stream-based output class. 
* `Fraction` will output a string such as `Fraction(30000, 1001)`
* `Coordinate` will output a string like `(0.5, 1)`
* `Point` will output a string based on its interpolation type:
    * `co(10, 5) interpolation(LINEAR)` or
    * `co(5, 2.5) interpolation(CONSTANT)` or
    * `co(20, 0.4) interpolation(BEZIER) handle_left(0.5, 1) handle_right(1, 2.5)`

Unit tests are included for the StreamLog class and all `operator<<` templates.

#### Other changes
* Because there was relatively little logging in `AudioReaderSource`, it was switched over to `DebugLog()` and `DEBUGVAR()`.
* The log message `Timeline::GetFrame (Cached frame found)` in `Timeline::GetFrame()` now uses `zmqLog()`. (Just that one message, as an experiment.)
* `openshot::Settings` now has a `GetSettings()` convenience macro to access the instance. 
    ```c++
    // Instead of
    name = openshot::Settings::Instance()->PLAYBACK_AUDIO_DEVICE_NAME;
    // You can write
    name = GetSettings()->PLAYBACK_AUDIO_DEVICE_NAME;
    ```
* An environment variable `LIBOPENSHOT_DEBUG` can now be set to any value, to activate the `DEBUG_TO_STDERR` setting. This makes it convenient to turn on debug output even when running e.g. the unit tests:
    `LIBOPENSHOT_DEBUG=1 cmake --build build --target test`
* A lot of header includes were cleaned up, making use of forward declarations and moving includes from `.h` files to `.cpp` files to reduce compilation times/complexity.
* Because I've removed it from all other header files, and because the `StreamLog` stuff will confuse the bindings, `ZmqLogger.h` is no longer wrapped by SWIG and can be treated as a private header. (Perhaps it should even be removed from the set of installed headers?)
* Oh, almost forgot to mention: Among the changes I made to ZmqLogger was to replace the `juce::CriticalSection` with a `std::recursive_mutex`. They have virtually identical semantics (in fact, Jules is thinking of deprecating `CriticalSection` and just telling people to use `recursive_mutex`), and replacing them eliminated the `ZmqLogger` dependency on JUCE which really made no sense at all.
    In files where we're using JUCE code _anyway_ (particularly the classes based on `juce::Thread`), there's no harm in using their mutexes as well. But in any places where the use of `CriticalSection` is the **only** reason `JuceHeader.h` is included (like `ZmqLogger`), it makes a ton of sense to switch over to the STL mutexes instead.

### DIFF
--- a/bindings/python/openshot.i
+++ b/bindings/python/openshot.i
@@ -214,7 +214,7 @@
 	}
 	const std::string __repr__() {
 		std::ostringstream result;
-		result << "Fraction(" << $self->num << ", " << $self->den << ")";
+		result << $self;
 		return result.str();
   }
 	/* Implement dict methods in Python */
--- a/src/AudioReaderSource.cpp
+++ b/src/AudioReaderSource.cpp
@@ -30,6 +30,7 @@
 
 #include "AudioReaderSource.h"
 #include "Exceptions.h"
+#include "ZmqLogger.h"
 
 using namespace std;
 using namespace openshot;
@@ -66,8 +67,8 @@ void AudioReaderSource::GetMoreSamplesFromReader()
 		amount_remaining = 0;
 	}
 
-	// Debug
-	ZmqLogger::Instance()->AppendDebugMethod("AudioReaderSource::GetMoreSamplesFromReader", "amount_needed", amount_needed, "amount_remaining", amount_remaining);
+	DebugLog("AudioReaderSource::GetMoreSamplesFromReader",
+		DEBUGVAR(amount_needed), DEBUGVAR(amount_remaining));
 
 	// Init estimated buffer equal to the current frame position (before getting more samples)
 	estimated_frame = frame_number;
@@ -148,7 +149,9 @@ juce::AudioSampleBuffer* AudioReaderSource::reverse_buffer(juce::AudioSampleBuff
 	int channels = buffer->getNumChannels();
 
 	// Debug
-	ZmqLogger::Instance()->AppendDebugMethod("AudioReaderSource::reverse_buffer", "number_of_samples", number_of_samples, "channels", channels);
+	DebugLog("AudioReaderSource::reverse_buffer",
+		DEBUGVAR(number_of_samples), DEBUGVAR(channels)
+	);
 
 	// Reverse array (create new buffer to hold the reversed version)
 	juce::AudioSampleBuffer *reversed = new juce::AudioSampleBuffer(channels, number_of_samples);
@@ -187,7 +190,7 @@ void AudioReaderSource::getNextAudioBlock(const juce::AudioSourceChannelInfo& in
 		// Do we need more samples?
 		if (speed == 1) {
 			// Only refill buffers if speed is normal
-			if ((reader && reader->IsOpen() && !frame) or
+			if ((reader && reader->IsOpen() && !frame) ||
 				(reader && reader->IsOpen() && buffer_samples - position < info.numSamples))
 				// Refill buffer from reader
 				GetMoreSamplesFromReader();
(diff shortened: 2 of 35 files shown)


ferdnyc:
The main advantage to `zmqLog()`, and I realize I didn't make this very clear or demonstrate it at all in the description, is that it doesn't constrain you to logging only pairs of ("variable name", float_value). You CAN still log that way, but you can also log other data types, or include something other than the value of a variable.

For example, there's a debug message in
`Timeline::update_open_clips(Clip *clip, bool does_clip_intersect)` that currently reads:
```c++
ZmqLogger::Instance()->AppendDebugMethod("Timeline::update_open_clips (before)",
 "does_clip_intersect", does_clip_intersect, "closing_clips.size()", closing_clips.size(),
 "open_clips.size()", open_clips.size());
```
You get the value of a few variables, but you don't even know _what_ clip might or might not be intersecting. With `zmqLog()`, this could be:
```c++
zmqLog() << "Timeline::update_open_clips: clip " << clip->Id()
    << " " << does_clip_intersect ? "intersects" : "does not intersect";
zmqLog() << "Timeline::update_open_clips (before) " << LOGVAR(closing_clips.size())
    << ", " << LOGVAR(open_clips.size());
```

(I _believe_ the stringization behind `LOGVAR()` should be compatible with any expression, not necessarily just a variable name. But that's something that would be good to test, come to think of it.)

ferdnyc:
Fingers crossed that last commit is compatible with even the older ZeroMQ versions on Bionic and Xenial.

ferdnyc:
How in HELL did I miss that OpenShot's `logger_libopenshot.py` directly calls into `ZmqLogger`?!? Back to WIP status for this one, until I come up with a way to keep at least that part of its interface public.

ferdnyc:
I wrapped the StreamLog stuff in `#ifndef SWIG`, the rest of `ZmqLogger.h` _should_ hopefully be compatible. (Unless I have to wrap the `DebugLog()` and `DEBUGVAR()` macro definitions as well.)

PR #594, opened 2020-11-27
Overrides and types

ferdnyc:
This PR fixes the return type for `GetCache()` in various readers, and fixes some Clip methods that were missing `override`.


ferdnyc:
Merging as uncontroversial, the type mismatches and missing overrides are flagged by the compiler with a high enough warning level.

PR #595, opened 2020-12-03
Fixing an issue caused by timeline::GetClip returning a ClipBase* instead of Clip*

jonoomph:
Fixing an issue caused by timeline::GetClip returning a ClipBase instead of a Clip (broke waveform generation). 
Also adding a swig definition for the Caption effect.

### DIFF
--- a/bindings/python/openshot.i
+++ b/bindings/python/openshot.i
@@ -219,6 +219,7 @@
 %include "effects/Bars.h"
 %include "effects/Blur.h"
 %include "effects/Brightness.h"
+%include "effects/Caption.h"
 %include "effects/ChromaKey.h"
 %include "effects/ColorShift.h"
 %include "effects/Crop.h"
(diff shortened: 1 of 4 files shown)


ferdnyc:
@jonoomph Gah! Should've realized I was accidentally slicing the Clip object there, thanks for fixing that up.

`GetClipEffect()` similarly returns `EffectBase*`, but I suppose there's no helping that. But by the same token, it shouldn't be a problem because the common API all lives there.

ferdnyc:
I'm surprised the unit tests didn't need changing, I guess I must've used `auto` for the types. Go me?

Though, a unit test that tries to access the `Clip` API of the returned object (such as `Clip::Waveform`) would've caught this, and would probably be a good thing to add to `Clip_Tests.cpp`.

ferdnyc:
Defense Against the Dark Slice invocations added to the Timeline unit tests, in #596.

PR #599, opened 2020-12-11
Switch to Github Actions for CI builds

ferdnyc:
Well, that somehow went horribly wrong. Stupid command-line `hub` tool. (**Edit:** Ah, because it opened against `master`. Fixed. ...But Codacy is still stuck on it targeting `master`, apparently. Sigh.)

In addition to creating a Github Actions workflow to replace Travis CI, this PR rearranges the handling of the `ENABLE_COVERAGE` option a bit.

Now, if `ENABLE_COVERAGE` is set, the `os_test` target is created to run the _coverage scan_, instead of just running the unit tests. Basically, I wanted to be able to run the coverage from the normal unit test build step, without having to change the script or build an additional target, just by having coverage enabled. So, it made sense to just have coverage usurp `os_test` if it's in use.

And, as part of the above, the `test` target is now _always_ an alias for `os_test`, instead of for `openshot-test`. That way, running `make test` will also trigger the coverage run, if coverage is enabled for the build.

The coverage scan is enabled conditionally for only the `(ubuntu-20.04, clang)` build matrix configuration. It uses Codecov's own Github Actions script to upload the results.


ferdnyc:
The coverage difference is almost certainly down to the switch from FFmpeg in Ubuntu 18.04 to FFmpeg 4 in Ubuntu 20.04, when running the coverage scan.

I should look into how we can do _both_ of those and combine them, actually. I know Codecov supports that.

ferdnyc:
Merging this, it seems to be working and we're dead in the water without Travis.

PR #602, opened 2020-12-22
Build: More conservative linking, private headers

ferdnyc:
**Update:** As I mentioned elsewhere, I am very dumb and somehow failed to realize that `logger_libopenshot.py` in OpenShot _uses the `ZmqLogger` class_ directly, so it can't be made private. Back to draft status til I fix that.

This PR adjusts our library dependency consumption to be more conservative, converting libraries to `PRIVATE` depdendencies of the `openshot` target where possible (thus preventing their linking from being propagated to all targets based on `openshot`, most especially the Python library).

`src/CMakeLists.txt` now contains explicit lists of header files to be installed, a change that has several consequences:
1. The glob-based installation of all `**/*.h` files under `src/` is eliminated
2. **Newly-created headers must now be added to the appropriate list in `src/CMakeLists.txt`, just as newly-created `.cpp` files are. (The header lists immediately follow the corresponding source file list, for maximum visibility and convenience.)**
3. Two files (`ZmqLogger.h` and `OpenMPUtilities.h`) are converted to private headers. All references to them are removed from other OpenShot `.h` files, and they are not included in the lists of public headers to be installed. `ZmqLogger.h` is also removed from the `openshot.i` files that control the SWIG bindings. (I explain in #593 why this is desirable for `ZmqLogger.h`, and suggest a move exactly like the one this PR represents.)
4. The files `json/json.h` and `json/json-forward.h` from `thirdparty/jsoncpp` are now _installed_ to `include/libopenshot/json/`, in cases where the bundled jsoncpp is used. Previously these headers were missing from a libopenshot build that used the bundled jsoncpp, despite `Json.h` having a `#include "json/json.h"` in it.

The library dependencies of the `openshot` target are adjusted so that:
1. OpenMP is a `PRIVATE` dependency, meaning targets that link with `libopenshot.so` will no longer be linked with `libgomp.so` unnecessarily.
2. ZeroMQ is a `PRIVATE` dependency, meaning targets that link with `libopenshot.so` will no longer be linked with `libzmq.so` unnecessarily.
3. The ImageMagick dependency is adjusted to remove `MagickWand` from the `COMPONENTS` list, as we don't use anything from MagickWand.
4. The FFmpeg dependency is adjusted to remove `avdevice`, `avfilter`, and `postproc` from the `COMPONENTS` list (all unused), and to only link with `avresample` in cases where `swresample` is not available.

All in all, this results in a reduction in unnecessary library dependencies for the Python `_openshot.so` module from:

```sh
$ ldd -r -u build/bindings/python/_openshot.so 
Unused direct dependencies:
	/lib64/libMagick++-6.Q16.so.8
	/lib64/libMagickWand-6.Q16.so.6
	/lib64/libMagickCore-6.Q16.so.6
	/lib64/libQt5Widgets.so.5
	/lib64/libavcodec.so.58
	/lib64/libavdevice.so.58
	/lib64/libavformat.so.58
	/lib64/libavfilter.so.7
	/lib64/libavutil.so.56
	/lib64/libpostproc.so.55
	/lib64/libswscale.so.5
	/lib64/libswresample.so.3
	/lib64/libavresample.so.4
	/lib64/libgomp.so.1
	/lib64/libzmq.so.5
```

to:

```sh
$ ldd -r -u build/bindings/python/_openshot.so
Unused direct dependencies:
	/lib64/libMagick++-6.Q16.so.8
	/lib64/libMagickCore-6.Q16.so.6
	/lib64/libQt5Widgets.so.5
	/lib64/libavcodec.so.58
	/lib64/libavformat.so.58
	/lib64/libavutil.so.56
	/lib64/libswscale.so.5
	/lib64/libswresample.so.3
```

So, it's a start.

There was a very old, long-since-closed bug report about the overlinking somewhere, but like I said it's closed anyway so meh.


ferdnyc:
> The FFmpeg dependency is adjusted to remove `avdevice`, `avfilter`, and `postproc` from the `COMPONENTS` list (all unused)

(BTW, hardware-accelerated decoding/encoding isn't part of `avdevice`, all of that stuff is part of `avutil`. `avdevice` is used for stuff like encoding a stream direct from a video capture interface or from the current desktop video output, encoding audio from a microphone, importing webcam streams, etc... none of which we support in any way.)

PR #604, opened 2020-12-24
CMake: Adjustments to build config with OpenCV/Protobuf/Boost

ferdnyc:
This PR against the `opencv` project branch adjusts the CMake build configuration for the newly-introduced OpenCV components, dropping the use of output variables in favor of CMake targets so that dependency relationships are automatically propagated by CMake.

A new CMake option, `ENABLE_OPENCV`, is created, and defaults to `ON`. It can be explicitly set `OFF` with `-DENABLE_OPENCV=0` on the initial `cmake` command line, or it should automatically disable itself if it fails to detect a usable OpenCV install on the system.

Users who have `OpenCV` installed somewhere not accessible to CMake by default may need to provide the path to its CMake configuration on the command line, using either:
1. `-DCMAKE_MODULE_PREFIX=/path/to/OpenCV/installation`
2. `-DOpenCV_ROOT=/path/to/prefix/used/when/installing/OpenCV`

cc: @BrennoCaldato


bindings/python/CMakeLists.txt
```diff
@@ -54,16 +54,6 @@ if (CMAKE_VERSION VERSION_LESS 3.12)
...
-################ OPENCV ##################
-find_package( OpenCV 4 )
-if (OpenCV_FOUND)
-  message("\nCOMPILING WITH OPENCV\n")
-  set(CMAKE_SWIG_FLAGS "-DUSE_OPENCV=1")
-  add_definitions( -DUSE_OPENCV=1 )
-else()
-  message("\nOPENCV NOT FOUND, SOME FUNCTIONALITIES WILL BE DISABLED\n")
-endif()
-
```
ferdnyc, line 66: This is all made unnecessary by the use of `target_link_libraries(openshot PUBLIC <opencv_related_targets>)` in the `src/CMakeLists.txt` file, which makes the target configs propagate to all library dependents automatically.

src/CMakeLists.txt
```diff
@@ -392,36 +392,39 @@ if (ENABLE_BLACKMAGIC)
...
-  find_package(Protobuf 3)
-  
-  if (NOT Protobuf_FOUND)
-    # Protobuf is required when compiling with opencv
-    message(FATAL_ERROR "\nPLEASE INSTALL PROTOBUF. Protobuf is required when compiling with opencv.\n")
+if(ENABLE_OPENCV)
+  find_package(OpenCV 4)
+  if(NOT OpenCV_FOUND)
+    set(ENABLE_OPENCV FALSE CACHE BOOL
+      "Build with OpenCV algorithms (requires Boost, Protobuf 3)" FORCE)
```
ferdnyc, line 399: Basically, `OpenCV` isn't discovered with `REQUIRED` because it's _not_, and if `ENABLE_OPENCV` is `ON` (by default or explicitly) but it's not found, the option will just forcibly reset itself to `OFF`. Any subsequent consumption of any of the OpenCV pieces is then made conditional on `ENABLE_OPENCV` being a truthy value.

src/CMakeLists.txt
```diff
@@ -392,36 +392,39 @@ if (ENABLE_BLACKMAGIC)
...
+      ${ProtoHeaders}
+    )
+    list(APPEND CMAKE_SWIG_FLAGS -DUSE_OPENCV=1)
+    target_compile_definitions(openshot PUBLIC USE_OPENCV=1)
+    target_link_libraries(openshot PUBLIC
+      opencv_core
+      opencv_video
+      opencv_highgui
+      opencv_dnn
+      opencv_tracking
```
ferdnyc, line 422: This list is a _total_ guess, I just basically flung stuff at it until the compile stopped failing with missing symbols. It may need to be corrected, augmented, or pared down.

src/CMakeLists.txt
```diff
@@ -392,36 +392,39 @@ if (ENABLE_BLACKMAGIC)
...
+      protobuf::libprotobuf
```
ferdnyc, line 423: Boost isn't represented here at all, BTW, because it doesn't seem to be needed. If it's required for OpenCV and/or Protobuf, then presumably their target(s) take care of loading it automatically, and we should let them. But it's not required for any of our code AFAICT, is it?

src/CMakeLists.txt
```diff
@@ -392,36 +392,39 @@ if (ENABLE_BLACKMAGIC)
...
-  target_include_directories(openshot PUBLIC ${ProtobufMessagePath})
-
-  # Add OpenCV target sources
-  target_sources(openshot PRIVATE
-    ${OPENSHOT_CV_SOURCES})
-
-  # Link libopenshot with OpenCV libs
-  target_link_libraries(openshot PUBLIC ProtoMessages ${OpenCV_LIBS} ${PROTOBUF_LIBRARY} ${Protobuf_INCLUDE_DIRS})
 endif()
+add_feature_info("OpenCV algorithms" ENABLE_OPENCV "Use OpenCV algorithms")
```
ferdnyc, line 427: This will report the OpenCV status in the `FeatureSummary` table at the end of the configuration run, which is a lot more pleasant than `message()`-ing out status lines that will end up getting repeated multiple times. But the text can be adjusted, augmented, or etc. however we please.

src/CMakeLists.txt
```diff
@@ -450,6 +453,9 @@ install(
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/libopenshot
   FILES_MATCHING PATTERN "*.h"
 )
+install(FILES ${ProtobufHeaders}
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/libopenshot
+)
```
ferdnyc, line 458: Currently, because these are `#include`d in the effect headers, I have the build installing these... but, it would be _really_ nice if we **didn't** have to install them, and they could be treated as private headers.

Do the contents of those files define types or classes that are actually used in the headers? Or could their `#include`s be moved to the `.cpp` files and treated as private headers, instead?

tests/CMakeLists.txt
```diff
@@ -95,15 +76,11 @@ set(OPENSHOT_TEST_FILES
...
-  set(OPENSHOT_CV_TEST_FILES
+if(ENABLE_OPENCV)
+  list(APPEND OPENSHOT_TEST_FILES
     CVTracker_Tests.cpp
     CVStabilizer_Tests.cpp
     # CVObjectDetection_Tests.cpp
-    )
-  set(OPENSHOT_CV_LIBRARIES
-    ${OpenCV_LIBS}
-    ${PROTOBUF_LIBRARY}
```
ferdnyc, line 106: This incorporates the OpenCV unit tests into the single `openshot-test` executable, the testrunner for all of our unit tests. No need for a separate one.

tests/CMakeLists.txt
```diff
@@ -113,21 +90,18 @@ message (STATUS "Tests enabled, test executable will be built as tests/openshot-
...
-  )
+  ${OPENSHOT_TEST_FILES}
+)
 
 # Link libraries to the new executable
 target_link_libraries(openshot-test
   openshot
   ${UnitTest++_LIBRARIES}
-  ${OPENSHOT_CV_LIBRARIES}
-  )
```
ferdnyc, line 125: Again, the use of CMake targets instead of variables means that it takes care of propagating these dependencies automatically.

ferdnyc:
I'll need to adjust the Github Actions CI configs to get OpenCV, Protobuf, and Boost installed into the build image, so that we can get CI runs with those components enabled. At the moment it'll be building with them auto-disabled (and the fact that the builds succeed tells me that worked, yay), which means the CI runs aren't actually testing anything here so far, it's just business as usual for them.

ferdnyc:
Having all of the OpenCV stuff `#include<>`ed publicly in the headers _really_ slows down the compilation of EVERY one of the test objects, BTW, not to mention it blows out their size like crazy. Here's a comparison in `meld` of the build output directory for the unit tests, `opencv_build_config` branch on the left, `develop` branch on the right:

image

A sampling:
|Source file|Previous size|New size|
|-------------|----------------|------------|
|`Cache_Tests.cpp.o`|225.7 kB|309.5 kB|
|`Clip_Tests.cpp.o`|145.7 kB|229.5 kB|
|`Coordinate_Tests.o`|31.1 kB|**115.1 kB (!!!)**|
|`ReaderBase_Tests.cpp`|65.6 kB|149.1 kB|

Unless we can get some of the new `#include`s out of the headers and make them private to indivdual `.cpp` translation units, I think the days of a one-size-fits-all `#include "OpenShot.h"` may be at an end, it's just becoming too costly in terms of compilation size/time.

ferdnyc:
Ping @jonoomph as well, regarding my [previous comment](https://github.com/OpenShot/libopenshot/pull/604#issuecomment-750913978).

src/CMakeLists.txt
ferdnyc: The other reason it would be nice to make some of these headers `PRIVATE` if possible is, it's less stuff for the Python bindings to have to wrap. Anything in the installable public headers is fair game for having to be wrapped in the Python API as well, which is why I've begun working on an effort (PR #602) to migrate references to internal headers out of the other, public headers and into the corresponding `.cpp` files instead. Any header file that isn't referenced by any of the _other_ header files can be considered private and left out of both the install and the generated bindings, which will make our installed APIs tighter and leaner, plus it'll help shave a bit of time off of the builds for both our own internal code and any external API consumers.

src/CMakeLists.txt
ferdnyc: (I forgot to qualify, though, that if it should turn out that it's possible to end up with a system where OpenCV/Protobuf are installed, but the correct Boost components _aren't_, then this whole question may have to be revisited and we may need to do at least some detection/configuration for ourselves. But IMHO that's one of those "don't dick around solving it before you know if it's even possible to _have_ it" type of schrödingproblems.)

ferdnyc:
OK, to mitigate the increased compile times in the unit tests, I've gone through all of the source files and replaced `#include "OpenShot.h"` with individual includes for each header required in the code. PR #607 does that for the _existing_ unit tests in the `develop` branch, and that commit is cherry-picked onto this branch as well, along with a second one that applies the same changes to the `CVfoo_Tests.cpp` files.

bindings/python/CMakeLists.txt
BrennoCaldato: Hi @ferdnyc I was checking your pull request and on my machine I had to include opencv on this cmake, otherwise the opencv effects are not displayed in openshot-qt

ferdnyc: Ok, I'll have to look at that. If that's the case, even with protocol buffers using targets, then most likely what's missing is either the compiler flag that `#define`s `USE_OPENCV` in the SWIG context, _or_ the include directories didn't get propagated to SWIG like they should've. Either way, it's definitely solvable, and will need to be solved in the target context. If any subdirectory config hacks like this are necessary, then the target config isn't right across the build which will bite us when we start generating `EXPORTED` targets from the build. (Which I'm very close to, I've already got it working in libopenshot-audio.)

I'll take a look at it in the morning. What version of CMake are you building with? (I might've asked that in the past, but I forget the answer sorry.) Some of these things do require certain versions, possibly 3.12 or 3.13 at a minimum for it to all work like it's supposed to. (But I can put in checks that use the older hacks to fake it, when they have to.)

BrennoCaldato: I'm currently using CMake 3.10.2 on my system, but I can update it if necessary

ferdnyc: No, that's OK, at least for the moment. The issue wasn't the version. (Though, you should update, 3.10 is ancient! Still, we theoretically support it.)

The real issue is that the `CMAKE_SWIG_FLAGS` propagation from the `src/` directory has been broken for some time, turns out. I just submitted PR #615 to sort that out. Once that's merged, the `SWIG_FLAGS` for the bindings will be set directly from the `COMPILE_DEFINITIONS` set on the `openshot` library target, so there won't be any need to play around with `CMAKE_SWIG_FLAGS` at all anymore.

BrennoCaldato: great @ferdnyc, thanks

ferdnyc: With CMake 3.10, you will have the issue that CMake won't automatically propagate your `INCLUDE_DIRECTORIES` from the `openshot` target. But that's OK, because you don't actually need to be setting any additional `target_include_directories` for the generated protocol buffer sources. What you should do is just always include the subdirectory when you reference them, e.g.:
```c++
#include "protobuf_messages/stabilizedata.pb.h"
```
That way, they'll be covered under the existing `INCLUDE_DIRECTORIES`. (That also means they'll work for external library consumers using the installed headers, because they'll be located at `/usr/local/include/libopenshot/protobuf_messages/`. The library's include path gets exported as `/usr/local/include/libopenshot` on install. (Or wherever it's installed.)

ferdnyc:
@BrennoCaldato I fixed up this PR, and verified that it works with CMake all the way back to 3.5. I also double-checked that the OpenCV-based effects are 100% for sure visible in the Python bindings.

bindings/python/CMakeLists.txt
ferdnyc: In the PR branch I've fixed all the includes to use `#include "protobuf_messages/..."` and `#include "sort_filter/..."`, as well as making sure that all of the includes get installed, and that all of the libraries go the right place. (DLLs on Windows install to the `RUNTIME DESTINATION` (which is `bin/`, not `lib/`), so that needs to be set in the `install(TARGETS...)` for any shared libraries.)

ferdnyc:
When I run `make os_test` from the build dir, I still get this:
```text
[ INFO:0] global /builddir/build/BUILD/opencv-4.3.0/modules/core/src/ocl.cpp (891) haveOpenCL Initialize OpenCL runtime...
[ INFO:0] global /builddir/build/BUILD/opencv-4.3.0/modules/core/src/ocl.cpp (433) OpenCLBinaryCacheConfigurator Successfully initialized OpenCL cache directory: /home/ferd/.cache/opencv/4.3/opencl_cache/
[ INFO:0] global /builddir/build/BUILD/opencv-4.3.0/modules/core/src/ocl.cpp (457) prepareCacheDirectoryForContext Preparing OpenCL cache configuration for context: NVIDIA_Corporation--GeForce_GT_710--455_45_01


 Error: 
  Loaded Data. Saved Time Stamp: 2021-01-13T17:16:06Z


 Error: 
Frame: 1 - good optical flow: 72
Frame: 2 - good optical flow: 72
Frame: 3 - good optical flow: 71
Frame: 4 - good optical flow: 67
Frame: 5 - good optical flow: 71
Frame: 6 - good optical flow: 73
Frame: 7 - good optical flow: 71
Frame: 8 - good optical flow: 72
Frame: 9 - good optical flow: 70
Frame: 10 - good optical flow: 72
Frame: 11 - good optical flow: 69
Frame: 12 - good optical flow: 73
Frame: 13 - good optical flow: 70
Frame: 14 - good optical flow: 72
Frame: 15 - good optical flow: 68
Frame: 16 - good optical flow: 68
Frame: 17 - good optical flow: 70
Frame: 18 - good optical flow: 72
Frame: 19 - good optical flow: 70
Frame: 20 - good optical flow: 68
Frame: 21 - good optical flow: 73

AVERAGE DX: 4.2783 AVERAGE DY: 2.99797 AVERAGE A: 0.00169399
MAX X: -1.68234 MAX Y: 9.09679 MAX A: 0.00234863

Frame: 1 - good optical flow: 72
Frame: 2 - good optical flow: 72
Frame: 3 - good optical flow: 71
Frame: 4 - good optical flow: 67
Frame: 5 - good optical flow: 71
Frame: 6 - good optical flow: 73
Frame: 7 - good optical flow: 71
Frame: 8 - good optical flow: 72
Frame: 9 - good optical flow: 70
Frame: 10 - good optical flow: 72
Frame: 11 - good optical flow: 69
Frame: 12 - good optical flow: 73
Frame: 13 - good optical flow: 70
Frame: 14 - good optical flow: 72
Frame: 15 - good optical flow: 68
Frame: 16 - good optical flow: 68
Frame: 17 - good optical flow: 70
Frame: 18 - good optical flow: 72
Frame: 19 - good optical flow: 70
Frame: 20 - good optical flow: 68
Frame: 21 - good optical flow: 73

AVERAGE DX: 4.2783 AVERAGE DY: 2.99797 AVERAGE A: 0.00169399
MAX X: -1.68234 MAX Y: 9.09679 MAX A: 0.00234863

  Loaded Data. Saved Time Stamp: 2021-01-13T17:16:06Z
Success: 119 tests passed.
Test time: 55.27 seconds.
----------------------------
```

Which is a bit disconcerting. If there's an "Error:", shouldn't the unit test fail?

BrennoCaldato:
> When I run `make os_test` from the build dir, I still get this:

I think I removed these messages in the keyframe-refactor branch. I'll check

BrennoCaldato:
@ferdnyc I tested this PR and so far I could not reproduce these error messages on my system

ferdnyc:
> @ferdnyc I tested this PR and so far I could not reproduce these error messages on my system

I figured out where it's coming from. In `CVTracker_Tests.cpp` you have a
```c++
        std::cout<<"\n\n Error: "<< processingController.GetErrorMessage() <<"\n";
```
Which will always display a message, even if there's no error. (Hence the empty string that gets printed following the "Error:".)

PR #607, opened 2020-12-26
Unit tests: Don't use OpenShot.h header

ferdnyc:
To prevent slow compiles of unit tests, this PR replaces all of the `#include "OpenShot.h"` invocations with includes of the individual headers actually needed by each test file. This has little effect on files like `Clip_Tests.cpp` and `Frame_Tests.cpp`, but **hugely** speeds up the compilation (and reduces the size) of files like `Coordinate_Tests.cpp` and `Fraction_Tests.cpp`.

### DIFF
--- a/tests/Cache_Tests.cpp
+++ b/tests/Cache_Tests.cpp
@@ -28,11 +28,16 @@
  * along with OpenShot Library. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <memory>
+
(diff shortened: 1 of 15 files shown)


ferdnyc:
I'm going to merge this one... #585 might've inspired it, but it's useful even without OpenCV in the mix. The `OpenShot.h` megaheader has always been too much of a kitchen-sink solution to dependencies.

PR #609, opened 2020-12-31
Fix videos with alpha channel

jonoomph:
This is a regression from using premultiplied colorspace in Qt. If a video has an alpha channel, we need to convert the RGBA correctly to premultiplied (which is slower), but without the conversion, all the alpha values are wildly wrong.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -1352,7 +1352,13 @@ void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
 		std::shared_ptr<Frame> f = CreateFrame(current_frame);
 
 		// Add Image data to frame
-		f->AddImage(width, height, 4, QImage::Format_RGBA8888_Premultiplied, buffer);
+		if (!ffmpeg_has_alpha(AV_GET_CODEC_PIXEL_FORMAT(pStream, pCodecCtx))) {
+			// Add image with no alpha channel, Speed optimization
+			f->AddImage(width, height, 4, QImage::Format_RGBA8888_Premultiplied, buffer);
+		} else {
+			// Add image with alpha channel (this will be converted to premultipled when needed, but is slower)
+			f->AddImage(width, height, 4, QImage::Format_RGBA8888, buffer);
+		}
 
 		// Update working cache
 		working_cache.Add(f);
--- a/src/FFmpegUtilities.h
+++ b/src/FFmpegUtilities.h
@@ -126,6 +126,16 @@
 		#define PIX_FMT_YUV444P AV_PIX_FMT_YUV444P
 	#endif
 
+	// Does ffmpeg pixel format contain an alpha channel?
+	inline static const bool ffmpeg_has_alpha(PixelFormat pix_fmt)
+	{
+		if (pix_fmt == AV_PIX_FMT_ARGB || pix_fmt == AV_PIX_FMT_RGBA || pix_fmt == AV_PIX_FMT_ABGR || pix_fmt == AV_PIX_FMT_BGRA || pix_fmt == AV_PIX_FMT_YUVA420P) {
+			return true;
+		} else {
+			return false;
+		}
+	}
+
 	// FFmpeg's libavutil/common.h defines an RSHIFT incompatible with Ruby's
 	// definition in ruby/config.h, so we move it to FF_RSHIFT
 	#ifdef RSHIFT


src/FFmpegReader.cpp
```diff
@@ -1352,7 +1352,13 @@ void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
 		std::shared_ptr<Frame> f = CreateFrame(current_frame);
 
 		// Add Image data to frame
-		f->AddImage(width, height, 4, QImage::Format_RGBA8888_Premultiplied, buffer);
+		if (!ffmpeg_has_alpha(AV_GET_CODEC_PIXEL_FORMAT(pStream, pCodecCtx))) {
+			// Add image with no alpha channel, Speed optimization
+			f->AddImage(width, height, 4, QImage::Format_RGBA8888_Premultiplied, buffer);
+		} else {
+			// Add image with alpha channel (this will be converted to premultipled when needed, but is slower)
+			f->AddImage(width, height, 4, QImage::Format_RGBA8888, buffer);
```
ferdnyc, line 1360: Hmm. It'd be good to get a test file with alpha channel, if possible, to unit-test this. Presumably someone must have one, if they noticed this in the first place?

src/FFmpegUtilities.h
```diff
@@ -126,6 +126,16 @@
...
 
+	// Does ffmpeg pixel format contain an alpha channel?
+	inline static const bool ffmpeg_has_alpha(PixelFormat pix_fmt)
+	{
+		if (pix_fmt == AV_PIX_FMT_ARGB || pix_fmt == AV_PIX_FMT_RGBA || pix_fmt == AV_PIX_FMT_ABGR || pix_fmt == AV_PIX_FMT_BGRA || pix_fmt == AV_PIX_FMT_YUVA420P) {
+			return true;
+		} else {
+			return false;
+		}
+	}
```
ferdnyc, line 137:
```c++
	inline static const bool ffmpeg_has_alpha(PixelFormat pix_fmt)
	{
		if (pix_fmt == AV_PIX_FMT_ARGB || pix_fmt == AV_PIX_FMT_RGBA || pix_fmt == AV_PIX_FMT_ABGR || pix_fmt == AV_PIX_FMT_BGRA || pix_fmt == AV_PIX_FMT_YUVA420P) {
			return true;
		} else {
			return false;
		}
	}
```

For maintainability this might be better as,
```c++
        inline static const bool ffmpeg_has_alpha(PixelFormat pix_fmt) {
            const AVPixFmtDescriptor *fmt_desc = av_pix_fmt_desc_get(pix_fmt);
            return bool(desc->flags & AV_PIX_FMT_FLAG_ALPHA);
        }
```
That way it'll support any new pixel formats they add in the future, too. Near as I can tell it'd be compatible with every FFmpeg back to at least 2.4, which is already way farther back than we should worry about going.

src/FFmpegReader.cpp
ferdnyc, line 1360: FFmpeg does actually have [a premultiply filter](https://ffmpeg.org/ffmpeg-filters.html#premultiply) (accessible from code as `avfilter_get_by_name("premultiply");`), which in theory could be imposed into the processing graph to have it produce premultiplied alpha values. But I have a feeling we're better off letting Qt handle that.

src/FFmpegUtilities.h
ferdnyc: Actually the `static` storage specifier on that may have to come off, if it's going to be taking struct-pointer return values from `av_pix_fmt_desc_get()` and dereferencing them. But an `inline` (non-`static`) function should work fine, I think?

Otherwise, it can always just be added to `FFmpegReader.cpp`.

ferdnyc: @jonoomph See my previous comment(s), regarding your email question.

ferdnyc: @jonoomph Oops. Just realized this had a typo in it, it should've been
```c++
        inline static const bool ffmpeg_has_alpha(PixelFormat pix_fmt) {
            const AVPixFmtDescriptor *fmt_desc = av_pix_fmt_desc_get(pix_fmt);
            return bool(fmt_desc->flags & AV_PIX_FMT_FLAG_ALPHA);
        }
```

src/FFmpegReader.cpp
jonoomph: I've recently learned that there are many formats which can allow an alpha channel, and thus, we can always add a transparent video into our unittests (I like that idea), but it won't test for all possible alpha formats. So, it still leaves the possibility that a format is not working correctly, but I suppose that will just be the case.

src/FFmpegUtilities.h
jonoomph: This looks terrific. I'm gonna test it right now! This is exactly what I was searching for, lol. I have a few transparent videos I've been using to test this, so it should be a quick check.

PR #613, opened 2021-01-03
Docs: Switch from UseDoxygen to doxygen_add_docs

ferdnyc:
Since CMake 3.11, the supported method of integrating Doxygen into
CMake projects has been via the `doxygen_add_docs()` CMake function,
which (crucially) can be passed a target name to create and doesn't
rely on tons of hardcoding like our `UseDoxygen.cmake` and its
Doxyfile.in template did.

This means our docs (only) now require CMake 3.11 to generate, but it
allows Doxygen to be configured/used multiple times in the same CMake
configuration, removing a major barrier to superproject builds.


CMakeLists.txt
```diff
@@ -146,26 +146,70 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/src/OpenShotVersion.h
 
 ################### DOCUMENTATION ###################
 # Find Doxygen (used for documentation)
-set(DOCS_ENABLED FALSE) # Only set true if Doxygen is found and configured
-if (ENABLE_DOCS)
-  include(cmake/Modules/UseDoxygen.cmake)
+set(LIB_DOCS_ENABLED FALSE) # Only set true if Doxygen is found and configured
```
jonoomph, line 149: @ferdnyc I'm a bit confused by this logic. We define ENABLE_DOCS, DOCS_ENABLED, LIB_DOCS_ENABLED, AND ENABLE_LIB_DOCS. Just wanted to verify this was the intention.

jonoomph:
All the variations of variable names hurt my brain :exploding_head: , but if this all works great, then feel free to merge when you are happy with it. :+1:

CMakeLists.txt
ferdnyc: No, I hear you, it is a mess.

`DOCS_ENABLED` and `ENABLE_DOCS` are no more, they've been replaced with `ENABLE_LIB_DOCS` and `LIB_DOCS_ENABLED`. So, there are still only two, they're just more specific.

Previously, we were using the same names in the `libopenshot-audio` build — which interferes with being able to build them together (with `-audio` as an `add_subdirectory()` of the libopenshot build), since the option selected affects both. (To say nothing of the mess that would happen with `DOCS_ENABLED`, when it was being set from both build trees.)

Now, libopenshot uses `ENABLE_LIB_DOCS` and libopenshot-audio uses `ENABLE_AUDIO_DOCS`, so the user has two distinct commandline options that don't interfere with each other. And the internal variables are similarly non-clashing.

The fact that there are even just _two_ similar-seeming names admittedly isn't great, still, though there is "logic" to it: 
* `ENABLE_DOCS` (now `ENABLE_LIB_DOCS`) is the user option that can be set false to disable the documentation build.
* `DOCS_ENABLED` (now `LIB_DOCS_ENABLED`) is the internal status variable that records whether the documentation part of the build should actually be run. If Doxygen or Python3 isn't found by the build system, `LIB_DOCS_ENABLED` will be set to `FALSE` even if the user specified `ENABLE_LIB_DOCS=1` initially, and the `doc` target won't be created.

I suppose I could change the name of `LIB_DOCS_ENABLED` to something like `LIB_DOCS_AVAILABLE`, `LIB_DOCS_ACTIVE`, `LIB_CAN_BUILD_DOCS`, etc... that would hopefully make things a bit less confusing.

ferdnyc: (Same reason I changed a bunch of strings to use `${PROJECT_NAME}` instead of hardcoding some fixed text. That way, when a combined summary is displayed by a superproject build, it will clearly differentiate items that pertain to `libopenshot` from the ones that pertain to `libopenshot-audio`.)

CMakeLists.txt
```diff
@@ -146,26 +146,70 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/src/OpenShotVersion.h
...
+if(NOT DEFINED ENABLE_LIB_DOCS)
+  set(ENABLE_LIB_DOCS ${ENABLE_DOCS})
+endif()
```
ferdnyc, line 152: @jonoomph Oh, right, but I do still check for the old `ENABLE_DOCS` option, just in case the user is still setting it on the command line. Purely a legacy-support thing, to avoid breaking builds (but not brains).

I should probably emit a deprecation message in that case, though, so we can eventually get rid of that variable.

ferdnyc:
@jonoomph 

Actually, turned out things were a little off, as I hadn't _fully_ deprecated `ENABLE_DOCS`. I think I was trying to set up some sort of complicated thing where the user _could_ still use `ENABLE_DOCS` to control _both_ projects' doc builds, but then override either individual one with the more specific variable.

That was too damn confusing, so now the option is just `ENABLE_LIB_DOCS`, and `ENABLE_DOCS` will be respected but cause a deprecation warning to be emitted.

`LIB_DOCS_ENABLED` is now renamed to `CAN_BUILD_LIB_DOCS`, though, which hopefully makes things clearer. And I did the same with `TESTS_ENABLED` (which had the same relationship to `ENABLE_TESTS`), that's now renamed to `CAN_RUN_UNIT_TESTS`.

ferdnyc:
> `LIB_DOCS_ENABLED` is now renamed to `CAN_BUILD_LIB_DOCS`, though, which hopefully makes things clearer. And I did the same with `TESTS_ENABLED` (which had the same relationship to `ENABLE_TESTS`), that's now renamed <s>to `CAN_RUN_UNIT_TESTS`</s>.

Make that `CAN_BUILD_TESTS`, for consistency.

ferdnyc:
There's also the wrinkle, as I recently re-discovered, that our Linux builder is running CMake 3.10 — one release shy of the required CMake 3.11.

ferdnyc:
Rebased, and wow was _that_ harrowing. (This PR originally dates from back before Catch2, even!) Hopefully I didn't break anything rearranging it all.

jonoomph:
LG, merging and 🤞

PR #614, opened 2021-01-11
Fixing INSTALL issue related to ProtoBufMessage.dylib

jonoomph:
Changing name of ProtoBufMessage to openshot_protobuf for clarity, also installing it in PREFIX/lib. Also, versioning the same as libopenshot, so we don't get conflicts between releases and updates to the protobuf message types.


jonoomph:
This is a bit experimental, and I'm testing things on the Mac builder specifically. I'm gonna merge this and see how it does on the builders.

PR #615, opened 2021-01-13
Propagate library defs to swig

ferdnyc:
I suspect since the bindings were moved out of the `src/` directory, the `CMAKE_SWIG_FLAGS` haven't been propagating to the `swig` command line that generates the wrapper script. This means that the `#ifdef`s in the `openshot.i` files are all interpreted wrong.

This PR adds generator expression logic to copy _every_ public compile definition from the `openshot` library target over to the `SWIG_FLAGS` of the `openshot.i` file, which should ensure that the bindings are generated correctly.

In my version of CMake, all compile definitions are stored in the target property _without_ the leading `-D`, even if it was originally set. So, the generator expression I added will prepend `-D` to each of the values. I'll have to check whether that holds for older CMake versions as well, otherwise we might end up with `-D-DUSE_RESVG=1` or the like, which would be a problem. The CI runs should be a fair test of that, though. (And just in case, I took the `-D`s out of the definitions _we_ set, at least.)

**Edit:** That wasn't the issue, but older CMakes did have a problem with the generator expressions. Added a workaround using `get_property()`, tested back to CMake 3.5.

Also added a fix to `FindResvg.cmake`, for compatibility with older CMake versions.

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -58,6 +58,12 @@ endif()
 set_property(SOURCE openshot.i PROPERTY CPLUSPLUS ON)
 set_property(SOURCE openshot.i PROPERTY SWIG_MODULE_NAME openshot)
 
+# Set the SWIG_FLAGS from the library target, IFF its
+# COMPILE_DEFINITIONS property is set (in practice, always true)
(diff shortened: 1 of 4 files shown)


ferdnyc:
Merging this, as the current state of affairs is quite broken, and it's part of the `opencv` branch now anyway.

PR #611, opened 2021-01-01
Eliminate TooManySeeks exception

ferdnyc:
Since commit bfa050409c22bd7e5cbc1d7e23c72b11751ded07 (2015-08-24), there is no code in libopenshot which ever throws `TooManySeeks`.

- Remove `catch()` statements for `TooManySeeks` from multiple functions
- Remove the exception from `Exceptions.h`
- in `Qt/AudioPlaybackThread.h`:
  - Remove the "`SafeTimeSliceThread`" class definition, as it only existed to catch `TooManySeeks`.
  - Replace `SafeTimeSliceThread` with a standard `juce::TimeSliceThread`

PR #618, opened 2021-01-26
Deprecate TooManySeeks, instead of removing

ferdnyc:
Recently (in #611) I removed the TooManySeeks exception, which has been unused in the code since 2015.

But, just because _we_ don't use it doesn't mean that nobody ELSE uses it. So, in case any of the library's consumers are still catching `TooManySeeks` (as we no longer do), to avoid breaking their code this PR restores the exception. However, both the class and the constructor are marked as deprecated, causing both `g++` and `clang++` to emit a warning if it's used:

`g++`:
```text
[ 60%] Building CXX object src/CMakeFiles/openshot.dir/Qt/PlayerPrivate.cpp.o
src/Qt/PlayerPrivate.cpp: In member function ‘std::shared_ptr<openshot::Frame> openshot::PlayerPrivate::getFrame()’:
src/Qt/PlayerPrivate.cpp:178:35: warning: ‘TooManySeeks’ is deprecated: The library no longer throws this exception. It will be removed in a future release. [-Wdeprecated-declarations]
  178 |     } catch (const TooManySeeks & e) {
      |                                   ^
In file included from src/Qt/PlayerPrivate.cpp:33:
src/Exceptions.h:375:2: note: declared here
  375 |  TooManySeeks : public ExceptionBase
      |  ^~~~~~~~~~~~
```

`clang++`:
```text
[ 61%] Building CXX object src/CMakeFiles/openshot.dir/Qt/PlayerPrivate.cpp.o
src/Qt/PlayerPrivate.cpp:178:20: warning: 'TooManySeeks' is deprecated: The library no longer throws this exception. It will be removed in a future release. [-Wdeprecated-declarations]
    } catch (const TooManySeeks & e) {
                   ^
src/Exceptions.h:373:18: note: 'TooManySeeks' has been explicitly marked deprecated here
        __attribute__ ((deprecated (TMS_DEP_MSG) ))
                        ^
1 warning generated.
```

`TooManySeeks` is also excluded from the SWIG bindings entirely using `#ifndef SWIG`.

A separate commit in this PR moves most (all?) `#include "Exceptions.h"` lines from the header files to their corresponding `.cpp` files, to reduce unnecessary recompilations and header bloat.

### DIFF
--- a/src/AudioReaderSource.cpp
+++ b/src/AudioReaderSource.cpp
@@ -29,6 +29,7 @@
  */
 
 #include "AudioReaderSource.h"
+#include "Exceptions.h"
 
 using namespace std;
 using namespace openshot;
--- a/src/AudioResampler.h
+++ b/src/AudioResampler.h
@@ -32,7 +32,6 @@
 #define OPENSHOT_RESAMPLER_H
 
 #include "AudioBufferSource.h"
-#include "Exceptions.h"
 #include "JuceHeader.h"
 
 namespace openshot {
--- a/src/CacheBase.h
+++ b/src/CacheBase.h
@@ -34,7 +34,6 @@
 #include <memory>
 #include <cstdlib>
 #include "Frame.h"
-#include "Exceptions.h"
 #include "Json.h"
 
 namespace openshot {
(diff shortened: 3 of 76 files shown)


ferdnyc:
Merging this, as it's what I should've done in the first place. (It just took me a bit more time to figure out the right syntax for deprecating a _class_, and to ensure that it worked with both GCC and Clang.)

PR #619, opened 2021-01-26
Improved Detection of AVPixelFormat containing Alpha channel

jonoomph:
Improved detection of AVPixelFormat containing Alpha channel, I'm going with a simple list. Not sure if this will work well with older versions of FFmpeg though, as these items change over time. Waiting to see how build servers handle it.

### DIFF
--- a/src/FFmpegUtilities.h
+++ b/src/FFmpegUtilities.h
@@ -129,7 +129,25 @@
 	// Does ffmpeg pixel format contain an alpha channel?
 	inline static const bool ffmpeg_has_alpha(PixelFormat pix_fmt)
 	{
-		if (pix_fmt == AV_PIX_FMT_ARGB || pix_fmt == AV_PIX_FMT_RGBA || pix_fmt == AV_PIX_FMT_ABGR || pix_fmt == AV_PIX_FMT_BGRA || pix_fmt == AV_PIX_FMT_YUVA420P) {
+		if (pix_fmt == AV_PIX_FMT_ARGB ||
+		    pix_fmt == AV_PIX_FMT_RGBA ||
+		    pix_fmt == AV_PIX_FMT_ABGR ||
+		    pix_fmt == AV_PIX_FMT_BGRA ||
+		    pix_fmt == AV_PIX_FMT_YUVA420P ||
+		    pix_fmt == AV_PIX_FMT_YA16LE ||
+		    pix_fmt == AV_PIX_FMT_YA16BE ||
+		    pix_fmt == AV_PIX_FMT_YA8 ||
+		    pix_fmt == AV_PIX_FMT_GBRAP ||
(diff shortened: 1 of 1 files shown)


ferdnyc:
@jonoomph 
> Not sure if this will work well with older versions of FFmpeg though, as these items change over time.

That's my biggest worry. In FFmpeg 3.2, the last four are undefined:
```sh
$ for format in AV_PIX_FMT_ARGB AV_PIX_FMT_RGBA AV_PIX_FMT_ABGR \
AV_PIX_FMT_BGRA AV_PIX_FMT_YUVA420P AV_PIX_FMT_YA16LE \
AV_PIX_FMT_YA16BE AV_PIX_FMT_YA8 AV_PIX_FMT_GBRAP \
AV_PIX_FMT_GBRAP10BE AV_PIX_FMT_GBRAP10LE AV_PIX_FMT_GBRAP12BE \
AV_PIX_FMT_GBRAP12LE AV_PIX_FMT_GBRAP16BE AV_PIX_FMT_GBRAP16LE \
AV_PIX_FMT_GBRPF32BE AV_PIX_FMT_GBRPF32LE AV_PIX_FMT_GBRAPF32BE \
AV_PIX_FMT_GBRAPF32LE; do echo -n "$format: "; grep -c $format pixfmt.h; done
AV_PIX_FMT_ARGB: 1
AV_PIX_FMT_RGBA: 4
AV_PIX_FMT_ABGR: 1
AV_PIX_FMT_BGRA: 4
AV_PIX_FMT_YUVA420P: 10
AV_PIX_FMT_YA16LE: 1
AV_PIX_FMT_YA16BE: 1
AV_PIX_FMT_YA8: 3
AV_PIX_FMT_GBRAP: 10
AV_PIX_FMT_GBRAP10BE: 1
AV_PIX_FMT_GBRAP10LE: 1
AV_PIX_FMT_GBRAP12BE: 1
AV_PIX_FMT_GBRAP12LE: 1
AV_PIX_FMT_GBRAP16BE: 1
AV_PIX_FMT_GBRAP16LE: 1
AV_PIX_FMT_GBRPF32BE: 0
AV_PIX_FMT_GBRPF32LE: 0
AV_PIX_FMT_GBRAPF32BE: 0
AV_PIX_FMT_GBRAPF32LE: 0
```

In older versions, there will be even more missing.

The committed code should be fully compatible with FFmpeg 3.4+, but only those versions. (We aren't testing anything older in the CI currently, due to the Github Actions switchover.)

As you know, I'm a big proponent of dropping all support for versions prior to 3.4, along with the associated maintenance headaches involved... this code effectively does just that. So if we're going to keep it, we can drop a _lot_ of legacy code, which would be great. There's no reason to keep the pre-3.4 compatibility code hanging around when those versions will no longer compile.

PR #620, opened 2021-01-27
Thanks to @ferdnyc for the suggestion to better detect alpha channels.

jonoomph:
Related to https://github.com/OpenShot/libopenshot/pull/609. Implementing a much improved alpha channel detection method, which works on all formats that contain an alpha channel. Shout out to @ferdnyc for the suggestion!

### DIFF
--- a/src/FFmpegUtilities.h
+++ b/src/FFmpegUtilities.h
@@ -127,32 +127,10 @@
 	#endif
 
 	// Does ffmpeg pixel format contain an alpha channel?
-	inline static const bool ffmpeg_has_alpha(PixelFormat pix_fmt)
-	{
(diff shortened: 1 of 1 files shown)


jonoomph:
Sorry to disappoint, lol. I'm not opposed to dropping older FFmpeg support though. As long as we support the past 4 or 5+ years of FFmpeg (which I think is around FFmpeg 3.4), that seems very reasonable.

ferdnyc:
@jonoomph 
Well, Ubuntu 16.04 shipped with FFmpeg 2.8, and won't technically be EOL'd until April of this year. (It also still uses Python 3.5, which has been EOL'd for nearly a year already, so that's confusing.)

Come April, though, there's really no reason to support anything pre-3.4, as that's what was included in Ubuntu 18.04.

PR #556, opened 2020-08-20
Use std::make_shared to allocate shared ptrs, instead of std::shared_ptr constructors

ferdnyc:
std::make_shared does in a single allocation what the constructors for std::shared_ptr usually take at least 2 allocations to do.

```cpp
//
// This PR replaces instances of shared-pointer allocations like:
//
std::shared_ptr<Type> name = std::shared_ptr<Type>(new Type(...));
name = std::shared_ptr<Type>(new Type(...));
std::shared_ptr<Type> name(new Type(...));
//
// with calls to the make_shared factory function, e.g:
//
auto name = std::make_shared<Type>(...);
name = std::make_shared<Type>(...);
```
As you can see, the revised syntax is at worst as concise, and in some cases much _more_ concise, than the previous code. And, doing this may give us an infinitesimal performance/memory improvement (though I don't expect it's measurable except in contrived benchmarks).

For said contrived benchmarks, see: https://www.modernescpp.com/index.php/memory-and-performance-overhead-of-smart-pointer

PR #621, opened 2021-01-27
Effect GetFrame(): Use std::make_shared

ferdnyc:
This PR tweaks the `GetFrame(int64_t frame_number)` overload in each effect from:
```c++
std::shared_ptr<openshot::Frame> GetFrame(int64_t frame_number) override {
    return GetFrame(std::shared_ptr<Frame> (new Frame()), frame_number);
}
```
to:
```c++
std::shared_ptr<openshot::Frame> GetFrame(int64_t frame_number) override {
    return GetFrame(std::make_shared<openshot::Frame>(), frame_number);
}
```
In accordance with #556, which did the same for other instances of shared-pointer creation. Extended rationale there.

### DIFF
--- a/src/effects/Bars.h
+++ b/src/effects/Bars.h
@@ -83,7 +83,7 @@ namespace openshot
 		///
 		/// @returns A new openshot::Frame object
 		/// @param frame_number The frame number (starting at 1) of the clip or effect on the timeline.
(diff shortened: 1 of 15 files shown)


jonoomph:
LGTM!

ferdnyc:
Conflicts resolved, merging.

PR #622, opened 2021-01-28
SVG Scaling Improvements (especially if scaling SVG larger than Timeline)

jonoomph:
Fixing a big issue where SVG files are not correctly scaled to larger resolutions, for cases where the default size is smaller than the Timeline size (or preview size). Now SVG files are rescaled/re-rasterized larger when needed, and otherwise cached.

Before this fix, for example, an emoji (default size 71x71) would be loaded, and upscaled to the preview window size... and was blurry. Now, it is re-rasterized to get an HD sharp image, so the emoji is much prettier.


jeffski:
Is it possible to do this for text based readers too (QtText, QtHTML and TextReader)? I've noticed that these are drawn at the original size but then if you apply a keyframe animation to scale the text bigger, over time, the text/image gets blurry as they are scaled. I presume the original size image of the text is used and scaled rather than redrawing on each frame.

Not sure if that is possible or if it would be slow to render but it would make the text a lot sharper on each frame.

jonoomph:
Sure, this can be done on all the Readers... but SVG was really noticeable, lol.

jeffski:
It would be great upgrade if you could do it!!

PR #625, opened 2021-01-29
Add Exceptions.h include to new effects

ferdnyc:
The new effects were picking up the definition of `InvalidJSON` from one of the other headers, which no longer `#include "Exceptions.h"`. So, they should include it directly.

### DIFF
--- a/src/effects/ObjectDetection.cpp
+++ b/src/effects/ObjectDetection.cpp
@@ -30,6 +30,7 @@
 
 #include "effects/ObjectDetection.h"
 #include "effects/Tracker.h"
+#include "Exceptions.h"
 
 using namespace std;
 using namespace openshot;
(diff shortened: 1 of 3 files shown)


jonoomph:
LGTM!

jonoomph:
Thanks for jumping in and helping! Appreciate it.

ferdnyc:
@jonoomph   Ooh, I was going to jump the gun on the CI (and probably still will), but the logs just reminded me that we need to update the CI dependency installation to include OpenCV and Protocol Buffers.

ferdnyc:
I'll take care of that after I merge this. Not a problem.

BrennoCaldato:
Thanks @ferdnyc. I'm reviewing the last changes now.

PR #626, opened 2021-01-29
Add Protobuf and OpenCV to CI preinstalls

ferdnyc:
Make sure our CI runs are built with OpenCV and Protocol Buffers enabled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,11 +24,13 @@ jobs:
       run: |
         sudo add-apt-repository ppa:openshot.developers/libopenshot-daily
         sudo apt update
-        sudo apt install cmake swig doxygen graphviz curl lcov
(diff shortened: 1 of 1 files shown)


ferdnyc:
From an apt perspective, the added installs are:

`sudo apt install libopencv-dev libprotobuf-dev protobuf-compiler`

(Leave out that last one, you'll get _very_ strange errors because instead of being replaced with the path to the protoc executable, the `protobuf::protoc` target ends up in the Makefile as a dependency. Which breaks it completely, `::` not being valid unescaped path characters in Make.)

Installing individual OpenCV library components (e.g. `libopencv-core-dev`, etc. is a nice idea, to avoid installation bloat, and that's how I started. But none of them install the CMake configs needed to properly configure the targets. Those are only in `libopencv-dev` itself. So it'll just have to install everything.

PR #591, opened 2020-11-21
Add class constructors taking STL container types, primarily for Python use

ferdnyc:
#### Update 2
I finally solved the (for me) holy grail:
```python3
>>> import openshot
>>> f = openshot.Fraction(30000, 1001)
>>> dict(f)
# The "10001" that used to be here was a typo
{'den': 1001, 'num': 30000}
>>> f["num"]
30000
>>> f["den"]
1001
```
Which makes Fraction _hugely_ flexible, because now we can _almost_ store it **directly** in Python data types and access it transparently as if it's a `dict`:
```python
>>> data = {}
>>> data.update({"fps": openshot.Fraction(30000, 1001)})
>>> data
{'fps': Fraction(30000, 1001)}
>>> data["fps"]
Fraction(30000, 1001)
>>> data["fps"]["num"]
30000
>>> data["fps"]["den"]
1001
```
The "almost" comes about because `data` above isn't serializable:
```python
>>> import json
>>> json.dumps(data)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib64/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib64/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Fraction is not JSON serializable
```
But it doesn't stop us from doing this:
```python3
>>> data = {}
>>> data.update({"fps": dict(openshot.Fraction(30000, 1001))})
>>> data
{'fps': {'den': 1001, 'num': 30000}}
```

#### Update
I added STL container constructors to various other OpenShot data types, including Coordinate, Point, and Keyframe. Coordinate now accepts a `std::pair` tuple, and `Keyframe` will accept a `std::vector<>` of either `openshot::Point` or `openshot::Coordinate`, and build its keyframe list from that.

You still can't do this, Python isn't that smart:
```python
>>> kf = openshot.Keyframe([(1, 1), (2, 2), (3, 3)])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "bindings/python/openshot.py", line 2728, in __init__
[...]

PR #627, opened 2021-01-29
KeyFrame.h: Avoid shadowing SWIG templates

ferdnyc:
Hopefully this will fix the SWIG build on launchpad.

### DIFF
--- a/bindings/python/openshot.i
+++ b/bindings/python/openshot.i
@@ -141,6 +141,16 @@
 %template() std::pair<double, double>;
 %template() std::pair<float, float>;
 
+/* Wrap std templates (list, vector, etc...) */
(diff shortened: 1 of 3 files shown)


ferdnyc:
This is kind of a shot in the dark because I can't reproduce the launchpad issue anywhere, but it's definitely related to my container changes in #591 so hopefully this will be more to its liking.

ferdnyc:
No love. I'm beginning to suspect the issue might just be an incompatibility between the new code and the old SWIG 3.0.8 being used in Ubuntu 16.04 builds (the only ones that fail). In which case, I'm wondering how much I really care.

ferdnyc:
Installed `swig3.0` on Ubuntu 20.04, building with it now to test my theory.

PR #628, opened 2021-01-29
Tweak the Python bindings for KeyFrame, Fraction

ferdnyc:
I screwed up the slot name for Fraction's string representation (`__str__` not `__string__`, dummy!), so this PR fixes that.

It also gets rid of the `Keyframe(std::vector<Coordinate>)` constructor, which as SWIG points out in an emitted warning will shadow `Keyframe(std::vector<Point>)` or vice versa, and the `Point` version is far more useful. The aliased names for the arguments are also removed, on the off chance they're part of what's messing up SWIG on Ubuntu 16.04.

### DIFF
--- a/bindings/python/openshot.i
+++ b/bindings/python/openshot.i
@@ -156,6 +156,7 @@
 %{
 	#include <sstream>
 	#include <map>
+	#include <vector>
 
 	static std::vector<std::string> _keys{"num", "den"};
 	static int fracError = 0;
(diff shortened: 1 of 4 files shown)


ferdnyc:
Excellent: [This worked](https://launchpad.net/~openshot.developers/+archive/ubuntu/libopenshot-daily/+recipebuild/2727237) to fix Ubuntu 16.04.

PR #637, opened 2021-02-18
Clip: Use QPainter opacity

ferdnyc:
Hey, I loves me a good bit-banging as much as the next guy.

(OK, that's a lie, I completely hate it and feel like, if we're fiddling with the raw, individual bits of an image, we're way too low-level and should be using better APIs or support classes.)

But I _especially_ can't countenance `openshot::Clip()` _manually applying_ alpha transparency in a `for()` loop, immediately before it uses `QPainter` to draw the resulting image... when **`QPainter` _has_ a `.setOpacity()` method!!**

...So let's just use that, why don't we?

**Edit:** Also, I finally figured out what that one Codacy message was complaining about, with the `source_image` in `Clip::apply_keyframes`. (We were double-encapsulating the waveform `QImage` in a `std::shared_ptr`.) Fixed.

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -1150,29 +1150,6 @@ void Clip::apply_keyframes(std::shared_ptr<Frame> frame, int width, int height)
 		frame->AddImage(std::shared_ptr<QImage>(source_image));
 	}
 
-	/* ALPHA & OPACITY */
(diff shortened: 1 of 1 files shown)


ferdnyc:
This code has changed so much in the wake of #638 that I can't even begin to reconcile it.

PR #638, opened 2021-02-18
Huge performance improvements (43% faster FFmpegReader, 34% faster Timeline)

jonoomph:
After some extensive profiling using a variety of tools, we have found a few interesting and actionable issues. Bottom line, OpenMP is not actually providing us any improvement where we need it, and instead, it's adding additional overhead, by spinning up many threads, waiting for threads to finish, shutting down threads, etc... 

Related openshot-qt PR: https://github.com/OpenShot/openshot-qt/pull/4012

**Improvements:**
 - I've removed **OpenMP** from `Timeline`, `FFmpegReader` classes. These are now optimized to return the first matching frame as quickly as possible (no more attempting to process a bunch of Frame instances in multiple threads).
 - I've refactored Clip::GetFrame, including an override that allows the Timeline to pass in the current Frame (to be used as a background image). This prevents calling `QPainter::drawImage` multiple times (and saves a bunch of CPU)
 - Reduced some unneeded memory copying and allocating

**Testing:**
 - I did many tests on many different machines, including A/B testing with the current stable release of OpenShot. 
 - I tested on Chromebook, Windows, Mac, and Linux (to verify the performance gains on a variety of different machines, some very powerful, some borderline decent).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,6 @@
 .project
 .cproject
 /.metadata/
+cmake-build-debug/*
 tags
 *~
-
--- a/examples/Example.cpp
+++ b/examples/Example.cpp
@@ -39,51 +39,52 @@ using namespace openshot;
 
 int main(int argc, char* argv[]) {
 
-    Settings *s = Settings::Instance();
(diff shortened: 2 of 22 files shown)


jonoomph:
Merging this one!

PR #649, opened 2021-03-23
Effect parenting

BrennoCaldato:
This PR adds in functionalities for using the new OpenCV-related effects. The main functionalities are:
- Added transform handlers for the Tracker and Object Detection effects. The user can now modify and correct the tracked objects.
- Fixed Tracker initial bounding box selection window
- Allow the user to animate Clips or Images with the tracked objects. The clip will be transformed to follow the tracked object
- Allow parenting clips. The transformations applied to a clip can be replicated to other clips.
- Effects can be parented and replicated to other clips.
- Improved usability of Tracker effect

A TrackedObjectBase class was created to handle custom objects and interact with them. TrackedObjectBBox allows storing and manipulating the bounding box information of a tracked object. The base class can be further extended to other objects such as facial tracking, body joints, etc.

Related to https://github.com/OpenShot/openshot-qt/pull/4058

**Tracker and ObjectDetection transform handlers:**
image

**Clip attached to a tracked object**
image

**Effect parenting**
image

**Object Detection with transform handlers**
image


jonoomph:
@BrennoCaldato This is a great start! A few notes...

Tracker & Object Detection Effect
------------------
 - Add some additional keyframes/properties:
   - Stroke: border line color
   - Background: background fill color
   - Stroke Width: thickness of border line
   - Background Corner: Radius of rounded corners (if possible or easy)
 ```
	root["stroke"] = add_property_json("Border", 0.0, "color", "", NULL, 0, 255, false, requested_frame);
	root["stroke"]["red"] = add_property_json("Red", stroke.red.GetValue(requested_frame), "float", "", &stroke.red, 0, 255, false, requested_frame);
	root["stroke"]["blue"] = add_property_json("Blue", stroke.blue.GetValue(requested_frame), "float", "", &stroke.blue, 0, 255, false, requested_frame);
	root["stroke"]["green"] = add_property_json("Green", stroke.green.GetValue(requested_frame), "float", "", &stroke.green, 0, 255, false, requested_frame);
	root["background_alpha"] = add_property_json("Background Alpha", background_alpha.GetValue(requested_frame), "float", "", &background_alpha, 0.0, 1.0, false, requested_frame);
	root["background_corner"] = add_property_json("Background Corner Radius", background_corner.GetValue(requested_frame), "float", "", &background_corner, 0.0, 60.0, false, requested_frame);
	root["background"] = add_property_json("Background", 0.0, "color", "", NULL, 0, 255, false, requested_frame);
	root["background"]["red"] = add_property_json("Red", background.red.GetValue(requested_frame), "float", "", &background.red, 0, 255, false, requested_frame);
	root["background"]["blue"] = add_property_json("Blue", background.blue.GetValue(requested_frame), "float", "", &background.blue, 0, 255, false, requested_frame);
	root["background"]["green"] = add_property_json("Green", background.green.GetValue(requested_frame), "float", "", &background.green, 0, 255, false, requested_frame);
	root["stroke_width"] = add_property_json("Stroke Width", stroke_width.GetValue(requested_frame), "float", "", &stroke_width, 0, 10.0, false, requested_frame);
```

Object Detection Effect
------------------
 - Add the following properties: 
 - Class filter: to only show a comma delimited list of filters
 - Confidence filter: 0 to 1 - filter out a specific confidence score
 - Ability to attach an effect to each matching region... or, if easier, we can have a dropdown of a few preselected effects (pixelate, blur, etc...), but that would be more limited and have more duplication of code. However, choosing a clip, and associating it with the object tracker, we could have the user select to display the clip contents, or simply apply the clip's effects to each region (ignoring the pixel values of the clip content). I kind of like this approach. For example:

1) We attach an object detector, filter it to only "people" of 90% confidence
2) We drop a clip on the timeline (anywhere), and select it's parent ID to this object tracker
3) We choose "Display Clip Image" or "Apply Clip Effects"
   Display Clip Image - we duplicate the clip content + effects and resize it into each matched region (hiding the original clip)
   Apply Clip Effects - we ignore the clip image (causing it to disappear), and apply each effect directly to the matching region pixels

File: src/EffectBase.cpp [hunk 1]
```diff
@@ -150,3 +187,36 @@ openshot::ClipBase* EffectBase::ParentClip() {
...
+	}
+	return;
+}
+
+// Return the ID of this effect's parent clip
+std::string EffectBase::ParentClipId() const{
+	if(clip)
+		return clip->Id();
+	else
+		return "None";
```
jonoomph, line 221: Let's return an empty string instead of "None"

examples/CMakeLists.txt
```diff
@@ -63,3 +63,18 @@ if (BLACKMAGIC_FOUND)
...
+	add_executable(openshot-example-opencv
+			Example_opencv.cpp)
+	
+	target_compile_definitions(openshot-example-opencv PRIVATE
+	-DTEST_MEDIA_PATH="${TEST_MEDIA_PATH}" )
+
+	# Link test executable to the new library
+	target_link_libraries(openshot-example-opencv openshot)
+
+endif()
```
ferdnyc, line 80:
```suggestion

############### OPENCV EXAMPLE ################
if (DEFINED CACHE{HAVE_OPENCV})
	# Create test executable
	add_executable(openshot-example-opencv
			Example_opencv.cpp)
	
	target_compile_definitions(openshot-example-opencv PRIVATE
	-DTEST_MEDIA_PATH="${TEST_MEDIA_PATH}" )

	# Link test executable to the new library
	target_link_libraries(openshot-example-opencv openshot)
endif()
```

`src/CMakeLists.txt` sets `HAVE_OPENCV` to `TRUE` as a `CACHE` variable, when it imports the OpenCV targets. So it's usable at every level of the build, and this way you're depending on whether libopenshot was actually configured to use OpenCV.

src/TrackedObjectBase.h
```diff
@@ -0,0 +1,115 @@
...
+    public:
+		
+		Keyframe visible;
+		Keyframe draw_box;
+
+        /// Blank constructor
+        TrackedObjectBase();
+
+		/// Default constructor
+		TrackedObjectBase(std::string _id);
```
ferdnyc, line 73:
```suggestion
		/// Default constructor
		TrackedObjectBase();

		/// Constructor which takes an object ID
		TrackedObjectBase(std::string _id);
```

The default construtor is one that can be called with no arguments. That is the definition

src/EffectBase.cpp
```diff
@@ -116,14 +123,44 @@ void EffectBase::SetJson(const std::string value) {
...
+		for (auto const& effect : effects){
+			// Set the properties of all effects which parentEffect points to this
+			if (effect->info.parent_effect_id == this->Id())
+				effect->SetJsonValue(root);	
+		}
+	}
+
+	// Set this effect properties with the parent effect properties (except the id and parent_effect_id)
+	if (parentEffect){
+		root = parentEffect->JsonValue();
```
ferdnyc, line 145: This feels very weird to me. So if we imagine I have an `EffectBase* blur` that points to my blur effect, and I call its `SetJson()` like this:
```cpp
blur->SetJson(R"json(
    "iterations": {
      "Points": [
        { "co": { "X": 1.0, "Y": 2.0 }, "interpolation": 1 }
       ] }
   )json");
```
Then whatever data I **actually pass in** to `SetJson()` or `SetJsonValue()` will be _ignored_ if there's a parent effect set, in favor of just re-duplicating whatever the parent effect's settings are?

If `SetJsonValue()` is going to ignore its own arguments when it's called on a parented effect, should it maybe just raise an exception instead? There doesn't seem to be any clear value to calling it. (Or maybe it should have a no-arguments overload that triggers a refresh of the parent effect data, if that's what calling it is intended to achieve.)

src/KeyFrame.cpp
```diff
@@ -31,28 +31,32 @@
...
 #include <utility>
+#include <cassert>         // For assert()
+#include <iostream>        // For std::cout
+#include <iomanip>         // For std::setprecision
 
 using namespace std;
 using namespace openshot;
 
-namespace {
+namespace openshot{
```
ferdnyc, line 44: The reason these functions were placed in an anonymous namespace is that they're not part of the Keyframe API, they're internal utility functions only visible to functions in `KeyFrame.cpp`.

(See https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces)

They shouldn't be moved to `openshot::` or declared in `KeyFrame.h` unless they need to be exposed beyond this file. (Do they?)

src/Timeline.cpp
```diff
@@ -313,6 +420,25 @@ openshot::EffectBase* Timeline::GetClipEffect(const std::string& id)
...
+	for (const auto& clip : clips) {
+		
+		// Get the clip's list of effects
+		std::list<EffectBase*> clipEffects = clip->Effects();
+
+		// Append the clip's effects to the list
+		ClipEffects.insert(ClipEffects.end(), clipEffects.begin(), clipEffects.end());
+	}
+
+	return ClipEffects;
```
ferdnyc, line 439: Please, _PLEASE_ don't use two distinct variables named `ClipEffects` and `clipEffects`, in a function named `ClipEffects`. That's just not OK. Neither variable should be named any permutation of "clip effects".

Call the output list `out` (or any better, but non-confusing, name) and the inner one just `effects`, or something.

src/EffectBase.h
```diff
@@ -73,6 +76,12 @@ namespace openshot
 
 	public:
 
+		/// Parent effect (which properties will set this effect properties)
+		EffectBase* parentEffect;
+
+		/// Map of Tracked Object's by their indices (used by Effects that track objects on clips)
+		std::map<int, std::shared_ptr<TrackedObjectBase> > trackedObjects;
```
ferdnyc, line 83: Something feels off about this, when something so specific to one type of effect has to be a member of the _base_ class for all effects.

Couldn't there be a class that manages this data for effects that do tracking, so that only a pointer or reference to _that_ object is stored in the `EffectBase`?

I dunno, I'm just trying to understand the object hierarchy here, and failing. Especially when you include `TrackedObjectBase` and its **one** derived class. (Why does it need a base class, if there's only one class that derives from it? And if `AddBox` and `GetBoxValues` are present in `TrackedObjectBase`, then the bounding box has been made part of the base class in all but implementation, so the artificial base/derived relationship between `TrackedObjectBase` and `TrackedObjectBBox` feels like a lie, or at a minimum wishful thinking.)

src/EffectBase.h
```diff
@@ -91,12 +100,29 @@ namespace openshot
...
 		void ParentClip(openshot::ClipBase* new_clip);
+		
+		/// Set the parent effect from which this properties will be set to
+		void SetParentEffect(std::string parentEffect_id);
+
+		/// Return the ID of this effect's parent clip
+		std::string ParentClipId() const;
+
+		/// Get the indexes and IDs of all visible objects in the given frame
+		virtual std::string GetVisibleObjects(int64_t frame_number) const {return {}; };
```
ferdnyc, line 111: Here too, "visible" seems like a very specific concept with no generalizable meaning that would warrant its inclusion in the base class. How would a Blur effect or a Brightness effect know what objects are visible in the current frame? What does that even _mean_, for a Blur or Brightness effect?

(Also, the documentation comment says "Get the indexes and IDs", but the method returns a `std::string`. So, feels like there's detail missing there.)

ferdnyc:
@BrennoCaldato 

Please look over the opening comment to #642, I see a lot of documentation comments here that contain the same sort of errors I just finished eliminating from the codebase with that PR.

src/TrackedObjectBBox.cpp
```diff
@@ -0,0 +1,576 @@
...
+	root["x2"] = add_property_json("X2", box.cx+(box.width/2), "float", "", NULL, 0.0, 1.0, false, requested_frame);
+	root["y2"] = add_property_json("Y2", box.cy+(box.height/2), "float", "", NULL, 0.0, 1.0, false, requested_frame);
+
+	// Add the bounding-box Keyframes to the JSON object
+	root["delta_x"] = add_property_json("Displacement X-axis", delta_x.GetValue(requested_frame), "float", "", &delta_x, -1.0, 1.0, false, requested_frame);
+	root["delta_y"] = add_property_json("Displacement Y-axis", delta_y.GetValue(requested_frame), "float", "", &delta_y, -1.0, 1.0, false, requested_frame);
+	root["scale_x"] = add_property_json("Scale (Width)", scale_x.GetValue(requested_frame), "float", "", &scale_x, 0.0, 1.0, false, requested_frame);
+	root["scale_y"] = add_property_json("Scale (Height)", scale_y.GetValue(requested_frame), "float", "", &scale_y, 0.0, 1.0, false, requested_frame);
+	root["rotation"] = add_property_json("Rotation", rotation.GetValue(requested_frame), "float", "", &rotation, 0, 360, false, requested_frame);
+    root["visible"] = add_property_json("Visible", visible.GetValue(requested_frame), "int", "", &visible, 0, 1, false, requested_frame);
```
jonoomph, line 455: Nitpick: indention issues

src/TrackedObjectBBox.h
```diff
@@ -0,0 +1,247 @@
...
+        void clear(); 
+
+        /// Get and Set JSON methods
+        std::string Json() const override;                  ///< Generate JSON string of this object
+        Json::Value JsonValue() const override;             ///< Generate Json::Value for this object
+        void SetJson(const std::string value) override;     ///< Load JSON string into this object
+        void SetJsonValue(const Json::Value root) override; ///< Load Json::Value into this object
+
+        /// Get all properties for a specific frame (perfect for a UI to display the current state
+		/// of all properties at any time)
```
jonoomph, line 233: Nitpick: Indention issue

src/TrackedObjectBase.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <utility>
+
+namespace openshot
+{
+
+	// Blank constructor
+	TrackedObjectBase::TrackedObjectBase() : visible(1.0), draw_box(1)
+	{
+		// Initializes the id as "None"
+		id = "None";
```
jonoomph, line 43: Would "" work here as well? I'm generally opposed to "some-init-string", lol. Thx.

src/TrackedObjectBase.h
```diff
@@ -0,0 +1,115 @@
...
+
+    public:
+		
+		Keyframe visible;
+		Keyframe draw_box;
+
+        /// Blank constructor
+        TrackedObjectBase();
+
+		/// Default constructor
```
jonoomph, line 72: Lots of indention issues here as well. I think we need to eventually convert all these files to spaces only (and get rid of the tabs)

src/effects/Bars.cpp
```diff
@@ -196,6 +196,9 @@ std::string Bars::PropertiesJSON(int64_t requested_frame) const {
 	root["right"] = add_property_json("Right Size", right.GetValue(requested_frame), "float", "", &right, 0.0, 0.5, false, requested_frame);
 	root["bottom"] = add_property_json("Bottom Size", bottom.GetValue(requested_frame), "float", "", &bottom, 0.0, 0.5, false, requested_frame);
 
+	// Set the parent effect which properties this effect will inherit
+	root["parent_effect_id"] = add_property_json("Parent", 0.0, "string", info.parent_effect_id, NULL, -1, -1, false, requested_frame);
```
jonoomph, line 200: Could this be incorporated into the Base class of Effects or Clips? Just curious...

src/effects/ObjectDetection.cpp
```diff
@@ -133,29 +157,32 @@ bool ObjectDetection::LoadObjDetectdData(std::string inputFilePath){
...
-        if (!objMessage.ParseFromIstream(&input)) {
-            cerr << "Failed to parse protobuf message." << endl;
-            return false;
-        }
+    
+    // Read the existing tracker message.
+    fstream input(inputFilePath, ios::in | ios::binary);
+    if (!objMessage.ParseFromIstream(&input)) {
+        cerr << "Failed to parse protobuf message." << endl;
+        return false;
```
jonoomph, line 165: Should this return an openshot::Exception instead? If for some reason, it fails here, I just want to make sure it bubbles up clearly/loudly.

src/effects/ObjectDetection.cpp
```diff
@@ -180,6 +210,26 @@ bool ObjectDetection::LoadObjDetectdData(std::string inputFilePath){
...
+            {
+                // Add a new BBox to it
+                trackedObject->second->AddBox(id, x+(w/2), y+(h/2), w, h, 0.0);
+            } 
+            else
+            {
+                // There is no tracked object with that id, so insert a new one
+                TrackedObjectBBox trackedObj;
+                trackedObj.AddBox(id, x+(w/2), y+(h/2), w, h, 0.0);
+	            std::shared_ptr<TrackedObjectBBox> trackedObjPtr = std::make_shared<TrackedObjectBBox>(trackedObj);
```
jonoomph, line 230: Nitpick: indention

src/effects/ObjectDetection.cpp
```diff
@@ -254,21 +343,46 @@ void ObjectDetection::SetJsonValue(const Json::Value root) {
...
+        selectedObject->SetJsonValue(root);
 }
 
 // Get all properties for a specific frame
 std::string ObjectDetection::PropertiesJSON(int64_t requested_frame) const {
 
 	// Generate JSON properties list
 	Json::Value root;
+
+    // Add the selected object Json to root
```
jonoomph, line 380: Nitpick: Indention, perhaps just scan all these files for indention issues and make them as close to the remaining source code as possible. Thx.

jonoomph (commented):
LG mostly, just lots of nitpicks on indention inconsistencies and string initialization values (i.e. "None"). :+1:

jonoomph:
Last comment, we need to resolve issues against `develop` and merge develop back into this before merge. I think I changed a few things on develop which are conflicting (i.e. some of the `std::cout` statements in the opencv code)

src/TrackedObjectBase.h
ferdnyc: I volunteer as tribute. I will happily run `clang-format` or something across the entire codebase. :wink: 

(There are also a few files of the OpenCV-or-related stuff — like most/all of the contents of `src/sort_filter/` — that snuck in with CRLF line endings. So, at some point we need to run those through `dos2unix`.)

src/effects/ObjectDetection.cpp
ferdnyc: @jonoomph I guess my question would be, how common a failure is it?

IOW, is it truly an exceptional condition? Or are there cases where it's "normal" that we can't parse certain files, for reasons? (In which instance, I'd argue for _at most_ a debug log message, and even then only if we **also** log successful parse attempts.)

If it truly is an exceptional this-shouldn't-happen, then agreed: 
```cpp
// 😝
return false;
// 🤩
throw ProtobufParseError("Failed to parse protobuf message", inputFilePath);
```

ferdnyc:
@jonoomph It doesn't help that I merged a complete rewrite of all the unit tests yesterday, as well. If this PR touches them (which it looks like it does, based on the `tests/KeyFrame_Tests.cpp` conflict) then the changes here will need to be ported to Catch2. I'll take a look at that particular file now.

ferdnyc:
OK, I hand-resolved all of the conflicts that resulted from merging `develop` into this branch. I **HOPE** correctly, though it got squirrely at some points. I'd definitely appreciate additional eyes on the changes.

In the process I also ported the KeyFrame unit test changes to Catch2. I'll leave some review comments there in a few minutes.

Happily, Sublime Merge's smart indentation cleaned up some of the tab/space issues, in the process of resolving conflicting chunks of code. Whenever I'd copy some lines from an input file that didn't follow the destination file's conventions, it silently fixed them up as I pasted them. **#Neat**

src/effects/Tracker.h
```diff
@@ -41,50 +41,31 @@
...
-      rotation = _rotation;
-      x1 = _x1;
-      y1 = _y1;
-      x2 = _x2;
-      y2 = _y2;
-  }
-};
-
-
+using namespace std;
```
ferdnyc, line 47: :no_entry_sign: Never, **ever** in a header file. See #296, which has links to additional references on the issue.
```suggestion
```

tests/KeyFrame.cpp
```diff
@@ -506,3 +513,195 @@ TEST_CASE( "Point_Vector_Constructor", "[libopenshot][keyframe]" )
 	CHECK(k1.GetLength() == 11);
 	CHECK(k1.GetValue(10) == Approx(30.0f).margin(0.0001));
 }
+
+TEST_CASE( "TrackedObjectBBox init", "[libopenshot][keyframe]" )
+{
+	TrackedObjectBBox kfb;
+	// XXX: This really needs to perform some sort of _test_ here,
+	//      like confirming some default values in the new object.
+}
```
ferdnyc, line 522: As the comment says, this test case really needs to _test_ something, for there to be any point to it. Otherwise, all of the _other_ test cases that start off by creating a default `TrackedObjectBBox` object are already performing the same test, making this redundant.

tests/KeyFrame.cpp
```diff
@@ -506,3 +513,195 @@ TEST_CASE( "Point_Vector_Constructor", "[libopenshot][keyframe]" )
...
+	CHECK(kfb.Contains(1) == true);
+	CHECK(kfb.GetLength() == 1);
+
+	kfb.RemoveBox(1);
+
+	CHECK_FALSE(kfb.Contains(1));
+	CHECK(kfb.GetLength() == 0);
+}
+
+TEST_CASE( "TrackedObjectBBox GetVal" "[libopenshot][keyframe]" )
```
ferdnyc, line 539: Whoops, my typo.
```suggestion
TEST_CASE( "TrackedObjectBBox GetVal", "[libopenshot][keyframe]" )
```

tests/KeyFrame.cpp
```diff
@@ -506,3 +513,195 @@ TEST_CASE( "Point_Vector_Constructor", "[libopenshot][keyframe]" )
...
 }
+
+TEST_CASE( "TrackedObjectBBox init", "[libopenshot][keyframe]" )
+{
+	TrackedObjectBBox kfb;
+	// XXX: This really needs to perform some sort of _test_ here,
+	//      like confirming some default values in the new object.
+}
+
+TEST_CASE( "TrackedObjectBBox AddBox and RemoveBox" "[libopenshot][keyframe]" )
```
ferdnyc, line 524:
```suggestion
TEST_CASE( "TrackedObjectBBox AddBox and RemoveBox", "[libopenshot][keyframe]" )
```

tests/KeyFrame.cpp
```diff
@@ -506,3 +513,195 @@ TEST_CASE( "Point_Vector_Constructor", "[libopenshot][keyframe]" )
...
+	// Change trackedData scale
+	trackedData->scale_x.AddPoint(1, 2.0);
+	CHECK(trackedData->scale_x.GetValue(1) == 2.0);
+
+	// Tracked Data JSON
+	auto trackedDataJson = trackedData->JsonValue();
+
+	// Get and cast the trakcedObject
+	auto trackedObject_base = t.GetTrackedObject("None");
+	std::shared_ptr<TrackedObjectBBox> trackedObject = std::static_pointer_cast<TrackedObjectBBox>(trackedObject_base);
```
ferdnyc, line 672: That's a really long line, which I feel would be cleaner split in two.

Also, [according to cppreference](https://en.cppreference.com/w/cpp/memory/shared_ptr/pointer_cast#Example) `std::static_pointer_cast` is for casting _up_ the inheritance tree, while `std::dynamic_pointer_cast` should be used when casting _down_ or _sideways_. So, to cast a `TrackedObjectBase` to a `TrackedObjectBBox`, this should be:
```suggestion
	auto trackedObject = std::make_shared<TrackedObjectBBox>();
	trackedObject = std::dynamic_pointer_cast<TrackedObjectBBox>(trackedObject_base);
```

tests/KeyFrame.cpp
```diff
@@ -506,3 +513,195 @@ TEST_CASE( "Point_Vector_Constructor", "[libopenshot][keyframe]" )
...
+
+	// Set trackedObject Json Value
+	trackedObject->SetJsonValue(trackedDataJson);
+
+	// Attach childClip to tracked object
+	std::string tracked_id = trackedData->Id();
+	childClip.Open();
+	childClip.AttachToObject(tracked_id);
+	
+	std::shared_ptr<TrackedObjectBBox> trackedTest = std::static_pointer_cast<TrackedObjectBBox>(childClip.GetAttachedObject());
```
ferdnyc, line 683: Ditto, re: line length and `dynamic_pointer_cast`.
```suggestion
	auto trackedTest = std::make_shared<TrackedObjectBBox>();
	trackedTest = std::dynamic_pointer_cast<TrackedObjectBBox>(childClip.GetAttachedObject());
```

tests/KeyFrame.cpp
```diff
@@ -506,3 +513,195 @@ TEST_CASE( "Point_Vector_Constructor", "[libopenshot][keyframe]" )
...
+	childClip.Close();
+	// XXX: Here, too, there needs to be some sort of actual _testing_ of the results
+}
+
+TEST_CASE( "GetBoxValues", "[libopenshot][keyframe]" )
+{
+	TrackedObjectBBox trackedDataObject;
+	trackedDataObject.AddBox(1, 10.0, 10.0, 20.0, 20.0, 30.0);
+
+	std::shared_ptr<TrackedObjectBase> trackedData = std::make_shared<TrackedObjectBBox>(trackedDataObject);
```
ferdnyc, line 697: Why are you storing this as a `TrackedObjectBase` pointer immediately after creating it as a `TrackedObjectBBox`? (If the mismatch was unintentional, I'd say just replace the declared type with `auto` and it's solved.)

tests/KeyFrame.cpp
```diff
@@ -506,3 +513,195 @@ TEST_CASE( "Point_Vector_Constructor", "[libopenshot][keyframe]" )
...
+	childClip.Open();
+	childClip.AttachToObject(tracked_id);
+	
+	std::shared_ptr<TrackedObjectBBox> trackedTest = std::static_pointer_cast<TrackedObjectBBox>(childClip.GetAttachedObject());
+	
+	CHECK(trackedData->scale_x.GetValue(1) == trackedTest->scale_x.GetValue(1));
+	
+	auto frameTest = childClip.GetFrame(1);
+	childClip.Close();
+	// XXX: Here, too, there needs to be some sort of actual _testing_ of the results
```
ferdnyc, line 689: Again, like the comment I added says, there should really be some sort of _tests_ done on `frameTest`, after it's returned by `GetFrame()`.

ferdnyc (commented):
Mostly things I noticed while resolving the `develop` merge.

tests/KeyFrame.cpp
ferdnyc: I'mma commit these two typo fixes, my bad.

ferdnyc:
So, I have a general question: How does all of this tracking stuff being added to the Clip and Effect base classes affect the the optional status of OpenCV, as a dependency? I haven't noticed `#ifdef USE_OPENCV` blocks around most of these changes...

BrennoCaldato:
> So, I have a general question: How does all of this tracking stuff being added to the Clip and Effect base classes affect the the optional status of OpenCV, as a dependency? I haven't noticed `#ifdef USE_OPENCV` blocks around most of these changes...
> 
> **EDIT:** Ah, I guess the current build failures answer my question, since Ubuntu 18.04 doesn't have a compatible OpenCV available, and the attempts to `#include` the autogenerated Protobuf headers when none were created are now breaking the build.

You are correct @ferdnyc, I couldn't have enough time yet to check and test if all the new additions are being covered by #ifdef USE_OPENCV but I will do that when I finish with the new additions.

As a side note, I have a few commits that I haven't pushed yet related to the ObjectDetector effect. I have been facing some issues when adding the ability to attach other clips to the detected objects. I especially concerned about it being too slow if we have a lot of detected objects

ferdnyc:
@BrennoCaldato 
> You are correct @ferdnyc, I couldn't have enough time yet to check and test if all the new additions are being covered by #ifdef USE_OPENCV but I will do that when I finish with the new additions.

I do not envy you that task.

In terms of these _particular_ failures, though, the build breaks when `TrackedObjectBBox.h` includes the protobuf headers — but there's nothing in that header that **uses** anything from those headers. All of the protobuf-related `#include`s, as well as the `using google::protobuf::util::TimeUtil;` statement, should be moved to `TrackedObjectBBox.cpp`, and then at least the header won't break the compile of any file that includes _it_. (And once they're `#ifdef`'d, hopefully nothing breaks at all.)

The header itself shouldn't even require any `#ifdef`s, I don't think, since I don't see any OpenCV/protobuf types or symbols anywhere in `TrackedObjectBBox.h`. It looks like they're constrained to the corresponding implementation. (Which is excellent news for the sanity of any future developers who find themselves reading the finished form of this code. `#ifdef` spaghetti is a nightmare. This OTOH looks like it should be relatively clean and self-contained, so you have my thanks.)

> As a side note, I have a few commits that I haven't pushed yet related to the ObjectDetector effect. I have been facing some issues when adding the ability to attach other clips to the detected objects. I especially concerned about it being too slow if we have a lot of detected objects

I can certainly imagine that might get a bit draggy, if there's a lot going on and all of that data needs to be retained.

....Although: _Does_ it? Hear me out here: In at least _some_ uses of the object detection effect, especially those that involve attaching Clips to objects, not all detected objects are equally relevant to the user. The ones with which they've defined some sort of interaction (like attaching a Clip) are kind of the only ones they care about — all of the other detection data can really be discarded. Why track an object if it's going to be ignored for its entire lifetime?

Which points me down two possible optimization paths, though you'd have to say whether either of them would be feasible to implement, and whether there would be any practical benefit:
1. Perhaps the detection/tracking effects could operate in a two-stage/two-pass mode, somewhat similar to Audacity's noise removal effect? (Oooh, in fact the more I think about it, the more apt that analogy really feels.)

    If you haven't used the effect, it's basically an audio cleanup effect that pulls out the background/underlying noise that can permeate a recording — tape hiss, environmental noise pollutiion, etc. Because each recording's noise floor is unique, there's a two-step process to using the effect: The first thing you do is collect a "noise profile" by having it scan a few frames  of "silence" in the recording, when _only_ the noise is audible. Then, once it knows what's not supposed to be there, you run it over the entire audio and it filters the waveform based on the noise profile it previously captured.
    
    Could the Object Detection effect work in a similar manner, and sparate its functionality into two distinct modes/passes? First, an initial scanning & data-collection mode where it processes all detected objects and presents full data on them for consumption by the caller. That, of course, can be slow at times, especially in complex scenes. But then when it comes time to render the effect's results for output, the detection data for any "unused" objects (ones which the project data never references) can hopefully just be discarded since it won't be used for anything.

1. Building on the previous thought, is it possible within the OpenCV API to have the _detection itself_ "focus" on just certain specific objects? (Or if not detection, then at least tracking? That is, if you aren't already limiting tracking to only the relevant objects.) Or is it an all-or-nothing situation, you just get data for every object in play?

    My thinking was, if we can deem certain objects as relevant and the rest as unimportant, is it possible to configure the detection/tracking algorithms with that information? And then have them only return tracking data for those objects in particular, while ignoring all the rest? Rather than discarding data for objects we're not interested in, it would hopefully be even _more_ efficient if that data was never produced in the first place, because the algorithms themselves were told which objects to focus on.

ferdnyc:
So, currently two of the unit tests are failing: There's a segfault during the KeyFrame  "TrackedObjectBBox GetVal interpolation" test, and the "Track_Video" test in CVTracker is failing because it's getting all `0` data returned to it.

It's entirely possible I caused this by messing something up during the merge of `develop`, I'm not denying that. I took a quick look, but I'm not immediately sure what's causing the failures.

Running the test executables directly gives more detailed output on the failures:
```console
$ build/tests/openshot-KeyFrame-test 
openshot-KeyFrame-test: /home/ferd/rpmbuild/REPOS/libopenshot/src/KeyFrame.cpp:99:
 double openshot::InterpolateBetween(const openshot::Point&, const openshot::Point&, 
double, double): Assertion `target <= right.co.X' failed.

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
openshot-KeyFrame-test is a Catch v2.11.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
TrackedObjectBBox GetVal interpolation
-------------------------------------------------------------------------------
/home/ferd/rpmbuild/REPOS/libopenshot/tests/KeyFrame.cpp:554
...............................................................................

/home/ferd/rpmbuild/REPOS/libopenshot/tests/KeyFrame.cpp:554: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases:  28 |  27 passed | 1 failed
assertions: 176 | 175 passed | 1 failed

zsh: IOT instruction (core dumped)  build/tests/openshot-KeyFrame-test

$ build/tests/openshot-CVTracker-test 
[ INFO:0] global /builddir/build/BUILD/opencv-4.3.0/modules/core/src/ocl.cpp (891) 
haveOpenCL Initialize OpenCL runtime...
[ INFO:0] global /builddir/build/BUILD/opencv-4.3.0/modules/core/src/ocl.cpp (433) 
OpenCLBinaryCacheConfigurator Successfully initialized OpenCL cache directory: 
/home/ferd/.cache/opencv/4.3/opencl_cache/
[ INFO:0] global /builddir/build/BUILD/opencv-4.3.0/modules/core/src/ocl.cpp (457) 
prepareCacheDirectoryForContext Preparing OpenCL cache configuration for context: 
NVIDIA_Corporation--GeForce_GT_710--460_67

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
openshot-CVTracker-test is a Catch v2.11.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
Track_Video
-------------------------------------------------------------------------------
/home/ferd/rpmbuild/REPOS/libopenshot/tests/CVTracker.cpp:45
...............................................................................

/home/ferd/rpmbuild/REPOS/libopenshot/tests/CVTracker.cpp:75: FAILED:
  CHECK( (int)(x * 640) == 259 )
with expansion:
  0 == 259 (0x103)

/home/ferd/rpmbuild/REPOS/libopenshot/tests/CVTracker.cpp:76: FAILED:
  CHECK( (int)(y * 360) == 131 )
with expansion:
  0 == 131

/home/ferd/rpmbuild/REPOS/libopenshot/tests/CVTracker.cpp:77: FAILED:
  CHECK( (int)(width * 640) == 180 )
with expansion:
  0 == 180

/home/ferd/rpmbuild/REPOS/libopenshot/tests/CVTracker.cpp:78: FAILED:
  CHECK( (int)(height * 360) == 166 )
with expansion:
  0 == 166

===============================================================================
test cases: 2 | 1 passed | 1 failed
assertions: 8 | 4 passed | 4 failed
```

BrennoCaldato:
> ....Although: _Does_ it? Hear me out here: In at least _some_ uses of the object detection effect, especially those that involve attaching Clips to objects, not all detected objects are equally relevant to the user. The ones with which they've defined some sort of interaction (like attaching a Clip) are kind of the only ones they care about — all of the other detection data can really be discarded. Why track an object if it's going to be ignored for its entire lifetime?
> 
> Which points me down two possible optimization paths, though you'd have to say whether either of them would be feasible to implement, and whether there would be any practical benefit:

@ferdnyc Both options are possible to implement. 
Let me just do a brief overview of the ObjectDetection effect. Currently, when the user applies ObjectDetection to a clip the first step is to detect and track the objects, and then serialize the bounding boxes to a protobuf message. This is done at CVObjectDetection. The protobuf message size is quite small (around 100kb for the test video I use).
The message is then deserialized at the ObjectDetection effect and a TrackedObjectBBox is created for every box in the message.

Every tracked bounding box was saved with protobuf to minimize the need of reapplying the effect. Because the preprocessing can be quite slow depending on the system. The user can further select in the UI the desired class and confidence threshold. (I believe this is more aligned with your suggestion 2.)

I uploaded a video in a non-listed playlist to demonstrate how the effect is operating right now

https://youtu.be/0dTK7dorFCU

Even so, in the example, @jonoomph cited, if we parent a blurred clip to each one of the detected persons it shouldn't be very fast. Especially because I'm composing the frames with qt::drawImage

jonoomph:
@BrennoCaldato Is it possible to specify the visibility of object labels? For example, if I wanted to outline all "truck" objects, but not display the word "Truck 001 0.85"? Seems like an easy addition, and very useful for certain cases. Thx for all this hard work though, don't want that to get minimized in all the feedback. This is very awesome and impressive stuff, with a ton of future potential!

jonoomph:
I also think all the @ferdnyc suggestions sound great, to clean up headers, make the optional opencv builds succeed, etc...

ferdnyc:
Two other things I noticed while I was resolving conflicts, but it's not actually in the code that's part of _this_ change... take this section of `CVTracker.cpp` as an example:

https://github.com/OpenShot/libopenshot/blob/8cefd49abf95901f7233d23fc6e04e2f9ea6e909/src/CVTracker.cpp#L70-L85

1. This code is using `0`-based frame numbers. (It was actually more obvious in the code that _calls_ this code, as it explicitly passed `0` and `20` as the start and end frame numbers.) Problem is, libopenshot starts counting from `1`. If you call `GetFrame(0)`, you'll silently<sup><b>\*</b></sup> be handed frame 1 instead. Then, next iteration, you'll call `GetFrame(1)` and be handed the same frame 1 _again_.
2. Bigger issue, the `ToInt()s` there just scream **#DangerZone**. If you treat the FPS as an _integer_ when computing the start frame, then you're going to be off by 1 frame for every 1000 when the frame rate is `29.97 fps` (AKA `30000 / 1001`), or any other non-integer rate. (Non-integer frame rates are very common.)

\* — (I personally think requests for frame 0 should throw an exception instead, but it's not my code.)

BrennoCaldato:
@ferdnyc 

After merging with your last commits I am facing an issue with cmake

`Could not find a package configuration file provided by Catch2`

So I tried to install it on my system with:

    $ git clone https://github.com/catchorg/Catch2.git
    $ cd Catch2
    $ cmake -Bbuild -H. -DBUILD_TESTING=OFF
    $ sudo cmake --build build/ --target install 

But I still get
`catch2/catch.hpp: No such file or directory`

Do you have any instructions on this?

src/KeyFrame.cpp
BrennoCaldato: I am using the same functions at the TrackedObjectBBox class. To not duplicated code I just included them in the openshot namespace. But if you prefer I can change that

src/TrackedObjectBase.h
BrennoCaldato: I fixed already some of the files, @ferdnyc. But certainly, there are more of these. I will double-check as well

jonoomph:
@BrennoCaldato Catch2 installation instructions (on Ubuntu 18.04 or 20.04):
```
wget https://launchpad.net/ubuntu/+archive/primary/+files/catch2_2.13.0-1_all.deb
sudo dpkg -i catch2_2.13.0-1_all.deb
```

src/effects/Tracker.cpp
```diff
@@ -72,96 +91,159 @@ void Tracker::init_effect_details()
...
-            return false;
-        }
-    }
+		// Set a Qt painter to the frame image
+		QPainter painter(&frameImage);
 
-    // Make sure the trackedData is empty
-    trackedDataById.clear();
+		// Draw the child clip image inside the bounding-box
+		painter.drawImage(boxRect, *childClipImage, QRectF(0, 0, frameImage.size().width(),  frameImage.size().height()));
```
ferdnyc, line 178:
```suggestion
		painter.drawImage(boxRect, *childClipImage, QRectF(frameImage.rect()));
```

(Though, if you're drawing the entire image, I don't _think_ this is necessary. If I'm not mistaken, QImage's `drawImage()` defaults to drawing the entire source image in the area specified, scaled accordingly. So  that would make this equivalent):
```c++
painter.drawImage(boxRect, *childClipImage);
```

BrennoCaldato:
@jonoomph following our last call I am merging this branch with develop. It's a big one. Fingers crossed

PR #652, opened 2021-03-31
Timeline: Add a new constructor that takes a ReaderInfo struct and copies parameters

ferdnyc:
This PR was inspired when I read this line in the latest `Example.cpp`:
```cpp
Timeline tm(r9.info.width, r9.info.height, r9.info.fps, r9.info.sample_rate, r9.info.channels, r9.info.channel_layout);
```

"Well, that's not using the compiler very efficiently.", I thought to myself. "Nobody should ever have to type that line more than once."

So, I typed that line once, as a new delegating constructor for the `Timeline` class. Its entire code:

```cpp
Timeline::Timeline(const ReaderInfo info) :
	Timeline::Timeline(info.width, info.height, info.fps, info.sample_rate,
	                   info.channels, info.channel_layout) {};
```

As a result, that line in `Example.cpp` becomes:
```cpp
Timeline tm(r9.info);
```

And effectively, the compiler transforms it into the code that used to be there. Easy-peasy. A new unit test is also added for the constructor.

### DIFF
--- a/examples/Example.cpp
+++ b/examples/Example.cpp
@@ -31,8 +31,10 @@
 #include <fstream>
 #include <iostream>
 #include <memory>
-#include "OpenShot.h"
-#include "CrashHandler.h"
+#include "Clip.h"
+#include "Frame.h"
+#include "FFmpegReader.h"
+#include "Timeline.h"
 
 using namespace openshot;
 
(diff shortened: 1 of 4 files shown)


ferdnyc:
Merging this, because then I have to port the test to Catch2.

PR #653, opened 2021-04-09
FrameMapper Audio Crackles/Popping Fixes

jonoomph:
Added new unit test which distributes audio samples between many different frame rates (30/1, 24/1, 30000/1001, 119/4), and fixes a huge issue with mapping frame numbers incorrectly causing audio crackles/pops. Also fixes a bug which causes crashes on NON-STEREO channel layouts.

For example, when using a source video with a 119/4 frame rate (29.75 fps) and converting it to 30 fps (and positioning the Clip at position 3.77 seconds), it would generate a huge amount of high frequency pops, due to incorrect frame # adjustments inside FrameMapper.

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -1669,10 +1669,8 @@ void FFmpegWriter::write_audio_packets(bool is_final) {
             av_opt_set_int(avr, "out_channels", info.channels, 0);
             SWR_INIT(avr);
         }
-        int nb_samples = 0;
-
         // Convert audio samples
-        nb_samples = SWR_CONVERT(
+        int nb_samples = SWR_CONVERT(
             avr,    // audio resample context
             audio_converted->data,           // output data pointers
             audio_converted->linesize[0],    // output plane size, in bytes. (0 if unknown)
@@ -1683,7 +1681,7 @@ void FFmpegWriter::write_audio_packets(bool is_final) {
         );
 
         // Set remaining samples
-        remaining_frame_samples = nb_samples * av_get_bytes_per_sample(AV_SAMPLE_FMT_S16);
+        remaining_frame_samples = total_frame_samples;
 
         // Create a new array (to hold all resampled S16 audio samples)
         all_resampled_samples = (int16_t *) av_malloc(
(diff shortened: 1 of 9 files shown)


src/FFmpegWriter.cpp
```diff
@@ -1683,7 +1681,7 @@ void FFmpegWriter::write_audio_packets(bool is_final) {
         );
 
         // Set remaining samples
-        remaining_frame_samples = nb_samples * av_get_bytes_per_sample(AV_SAMPLE_FMT_S16);
+        remaining_frame_samples = total_frame_samples;
```
jonoomph, line 1684: This fixes a bug where MONO channel layout crashes libopenshot

src/FrameMapper.cpp
```diff
@@ -263,8 +273,9 @@ void FrameMapper::Init()
 
 			while (remaining_samples > 0)
 			{
-				// get original samples
-				int original_samples = Frame::GetSamplesPerFrame(AdjustFrameNumber(end_samples_frame), original, reader->info.sample_rate, reader->info.channels) - end_samples_position;
+				// Get original samples (with NO framerate adjustments)
+				// This is the original reader's frame numbers
+				int original_samples = Frame::GetSamplesPerFrame(end_samples_frame, original, reader->info.sample_rate, reader->info.channels) - end_samples_position;
```
jonoomph, line 278: Removing `AdjustFrameNumber()` method from original samples calculation, since this access the original reader, and is always in original frame #'s... never mapped to the Timeline offset.

BrennoCaldato: Great! Does this also fix the export issue? When we were getting deformed waveforms when exporting to different FPS?

jonoomph: Probably! But I would love to test that again. The new unit test had hundreds of thousands of failures without this fix. And zero failures with it.  But I didn't notice the failures until the Timeline merged the audio, and some frames had the wrong # of samples, and thus introduced some high frequency pops.

ferdnyc:
@jonoomph 
The new unit test — like the ones it's based on — leaks memory like a sieve, since it's allocating audio buffers in a loop, but never frees any of that memory. `new` without `delete` is always a red flag.

There needs to be a `delete[] audio_buffer;` at the end of each loop iteration, after the samples are copied into the `Frame` objects using `AddAudio()`.

I fixed all of them while I was porting the tests to Catch2, so when that gets merged (hopefully ASAP, just had some fine-tuning to do) it'll be sorted out. Just an FYI.

PR #660, opened 2021-04-19
CMake: Add separate unit test targets

ferdnyc:
Each unit test file <code><var>ClassName</var>.cpp</code>, which results in the creation
of a test target <code>openshot-<var>ClassName</var>-test</code>, will now also be labeled
with the CTest label <code><var>ClassName</var></code>, and a target <code><var>ClassName</var>_coverage</code> will be generated that runs only the tests in that file.

Just like the main `coverage` target, the target name <code><var>ClassName</var>_coverage</code> is just a generic name. Running the target will collect coverage data if `ENABLE_COVERAGE` is set, but will run the tests without coverage if it is left disabled. It does not imply collection of coverage data.

The separate targets are especially useful when developing tests for a class, as the tests in the class being worked on can be re-run without having to wait a minute or more while the other tests run.

#### Example workflow
```sh
# Configure project
cmake -B build -S .
# Work on tests for MyClass
vim tests/MyClass.cpp
# Build and run only the tests for MyClass
cmake --build build --target MyClass_coverage
```

(The first time through, libopenshot itself will still have to be built before the tests can run. But in subsequent executions of `cmake --build build --target MyClass_coverage`, the build will be quick because the only file that will be recompiled is `tests/MyClass.cpp`. When running the tests, the very lengthy runtimes of certain tests in `FrameMapper`, `FFmpegWriter`, and `CVTracker` are avoided.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,8 +178,9 @@ find_package(Catch2 QUIET)
 if(NOT Catch2_FOUND)
   set(CAN_BUILD_TESTS FALSE)
 endif()
-if(CAN_BUILD_TESTS)
-  enable_testing()
+if(CAN_BUILD_TESTS AND BUILD_TESTING)
+  include(CTest)
+  include(Catch)
   if(ENABLE_PARALLEL_CTEST)
     # Figure out the amount of parallelism for CTest
     include(ProcessorCount)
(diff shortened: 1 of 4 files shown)


ferdnyc:
> When running the tests, the very lengthy runtimes of certain tests in FrameMapper, FFmpegWriter, and CVTracker are avoided.

In fact, the labeling has an unexpected side benefit: This summary, which CTest displays at the end of the run if the full suite of tests are enabled:

```text
100% tests passed, 0 tests failed out of 135

Label Time Summary:
CVStabilizer     =   3.56 sec*proc (2 tests)
CVTracker        =  62.05 sec*proc (2 tests)
CacheDisk        =   6.82 sec*proc (6 tests)
CacheMemory      =  10.67 sec*proc (11 tests)
Clip             =   7.18 sec*proc (6 tests)
Color            =   8.81 sec*proc (9 tests)
Coordinate       =   5.10 sec*proc (5 tests)
DummyReader      =   7.08 sec*proc (7 tests)
FFmpegReader     =  14.32 sec*proc (8 tests)
FFmpegWriter     =  10.08 sec*proc (2 tests)
Fraction         =   4.33 sec*proc (5 tests)
Frame            =   5.55 sec*proc (5 tests)
FrameMapper      =  38.04 sec*proc (13 tests)
ImageWriter      =   5.00 sec*proc (1 test)
KeyFrame         =  23.89 sec*proc (24 tests)
Point            =  13.41 sec*proc (11 tests)
QtImageReader    =   3.54 sec*proc (3 tests)
ReaderBase       =   1.14 sec*proc (1 test)
Settings         =   1.96 sec*proc (2 tests)
Timeline         =  14.52 sec*proc (12 tests)

Total Test time (real) =  63.19 sec
```

So we can now see at a glance that `CVTracker` is by far the most time-consuming test (as one would expect, really, given the complexity of its function), but `FrameMapper` is shockingly the second-place occupant on that list, which feel worth investigating.

The long runtime may simply be a result of test thoroughness and/or complexity. (That's certainly why `KeyFrame` takes so long. It runs 24 separate test cases, almost twice as many as `FrameMapper`.) But long runtimes may also may be a sign of inefficiencies — either in the tests, or the class itself.

ferdnyc:
Merging this, as it's relatively inconsequential unless used.

(Except for a switch I snuck in, to use the standard CTest variable `BUILD_TESTING` as the preferred way to enable/disable unit tests. It's on by default, but can be switched off to disable all unit test / coverage targets. The old `DISABLE_TESTS` _and_ `ENABLE_TESTS` variables will be forwarded to `BUILD_TESTING` if they're set.)

PR #662, opened 2021-04-23
Markdown fixes, INSTALL-LINUX partial update

ferdnyc:
I got sick of Codacy complaining about the Markdown syntax, so I ran all of the `*.md` files through the `remark` formatter to ensure they're compliant.

Then I noticed how out of date `docs/INSTALL-LINUX.md` had gotten, so I at least partially updated it. It still doesn't mention OpenCV or Protobuf (though it does include them in the dependency install commands provided), but at least it covers Catch2/CTest now, and no longer mentions UnitTest++. 

CMake/apt/etc. command lines and some of the configurables mentioned have also been freshened up.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -24,6 +24,7 @@ Instructions on how to install these dependencies vary for each operating system
 Libraries and executables have been labeled in the list below to help distinguish between them.
 
 #### FFmpeg (libavformat, libavcodec, libavutil, libavdevice, libavresample, libswscale)
+
 *   <http://www.ffmpeg.org/> **(Library)**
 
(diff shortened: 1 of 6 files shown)


ferdnyc:
Grrr. When updating `docs/INSTALL-LINUX.md`, I wish I'd remembered that a lot of the same ground had _already_ been covered in `/INSTALL.md`, which I should've been updating instead.

...Also, why do we have the same information duplicated across four different documents, but presented with *slight* differences each time? (In ways that have nothing to do with the Linux/Mac/Windows split.)

ferdnyc:
Well, at least Codacy's happy.

Screenshot from 2021-04-23 14-52-59

PR #663, opened 2021-04-24
Defense Against the Dark OpenCV 4.5.1

ferdnyc:
Because OpenCV 4.5.1 is irrevocably broken with regards to the APIs we need, when we detect that version we simply have to disable it entirely in the build. These CMake changes do that, as well as preparing us to support OpenCV 4.5.2+ (which is mostly being done in #639), and exporting the detected OpenCV version into the library as the symbol `OPENCV_VERSION_STR`, for reference purposes.


ferdnyc:
Passed CI, merging to unbreak our PPA builds on Ubuntu Impish (now w/ broken OpenCV 4.5.1!)

PR #667, opened 2021-04-27
Settings: Support LIBOPENSHOT_DEBUG envvar

ferdnyc:
This was originally part of my PR #593, but that was too ambitious and is stuck on hold for the moment. So, I broke this useful feature out into its own PR. With this change, setting `LIBOPENSHOT_DEBUG` to any value in the process environment directs the ZeroMQ debug logging stream to stderr (in addition to any other file/subscriber destinations).


ferdnyc:
I also added a unit test for the feature (which involved working out how to have CTest configure the environment for certain unit tests), and added a `VERBOSE_TESTS` option to the CMake configuration. It defaults `OFF`, but if enabled on the command line will run `ctest` with the `-VV` argument, causing it to output detailed setup/execution information as each test is run. (It also shows the stdout/stderr of the tests themselves, which is normally suppressed.)

ferdnyc:
With testing in place, I'm going to merge this. It's handy to have, and shouldn't affect anything else unless it's used.

PR #670, opened 2021-05-02
(WIP) Patches to let libopenshot compile with ffmpeg 4.4

eisneinechse:
This is a work in progress. The export of the audio is still broken
which started on June 18 with the removal of deprecated call in
ffmpeg.


src/FFmpegWriter.cpp
```diff
@@ -2046,7 +2059,11 @@ bool FFmpegWriter::write_video_packet(std::shared_ptr<Frame> frame, AVFrame *fra
 		pkt.flags |= AV_PKT_FLAG_KEY;
 		pkt.stream_index = video_st->index;
 		pkt.data = (uint8_t *) frame_final->data;
+#if IS_FFMPEG_4_4
+		pkt.size = sizeof(AVFrame);
+#else
 		pkt.size = sizeof(AVPicture);
+#endif
```
ferdnyc: Hah! Well, crap. I only just noticed that this 4-month-old comment was still "Pending", and only visible to me! Whoops. Sorry about that.

----

Instead of directly setting `packet.size`, should this be using `av_packet_from_data(&pkt, buffer, size)`? 

Or, in the cases where the data isn't in an `av_malloc()`'d `buffer` like `_from_data()` requires (i.e. audio), at the very least an `av_new_packet(&pkt, size)` followed by a `memcpy()`?

`av_init_packet()` has now been deprecated, a change that was _formally_ made only recently... but in part, it's been deprecated because there have been better replacements for a very long time. Both `av_new_packet()` and `av_packet_from_data()` were available as far back as FFmpeg 2.4. So, we could switch our existing code away from it for all FFmpeg versions, not just 4.4+.

----

...Ultimately that's what I ended up doing in #784, since `av_packet_from_data` has "always" (as far as we're concerned) been available. So now we  just leave it to avcodec (or is it avformat? avutil?) to figure out how to size things. I figure it should know best!

src/FFmpegWriter.cpp
```diff
@@ -236,7 +236,11 @@ void FFmpegWriter::SetVideoOptions(bool has_video, std::string codec, Fraction f
 			info.vcodec = new_codec->name;
 
 			// Update video codec in fmt
+#if IS_FFMPEG_4_4			
+//			fmt->video_codec = new_codec->id;
```
JacksonRG, line 240: Does the codec not need to be set in this case?

src/FFmpegWriter.cpp
ferdnyc: Sorry, no. That change was in #809.

eisneinechse:
Was already resolved by another patch

eisneinechse:
Was already resolved by another patch

PR #672, opened 2021-05-06
 Linux builder: Don't force bundled jsoncpp

ferdnyc:


ferdnyc:
With CMake on the builder upgraded to version 3.20.x, this is no longer necessary, and it's not the _worst_ idea to have our builders testing compilation with the bundled sources, since we're bundling them. Closing.

PR #673, opened 2021-05-14
fix cmake instructions using '-d' argument in MacOs

wasimafser:
Found it to be a problem while setting up.

### DIFF
--- a/doc/INSTALL-MAC.md
+++ b/doc/INSTALL-MAC.md
@@ -151,11 +151,12 @@ brew install zeromq
 Since libopenshot-audio is not available in a Homebrew or MacPorts package, we need to go through a 
 few additional steps to manually build and install it. Launch a terminal and enter:
 
-```
+```zsh
 cd [libopenshot-audio repo folder]
 mkdir build
 cd build
-cmake --debug-output -G "Unix Makefiles" -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang ../   (CLang must be used due to GNU incompatible Objective-C code in some of the Apple frameworks)
+# (CLang must be used due to GNU incompatible Objective-C code in some of the Apple frameworks)
+cmake --debug-output -G "Unix Makefiles" -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang ../
 make
 make install
 ./src/openshot-audio-demo  (This should play a test sound)
@@ -164,11 +165,11 @@ make install
 ## Mac Build Instructions (libopenshot)
 Run the following commands to build libopenshot:
 
-```
-$ cd [libopenshot repo folder]
-$ mkdir build
-$ cd build
-$ cmake --debug-output -G "Unix Makefiles"  -DCMAKE_CXX_COMPILER=/usr/local/opt/gcc48/bin/g++-4.8 -DCMAKE_C_COMPILER=/usr/local/opt/gcc48/bin/gcc-4.8 -DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt5/5.4.2/ -DPYTHON_INCLUDE_DIR=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/include/python3.3m/ -DPYTHON_LIBRARY=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/libpython3.3.dylib -DPython_FRAMEWORKS=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/ ../ -D"CMAKE_BUILD_TYPE:STRING=Debug"
+```zsh
+cd [libopenshot repo folder]
+mkdir build
+cd build
+cmake -G "Unix Makefiles" -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DCMAKE_PREFIX_PATH="$(brew --cellar qt5)/<version>" -DPython_FRAMEWORKS="$(brew --cellar python3)/<version>/Frameworks/Python.framework/" ../ -D"CMAKE_BUILD_TYPE:STRING=Debug"
 ```
 
 The extra arguments on the cmake command make sure the compiler will be gcc4.8 and that cmake 
--- a/doc/INSTALL-MAC.md
+++ b/doc/INSTALL-MAC.md
@@ -155,10 +155,10 @@ few additional steps to manually build and install it. Launch a terminal and ent
 cd [libopenshot-audio repo folder]
 mkdir build
 cd build
-cmake -d -G "Unix Makefiles" -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang ../   (CLang must be used due to GNU incompatible Objective-C code in some of the Apple frameworks)
+cmake --debug-output -G "Unix Makefiles" -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang ../   (CLang must be used due to GNU incompatible Objective-C code in some of the Apple frameworks)
 make
 make install
-./src/openshot-audio-test-sound  (This should play a test sound)
+./src/openshot-audio-demo  (This should play a test sound)
 ```
 
 ## Mac Build Instructions (libopenshot)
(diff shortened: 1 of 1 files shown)


jonoomph:
FYI: We also have similar content in our wiki: https://github.com/OpenShot/libopenshot/wiki/Linux-Build-Instructions, that probably needs to be updated as well.

jonoomph:
Or maybe the Mac instructions: https://github.com/OpenShot/libopenshot/wiki/Mac-Build-Instructions

ferdnyc:
Thanks for submitting @wasimafser , as @jonoomph indicates there are so many places where this content is duplicated that it can be hard to keep track of all of them.

I actually don't understand why we're telling people to run `cmake` with **debug** output turned on at all, here. That's just useless noise. I'd much prefer to just take out that argument entirely, as there's no reason for it unless you're debugging a CMakeLists.txt file itself.

cc: @jonoomph

jonoomph:
@ferdnyc One idea I've seen, regarding GitHub wiki systems... have the wiki be part of our source code, for each repo, and use GitLab CI or GitHub automation to publish/sync the source markdown files to the GitHub wiki. That way, people could submit PRs to update the wiki, and it could also serve as local documentation with the source code. No more duplication between them. Just a thought....

doc/INSTALL-MAC.md
```diff
@@ -168,7 +168,7 @@ Run the following commands to build libopenshot:
 $ cd [libopenshot repo folder]
 $ mkdir build
 $ cd build
-$ cmake -G "Unix Makefiles"  -DCMAKE_CXX_COMPILER=/usr/local/opt/gcc48/bin/g++-4.8 -DCMAKE_C_COMPILER=/usr/local/opt/gcc48/bin/gcc-4.8 -DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt5/5.4.2/ -DPYTHON_INCLUDE_DIR=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/include/python3.3m/ -DPYTHON_LIBRARY=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/libpython3.3.dylib -DPython_FRAMEWORKS=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/ ../ -D"CMAKE_BUILD_TYPE:STRING=Debug"
+$ cmake --debug-output -G "Unix Makefiles"  -DCMAKE_CXX_COMPILER=/usr/local/opt/gcc48/bin/g++-4.8 -DCMAKE_C_COMPILER=/usr/local/opt/gcc48/bin/gcc-4.8 -DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt5/5.4.2/ -DPYTHON_INCLUDE_DIR=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/include/python3.3m/ -DPYTHON_LIBRARY=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/libpython3.3.dylib -DPython_FRAMEWORKS=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/ ../ -D"CMAKE_BUILD_TYPE:STRING=Debug"
```
ferdnyc, line 171: Yeah, this is actually all wrong. It's a very bad idea to just randomly switch between GCC and clang when compiling interoperable libraries, for starters.

```suggestion
$ cmake -G "Unix Makefiles"  -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DCMAKE_PREFIX_PATH="$(brew --cellar qt5)/<version>" -DPython_FRAMEWORKS="$(brew --cellar python3)/<version>/Frameworks/Python.framework/" ../ -D"CMAKE_BUILD_TYPE:STRING=Debug"
```

Unfortunately there doesn't seem to be a generic `brew` command to look up a _specific_ version's Cellar directory, but that's mostly because you're really not supposed to have to do that. 

I'd be surprised if Homebrew CMake can't find Homebrew Python _without_ needing any paths specified, or at most by adding something like the root `$(brew --cellar)` directory to `CMAKE_PREFIX_PATH`. But I don't have a Mac to test that on. I do know that they avoid installing Python into the default-accessible paths to avoid conflicts with system packages, which makes sense, but CMake can handle that if properly configured.

I'm **certain** there's a better way to go about this than the instructions we've been providing here, though, which were based on truly _ancient._ versions of CMake and don't take advantage of any of its more modern discovery features.

ferdnyc: Also, I'd suggest to remove the preceding `$ ` characters from all of these lines, again to facilitate copy-pasting.

ferdnyc:
Actually, I just remembered... I've already solved this, in the Github CI workflows for my ferdnyc/libopenshot#9 self-PR. Let's see, according to https://github.com/ferdnyc/libopenshot/blob/ci-artifacts/.github/workflows/ci.yml the correct set of commandline args to build libopenshot on MacOS using Homebrew is...
```sh
export PATH="/usr/local/opt/qt@5/bin:$PATH"
cd [libopenshot-repo-dir]
cmake -B build -S . -G "Unix Makefiles" \
-DCMAKE_INSTALL_PREFIX="build/dist" \
-DCMAKE_CXX_COMPILER=clang++ \
-DCMAKE_C_COMPILER=clang \
-DCMAKE_OSX_DEPLOYMENT_TARGET=10.9 \
-DQt5_DIR="/usr/local/opt/qt@5/lib/cmake/Qt5" \
-DCMAKE_EXE_LINKER_FLAGS="-stdlib=libc++" \
-DCMAKE_SHARED_LINKER_FLAGS="-stdlib=libc++"

# Then, for non-verbose builds
cmake --build build -j 4
# For verbose builds
cmake --build build -j 4 --verbose
```

Setting the `$PATH` makes sure the correct version of `qmake` is accessible, which is used to help discover the configuration.

Python3 is indeed found in `/usr/local/bin/python3` and `/usr/local/Frameworks/Python.framework/Versions/3.9/lib/libpython3.9.dylib` without requiring any special arguments at all.

On the official project Mac builders we also add this:
```sh
-DCMAKE_OSX_SYSROOT="/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk"
```
Not sure it's strictly required, but I suppose it can't hurt.

And we add this:
```sh
-DCMAKE_INSTALL_RPATH_USE_LINK_PATH=1
```
...which probably is important for getting relocatable builds. (Though that's at least partly necessary because we're still handling `INSTALL_NAME` wrong for the libraries, on MacOS. It might not be, if we can get `INSTALL_NAME` right.)

wasimafser:
> @ferdnyc One idea I've seen, regarding GitHub wiki systems... have the wiki be part of our source code, for each repo

This is definitely better , if you look at the docs and wiki now, they're a bit different, refactoring the docs to also serve as a wiki will be a good idea.

ferdnyc:
> This is definitely better , if you look at the docs and wiki now, they're a bit different, refactoring the docs to also serve as a wiki will be a good idea.
> 

Agreed, though in a sense it's a bit academic since **neither** set is accurate with respect to reality. As tends to be the nature of docs.

(There's also an _additional_ version of all this info in the OpenShot user manual, which currently is not kept in sync with any of the other versions.)

Updating docs in just one place would certainly be better than updating... hmm... at LEAST 5, by my count.

wasimafser:
> I actually don't understand why we're telling people to run `cmake` with **debug** output turned on at all, here.

The docs had the argument "-d" which is completely not understandable, I just assumed it would be "--debug-output".
Thanks for the information !

ferdnyc:
> The docs had the argument "-d" which is completely not understandable, I just assumed it would be "--debug-output".

And, honestly, that's as good an assumption as any from what was actually there. The argument was definitely nonsensical as given. I made the same assumption, on seeing it.

However, looking back at old-old documentation, it appears `-d` has only ever been a synonym for `-D`. And `-d`/`-D` apparently _can_ legally be followed by an optional space, before the variable definition. So it appears these are valid alternate forms of the same thing:
```sh
cmake -d CMAKE_CXX_COMPILER=clang++
cmake -DCMAKE_CXX_COMPILER=clang++
```

Given that, I suspect the rogue `-d` was simply a remnant of some earlier edit, which makes me even more inclined to just remove it. Whatever it was intended to mean by the author of that command line, it does not actually mean.

wasimafser:
Thank you for the information @ferdnyc, with that, shall I complete the changes in this PR? or just abandon this to work with the doc as wiki idea?

jonoomph:
Let's go ahead and merge these change in the meantime, since it is a valid fix

doc/INSTALL-MAC.md
```diff
@@ -155,10 +155,10 @@ few additional steps to manually build and install it. Launch a terminal and ent
 cd [libopenshot-audio repo folder]
 mkdir build
 cd build
-cmake -d -G "Unix Makefiles" -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang ../   (CLang must be used due to GNU incompatible Objective-C code in some of the Apple frameworks)
+cmake --debug-output -G "Unix Makefiles" -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang ../   (CLang must be used due to GNU incompatible Objective-C code in some of the Apple frameworks)
```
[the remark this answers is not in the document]
jonoomph: Yeah, I like this as well

jonoomph (commented):
I would suggest making the changes @ferdnyc mentioned, with respects to copy/paste friendly syntax... then let's merge this! Thanks again!

doc/INSTALL-MAC.md
```diff
@@ -151,24 +151,25 @@ brew install zeromq
...
 
 ## Mac Build Instructions (libopenshot)
 Run the following commands to build libopenshot:
 
-```
-$ cd [libopenshot repo folder]
-$ mkdir build
-$ cd build
-$ cmake -G "Unix Makefiles"  -DCMAKE_CXX_COMPILER=/usr/local/opt/gcc48/bin/g++-4.8 -DCMAKE_C_COMPILER=/usr/local/opt/gcc48/bin/gcc-4.8 -DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt5/5.4.2/ -DPYTHON_INCLUDE_DIR=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/include/python3.3m/ -DPYTHON_LIBRARY=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/libpython3.3.dylib -DPython_FRAMEWORKS=/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/ ../ -D"CMAKE_BUILD_TYPE:STRING=Debug"
+```zsh
```
jonoomph, line 168: Is this a formatting thing? I haven't seen that before... just checking
"```zsh"

wasimafser: Yes, instead of just having gray text we can format it to what we want. Mac uses zsh and fortunately GitHub markdown too.

Just try it out...
```python
for I in range(10):
  print("haha")
```

```cpp
for(int I=0; I<10;I++){
   cout<<"amazing";
}
```

jonoomph:
LG, merging

PR #681, opened 2021-06-02
Clip: Restore property choice names

ferdnyc:
2cf2c55f was a bit too ambitious, and swept away all of the property value choice labels that used to contain `"None"`.

cc: @BrennoCaldato


BrennoCaldato:
Thanks for reviewing it @ferdnyc. I think you are right. This can probably be causing the inconvenient of having to click twice to select a parent clip

ferdnyc:
@BrennoCaldato :

It shouldn't affect setting the parent clip, those strings I _didn't_ revert.

The ones reverted here were changes to the labels for certain multiple-choice properties that represent enum values — OpenShot displays the (translated) string labels, instead of the integer value. So, when e.g. `scale` is `3` (which is `openshot::ScaleType::SCALE_NONE`), we want it to display as `"None"`, or the local translation of `"None"`. It looked like the labels got caught up in a file-wide search&replace for `"None"`.

PR #683, opened 2021-06-05
Non-abstract effects

ferdnyc:
There were some issues with recent changes that caused SWIG to lose the definitions of all of our effect classes, making them abstract in the Python bindings. This PR brings them back to reality, and fixes the unit-test failures occurring in OpenShot due to the lack of `openshot.Mask()` in the bindings.


ferdnyc:
I meant to courtesy-CC the primary author for this code — cc: @BrennoCaldato

ferdnyc:
Merging to fix our OpenShot CI runs.

PR #684, opened 2021-06-06
CMake: Don't parallel-test on Windows, protect FFmpeg version checks

ferdnyc:
Somehow the old UnitTest++ version of tests/KeyFrame_Tests.cpp
got resurrected back into the repo. Shotgun blast to the head.

<s>(@jonoomph : I THINK this is the cause of the Win32 builder issues, oddly enough. We'll find out as soon as I blow it away.)</s>

**Update:** I was wrong. It appears that on Windows the tests just experience occasional hangs. I'm hoping it's just because the tests are being run in parallel. I've added logic to turn off the parallel testing on Windows platforms, we'll see if that clears it up. It "felt" to me like the hangs were the result of multiple test runs trampling all over each other. (Possibly some race condition as they contend over the same ZeroMQ port bindings, or the like.)

Also, it seems I missed something pretty important, when I recently added new version logic to CMake: The version variables for the FFmpeg components are _only_ set if FFmpeg is detected using PkgConfig. If it's _not_ (i.e., on Windows) those variables will exist but contain the empty string, something I hadn't properly protected against. This should fix it.


ferdnyc:
> Also, it seems I missed something pretty important, when I recently added new version logic to CMake: The version variables for the FFmpeg components are _only_ set if FFmpeg is detected using PkgConfig. If it's _not_ (i.e., on Windows) those variables will exist but contain the empty string, something I hadn't properly protected against.

The only reason our project builder Windows builds were still working at all, I think, is that we don't start from a completely fresh build directory each time. (Or do we? I don't _think_ so, or none of the Windows builds would have been able to get past the CMake configuration stage.) The existing build files were already set up, so the failure to detect FFmpeg component versions never came up.

ferdnyc:
Checks out on Linux, but I'm going to let my own local Windows test builds finish before merging.

ferdnyc:
> (@jonoomph : I THINK this is the cause of the Win32 builder issues, oddly enough. We'll find out as soon as I blow it away.)

I was wrong. It appears that on Windows they just experience occasional hangs. I'm _hoping_ it's just because the tests are being run in parallel. I've added logic to turn off the parallel testing on Windows platforms, we'll see if that clears it up.

PR #688, opened 2021-06-10
Revert final_cache size breakage to unbreak OpenShot caching

ferdnyc:
Reverting the "Adjusting Timeline final_cache to match the video caching thread max_frames, so one doesn't clobber the other." part of a previous commit.

Adjusting the cache size in `SetMaxSize()` ignores OpenShot's cache preferences. This prevents using a cache larger than the preview buffer, which is often a desirable thing. The initial size set for `final_cache` on initialization is merely a starting point, meant to be overridden by OpenShot based on preferences.

This partially reverts commit 0c4e1bcce4d8cbdc1edc752fe3339acf94ac8eb1.

My OpenShot PR OpenShot/openshot-qt#4075 (which I will now close) was based on confusion created by the fallout from this change. (I thought the Timeline cache was removed along with the Clip cache, since it appeared to be non-persistent due to the tiny cache forced by `SetMaxSize()`!)

### DIFF
--- a/src/Timeline.cpp
+++ b/src/Timeline.cpp
@@ -80,11 +80,12 @@ Timeline::Timeline(int width, int height, Fraction fps, int sample_rate, int cha
 	info.acodec = "openshot::timeline";
 	info.vcodec = "openshot::timeline";
 
-    // Init cache
-    final_cache = new CacheMemory();
-
 	// Init max image size
 	SetMaxSize(info.width, info.height);
+
+	// Init cache
+	final_cache = new CacheMemory();
+	final_cache->SetMaxBytesFromInfo(max_concurrent_frames * 4, info.width, info.height, info.sample_rate, info.channels);
 }
 
 // Delegating constructor that copies parameters from a provided ReaderInfo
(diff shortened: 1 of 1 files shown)


ferdnyc:
I haven't heard any response on this, so last call @jonoomph, @JacksonRG, @ anyone else... my recent testing in 2.6.1 has convinced me this is necessary. (I've been trying to sort out OpenShot/openshot-qt#4358, wherein OpenShot cannot run under Python 3.10 because they finally pulled the trigger on removing automatic float->int conversion in the language.) I'm inclined to merge this unless anyone has any reasons why I shouldn't.

Performance on my relatively-slow, relatively-underpowered system with 20GB of RAM is _abysmal_, because even though I have OpenShot configured to use 2.5GB of memory for caching, the changes this PR will revert effectively cause that setting to be ignored. Since that change was merged the cache is forced down to a comically small value and cannot be increased, because it will be re-shrunk whenever the preview is resized. That's just not workable in practice.

With a larger cache I can at least play through in stutter-mode once to fill it, and then have smooth playback when _replaying_ the cached section.

ferdnyc:
Well, I made every attempt to call attention to this, and it's undeniably necessary from my POV, so... merging. If it causes problems, please begin a discussion on how we can address them _without_ effectively disabling caching for the Timeline. OpenShot is too slow to operate in realtime-only mode.

(IMHO we should actually be looking into an idle pre-rendering feature that will automatically walk even farther ahead. That way, the painful first jerky-mode playthrough wouldn't be required just to have smooth performance on subsequent plays.)

PR #661, opened 2021-04-20
Make console-output functions redirectable; add operator<< for Fraction, Point, Coordinate; new KeyFrame::PrintPoints & PrintValues with unit tests

ferdnyc:
### Redirectable output functions

To enable unit-testing of functions like `ReaderBase::DisplayInfo()` and `Keyframe::PrintPoints()` (which normally dump information to `std::cout`), each such function has been updated to have a signature like this:
```c++
/// Display effect information in the standard output stream (stdout)
void DisplayInfo(std::ostream* out=&std::cout);
```

The function now takes a pointer to the output stream to use, with `std::cout` being only the overridable _default_ for that stream. In the output function, `*out` is streamed to instead of the hardcoded `std::cout`:
```c++
void WriterBase::DisplayInfo(std::ostream* out) {
	*out << std::fixed << std::setprecision(2) << std::boolalpha;
	*out << "----------------------------" << std::endl;
	*out << "----- File Information -----" << std::endl;
	*out << "----------------------------" << std::endl;
	*out << "--> Has Video: " << info.has_video << std::endl;
	*out << "--> Has Audio: " << info.has_audio << std::endl;
```

This makes it possible to unit test the functions by passing an alternate stream, then examining the captured output:
```c++
std::string expected(R"(----------------------------
----- File Information -----
----------------------------
--> Has Video: true
--> Has Audio: true
--> Has Single Image: false
--> Duration: 51.95 Seconds
--> File Size: 7.26 MB
----------------------------
----- Video Attributes -----
----------------------------
--> Width: 1280
--> Height: 720)");

// Store the DisplayInfo() text in 'output'
std::stringstream output;
r.DisplayInfo(&output);

// Compare a [0, expected.size()) substring of output to expected
auto compare_value = output.str().compare(0, expected.size(), expected);
CHECK(compare_value == 0);
```

Unit tests for `ReaderBase::DisplayInfo()` (via `FFmpegReader`), `WriterBase::DisplayInfo()` (via `FFmpegWriter`), and `FrameMapper::PrintMapping()` are included in the PR.

### Output-stream compatible formatting for Coordinate, Point, Fraction

Also, the `operator<<` overloads taking `openshot::Coordinate`, `openshot::Point`, and `openshot::Fraction` which I'd written for another PR that ended up on hold are included here, along with their unit tests.

### Updated `Keyframe::PrintPoints()` and `Keyframe::PrintValues()`

[...]

PR #689, opened 2021-06-11
FFmpeg: Create, use av_err2string()

ferdnyc:
(Note: Split out of #661)

My commit message for this reads:
- Previously 'av_make_error_string' was defined in FFmpegUtilities.h
  for the sole purpose of redefining `av_err2str()` as a call to that
  function. `av_err2str()` was then used in our code, often in string
  contexts where its output was cast to `std::string`.
- Since that was excessively circular, instead the function is named
  `av_err2string()`, and it's used directly in contexts where a
  std::string is expected.
- `av_err2str()` is still #defined as `av_err2string(...).c_str()`

What _actually_ happened was worse than I made it sound:

1.  `av_make_error_string()` wrapped FFmpeg's `av_strerror()` to return a `std::string`, for C++ compatibility
2. `av_err2str()` was then redefined:
    ```c++
    #define av_err2str(errnum) av_make_error_string(errnum).c_str()
    ```
    thereby producing the `char*` version of `av_make_error_string()`
3. <s>Most</s> All uses of `av_err2str(errnum)` in the code were then written as
   ```c++
   (std::string)av_err2str(errnum)
   ```

Like I said, dizzyingly circular. So all of those `(std::string)av_err2str(errnum)`s are now replaced with `av_err2string(errnum)`. No casting needed.


ferdnyc:
I have no idea why I haven't merged this yet, so... doing so.

PR #691, opened 2021-06-21
CMake: USE_SWIG_DEPENDENCIES for CMake 3.20+

ferdnyc:
**(F---ING _FINALLY!!!_)**

Mostly a developer convenience, when building using the newest CMake 3.20 or later, the SWIG targets will be configured with `USE_SWIG_DEPENDENCIES` set to `TRUE`, which causes the target to automatically set up dependencies on all of the files referenced by the SWIG bindings.

Which means, when you make changes to `Frame.h` or `Clip.h` or whatever, the bindings will be automatically regenerated, and you no longer have to `rm -r build/bindings` in order to force a re-sync with the header files.

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -95,6 +95,12 @@ if (DEFINED _inc)
   set_property(SOURCE openshot.i PROPERTY INCLUDE_DIRECTORIES ${_inc})
 endif()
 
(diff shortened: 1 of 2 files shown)


ferdnyc:
Merging, as this is a Good Thing™ for the very few developers it will currently affect (CMake 3.20 being quite new, at the moment).

PR #693, opened 2021-06-25
FFmpeg: Move av/swresample decision into CMake

ferdnyc:
:heavy_check_mark:  <s> **Note: PR #694 MUST be merged before this PR.** </s>

I discovered with @BrennoCaldato 's help that having CMake decide which resampling library to link with, but having the code in the `FFmpegUtilities.h` header decide which one to _use_, wasn't really working out well in all cases. In some FFmpeg versions, CMake would link with libswresample, but the header would select the avresample API, resulting in a broken libopenshot build that didn't have the right set of library dependencies configured.

This PR moves the determination as to which resampling library to use into the `src/CMakeLists.txt` code, which will set flags to _tell_ `FFmpegUtilities.h` which one has been linked in (and therefore, which one to use). This ensures no broken builds due to mismatching configs.

Plus the choice can be displayed in the FeatureSummary at configure time:

<pre><code>
$ cmake -B build -S . -DOpenShotAudio_ROOT=../libopenshot-audio/build
[...]
-- Checking for module 'libavcodec'
--   Found libavcodec, version 58.91.100
-- Checking for module 'libavformat'
--   Found libavformat, version 58.45.100
-- Checking for module 'libavutil'
--   Found libavutil, version 56.51.100
-- Checking for module 'libswscale'
--   Found libswscale, version 5.7.100
-- Checking for module 'libswresample'
--   Found libswresample, version 3.7.100
-- Checking for module 'libavresample'
--   Found libavresample, version 4.0.0
-- Found FFmpeg: /usr/lib64/libavcodec.so;/usr/lib64/libavformat.so;/usr/lib64/libavutil.so;
/usr/lib64/libswscale.so;/usr/lib64/libswresample.so;/usr/lib64/libavresample.so 
(found version "4.3.2") found components: avcodec avformat avutil swscale swresample avresample
[...]
-- Build configuration:
-- The following features have been enabled:

<b> * FFmpeg swresample, Audio resampling uses swresample</b>
 * FFmpeg hwaccel, GPU-accelerated routines (FFmpeg 3.4+)
 * OpenCV algorithms, Use OpenCV algorithms
 * Parallel tests, Unit tests can use 4 processors
 * Unit tests, Compile unit tests for library functions
 * Non-default target 'coverage', Run unit tests and (if enabled) collect coverage data
 * Non-default target 'doc', Build formatted API documentation (HTML+SVG)
[...]
</code></pre>

PR #694, opened 2021-06-25
Enhance FindFFmpeg.cmake with fallback version parsing, FFmpeg_VERSION, etc.

ferdnyc:
My recent PR #693 depends on the availability of the `avformat_VERSION` value in CMake, just as a previous PR added a dependency on the `avcodec_VERSION` for determining hardware-acceleration support. Previously, FFmpeg component versions were only available if PkgConfig had been used to retrieve the FFmpeg configuration.

This PR enhances the Find module so that, if PkgConfig is not available/used, the component versions are parsed out of the relevant `version.h` file in each component's include directory.

Along the way, it also makes several other enhancements to the Find module:

- Probing of non-requested FFmpeg components (a serious time-waster) is eliminated, so if the `find_package()` line specifies `COMPONENTS` or `OPTIONAL_COMPONENTS` only those libraries will be available.
- All component variables now use the CMake standard naming of `FFmpeg_<component>_<var>`, instead of the previous `<component>_<var>`.
- `find_package_handle_standard_args()` is  now called with `HANDLE_COMPONENTS` to verify the component list against `FFmpeg_FIND_COMPONENTS`
- The overall `FFmpeg_VERSION` value is now parsed from `libavutil/ffversion.h` and published as the module's `VERSION_VAR`


ferdnyc:
The version number changes particularly affect the WIN32 platform, which previously lacked FFmpeg component versions because PkgConfig was not used to discover the library.

ferdnyc:
> * Probing of non-requested FFmpeg components (a serious time-waster) is eliminated, so if the `find_package()` line specifies `COMPONENTS` or `OPTIONAL_COMPONENTS` only those libraries will be available.

(Except for `avutil`, which will always be resolved due to the aforementioned parsing of `libavutil/ffversion.h`.)

Issue #676, opened 2021-05-24
FFmpegReader test fails, the output file has no audio


steils:
I have no sound when I export a video using openshot itself. I guess it's related to this error:

[openshot-FFmpegWriter-test.log](https://github.com/OpenShot/libopenshot/files/6530137/openshot-FFmpegWriter-test.log)
```
/home/stefan/projects/libopenshot/tests/FFmpegWriter.cpp:92
...............................................................................

/home/stefan/projects/libopenshot/tests/FFmpegWriter.cpp:121: FAILED:
  CHECK( r1.info.has_audio == true )
with expansion:
  false == true

/home/stefan/projects/libopenshot/tests/FFmpegWriter.cpp:124: FAILED:
  CHECK( r1.GetFrame(1)->GetAudioChannelsCount() == 2 )
with expansion:
  0 == 2

/home/stefan/projects/libopenshot/tests/FFmpegWriter.cpp:125: FAILED:
  CHECK( r1.info.channel_layout == LAYOUT_STEREO )
with expansion:
  4 == 3

===============================================================================
test cases:  2 |  1 passed | 1 failed
assertions: 15 | 12 passed | 3 failed
```

I have tracked down the issue to this change: https://github.com/OpenShot/libopenshot/commit/1dba02f5b01fb0aefe30a2286f0507839ad0ecc9

If I revert it (or the entire PR #653, it has to resolve some conflicts) the test passes.

steils:
Yes, I'm using the latest snapshots:
libopenshot-audio: 92ff6084d3a773ae2866d786c3107ff6973ceee1
libopenshot: b09416c2e3979246c788abfcf632dd36ab7da046

jonoomph:
Yes, this is an issue with FFmpeg 4, and some upstream changes to FFmpeg that have broken audio encoding in libopenshot. Can you confirm your version of FFmpeg you are using?

steils:
Yes, 4.4.

Apteryks:
Perhaps useful information for a fix: https://devtalk.blender.org/t/newest-ffmpeg-v4-4-is-having-issues-with-blender-2-92-animation-audio-is-not-rendered-with-video/18663/3

jonoomph:
[...]

PR #698, opened 2021-07-07
Fix audio encoding on FFmpeg 4+

jonoomph:
AVFrame was missing a few properties that need to be set in newer versions of FFmpeg. This is also compatible with FFmpeg 3+, based on my testing.

Closes: https://github.com/OpenShot/libopenshot/issues/676


JacksonRG:
Just tested, mono/stereo both audio and audio with video. No issues!

LGTM

JacksonRG (approved):

Issue #702, opened 2021-07-13
Missing transition files cause openshot::InvalidFile to be thrown, crash OpenShot


ferdnyc:
While testing OpenShot/openshot-qt#4229, I discovered that if a project file is loaded that contains any invalid transition paths, libopenshot crashes OpenShot entirely.

This appears to be due to the relatively-new project file parsing code, as even fixing the paths in OpenShot's in-memory project `dict` doesn't prevent the crash (implying that the original data is still being read from disk).

Needless to say, a missing file shouldn't cause the application to completely crash.

I also notice there are no unit tests _at all_ for the `.osp` file loading, which is disturbing.

ferdnyc:
I'm becoming more and more convinced, in light of this issue, that the transition files should actually live with _libopenshot_, not OpenShot itself. If libopenshot needs the transition files to operate properly, then it should own them, it shouldn't need to be pointed at an OpenShot installation just to not-crash when loading project data. That's completely ass-backwards.

The effects files, too, should really be part of libopenshot, since the effects are completely defined in its code. The fact that we have descriptions of, and icons for, them that only exist in an external repo is kind of crazy. OpenShot should be pulling those resources from the library's install, not the other way around.

JacksonRG:
Ohh, I've been investigating a few crashes I've been having. This could explain some of those

ferdnyc:
@JacksonRG I wouldn't be at all surprised. libopenshot's incredible fragility in the face of missing media files is a whole 'nother issue in its own right, we really need to address that.

_This_ issue certainly **exacerbates** it (or is exacerbated by it?), and the tetchiness it's displayed of late makes it all the more imperative that the library itself take ownership of all the files it requires. But, really it should just **never** _crash hard_, period. Not for something as silly as, "can't find an input file mentioned in the project JSON".

PR #704, opened 2021-07-13
Exceptions: Add py_message() for Python

ferdnyc:
- Most exceptions are now based on one of two new base classes,  FileExceptionBase and FrameExceptionBase. Both inherit from ExceptionBase.
- **All** exception classes derived from ExceptionBase now export a `py_message()` method
  (inherited or defined) which constructs a message for Python that includes **all** arguments to the exception, not just the `why()` text.
- Python's exception handling is updated to catch `openshot::ExceptionBase` and display `py_message()` instead of `why()`
- (I didn't modify `why()` because it returns `char*`; `py_message()` returns a `std::string`. `why()` is a `const` method not conducive to constructing and returning a temporary `char*`.)

### DIFF
--- a/bindings/python/openshot.i
+++ b/bindings/python/openshot.i
@@ -131,6 +131,9 @@
 	try {
 		$action
 	}
+	catch (openshot::ExceptionBase &e) {
+		SWIG_exception_fail(SWIG_RuntimeError, e.py_message().c_str());
+	}
 	catch (std::exception &e) {
 		SWIG_exception_fail(SWIG_RuntimeError, e.what());
 	}
--- a/src/Exceptions.h
+++ b/src/Exceptions.h
@@ -32,6 +32,7 @@
 #define OPENSHOT_EXCEPTIONS_H
 
 #include <string>
+#include <cstring>
 
 namespace openshot {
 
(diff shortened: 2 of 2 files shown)


ferdnyc:
(The reason for this, btw, is that when debugging problems like #703, it's incredibly frustrating to see OpenShot crashing with a libopenshot exception, and you KNOW there's more data available that it's just not showing you.)

ferdnyc:
I meant #702.

ferdnyc:
(I wonder if Sentry has a more-robust C++ crash reporter, that could replace our ad-hoc one? That could potentially open up even bigger cans of worms, re: privacy and whatnot. ...But even if we didn't use it in production builds, just for our own developer use it could come in handy.)

jonoomph:
@ferdnyc Yes, Sentry comes with support for mini-dumps: https://docs.sentry.io/platforms/native/guides/minidumps/, including both Crashpad and Breakpad. And from what I understand, we have X seconds to append additional info to a reported crash in Sentry.io. So, I "think" if we could package one of these mini-dump handlers, we could generate a dump file, and then POST it to Sentry to be appended to our crash report.

Also, I think we only need to do this for Linux (for simplicity)... since most crashes happen in a cross-platform way, and that would probably help us track down the majority of our libopenshot C++ crashes.

jonoomph:
LGTM though

PR #705, opened 2021-07-16
Various small code cleanup fixes

ferdnyc:
Compile libopenshot with warnings set high enough, and you get a LOT of messages about issues in our code. Most prevalent are unused-variable and unused-parameter errors.

This PR cleans up a number (but not all) of those warnings.

### Unused parameters

In particular, where there were unused parameter warnings, it attempts to follow the [C++ Core Guidelines' advice](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-unused):

> #### F.9: Unused parameters should be unnamed
> Reason Readability. Suppression of unused parameter warnings.Example
> ```
> X* find(map<Blob>& m, const string& s, Hint);   // once upon a time, a hint was used
> ```
> Note Allowing parameters to be unnamed was introduced in the early 1980 to address this problem.

Where the name was needed as part of a signature (such as in Qt code), an artificial use was placed at the top of the function. Either `Q_UNUSED(name)`, or `(void)name;  // UNUSED`

### `switch()` `case`s: fallthrough and `#ifdef`

There was some apparently-intentional use of switch case fallthrough, done in concert with `#ifdef` wrapping to restrict code to certain FFmpeg versions.

Fallthrough is too magical at the best of times. (In fact it'll cause compiler warnings, with `-Wall -Wextra`. Unless it's in the form of simple stacked `case:`s, which are considered OK).

`#ifdef` inside or around a switch (effectively creating mutable switch statements) is pretty bad, too. Combining the two is a recipe for disaster.

So, those sections of code are replaced with equivalent simple `if()`s, wrapped in the same `#ifdefs`. One switch still uses an `#ifdef` with fallthrough, but only in the simple stacked-case form:

```c++
    switch (condition) {
#ifdef
    case MAYBETHING :
#endif
    case THING2 :
    case THING3 :
    case THING4 :
        // stuff here...
        break;
```

...Which I think is clear enough at face value to be OK.

### Timeline inheritance

**ClipBase and TimelineBase derive from ReaderBase**

- This fixes our Ruby bindings buy eliminating the unsupported multiple inheritance in Clip and Timeline

- As a side-effect, all Effects are now Readers as well (EffectBase inherits from ClipBase); the ReaderBase methods are stubbed in EffectBase so that the individual effects don't have to implement them.

#### CI builds

In addition to patching up many of these issues, the PR also sets our GitHub CI build to run with warnings enabled (except for a couple which are too widespread to be included). However, it removes the `VERBOSE` flag from the build, because honestly seeing the compile commands used to generate the source is of little value, as compared to focusing on any warnings or errors generated.


src/CVObjectDetection.cpp
```diff
@@ -318,15 +318,15 @@ bool CVObjectDetection::SaveObjDetectedData(){
...
+    for(size_t i = 0; i<classNames.size(); i++){
         std::string* className = objMessage.add_classnames();
         className->assign(classNames.at(i));
     }
 
     // Iterate over all frames data and save in protobuf message
     for(std::map<size_t,CVDetectionData>::iterator it=detectionsData.begin(); it!=detectionsData.end(); ++it){
         CVDetectionData dData = it->second;
-        pb_objdetect::Frame* pbFrameData;
+        // pb_objdetect::Frame* pbFrameData;
```
jonoomph, line 329: Is this pointer used?

jonoomph:
@ferdnyc Otherwise looks good... lots of changes... and a bit hard to review. I made a few notes about commented out code snippets (on purpose or accident?)

src/CVObjectDetection.cpp
ferdnyc: AFAICT no, I think I commented it out rather than deleting it only to act as a placeholder/reminder to @BrennoCaldato (or anyone) that a `pbFrameData` **was** being allocated here, presumably for some reason (but then nothing was ever done with it).

Probably would've made sense to leave an actual **comment**, along with the commented-out code.

ferdnyc:
@jonoomph TBH, enough of this is covered (or done differently) in other, already-merged PRs (as the conflicting files list makes very obvious) that I'm going to close this one... I _may_ salvage some of the commits into a new PR, based off the current `develop`, but this one's just diverged too far to tease apart.

PR #712, opened 2021-08-11
Color: Add Color(QColor) constructor with unit test, move initialization into initializer lists

ferdnyc:
This PR adds a new constructor overload for openshot::Color, which accepts a QColor (by const reference) as its argument, and initializes the Color keyframes to the corresponding `.red()`, `.green()`, `.blue()`, and `.alpha()` values. A new unit test is added for this constructor.

The PR also changes the other Color constructors to perform member variable initialization in the constructor's initializer list, rather than in the internal code. (Something the new constructor made possible, as `openshot::Color::Color("#rrggbb")` can now delegate to `openshot::Color::Color(QString(hex_string))` which will call the `QColor` overload.)

This required a small modification to `KeyFrame.cpp`, because for some reason the `Point(double)` constructor defaults to `CONSTANT` interpolation instead of `BEZIER`. So, changing the Color initialization from:
```c++
Color::Color(unsigned char Red, unsigned char Green, unsigned char Blue, unsigne
d char Alpha)
{
    // Set initial points
    red.AddPoint(1, (float)Red);
    green.AddPoint(1, (float)Green);
    blue.AddPoint(1, (float)Blue);
    alpha.AddPoint(1, (float)Alpha);
}
```
to
```c++
Color::Color(unsigned char Red, unsigned char Green, unsigned char Blue, unsigned char Alpha) :
    red(static_cast<double>(Red)),
    green(static_cast<double>(Green)),
    blue(static_cast<double>(Blue)),
    alpha(static_cast<double>(Alpha)) { }
```
Initially changed the default interpolation of the keyframes created.

By modifying the `openshot::Keyframe::Keyframe(double)` constructor to call `AddPoint(Point(1, value))` insead of `AddPoint(Point(value))`, the semantics were restored and keyframes now always default to `BEZIER` interpolation even with only a single point.


jonoomph:
LGTM!

PR #713, opened 2021-08-11
SCALE_NONE was incorrectly sizing SVG, Image, and Video files.

jonoomph:
Since OpenShot changes the preview window size during editing, the SCALE_NONE clips should always be scaled to the ratio: `preview width / timeline width`... so they stay relative sized to the timeline size.

For example, if an SVG is 500px wide, and the project is 1000px wide... the SVG in SCALE_NONE mode should always be 50% with width of the video, regardless of how small you make the preview window.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -33,6 +33,7 @@
 
 #include "FFmpegReader.h"
 #include "Exceptions.h"
+#include "Timeline.h"
 
 #include <thread>    // for std::this_thread::sleep_for
 #include <chrono>    // for std::chrono::milliseconds
@@ -1285,9 +1286,18 @@ void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
 			}
 
 		} else {
-			// No scaling, use original image size (slower)
-			max_width = info.width;
-			max_height = info.height;
+            // Scale video to equivalent unscaled size
+            // Since the preview window can change sizes, we want to always
+            // scale against the ratio of original video size to timeline size
+            float preview_ratio = 1.0;
+            if (parent->ParentTimeline()) {
+                Timeline *t = (Timeline *) parent->ParentTimeline();
+                preview_ratio = t->preview_width / float(t->info.width);
+            }
+            float max_scale_x = parent->scale_x.GetMaxPoint().co.Y;
+            float max_scale_y = parent->scale_y.GetMaxPoint().co.Y;
+            max_width = info.width * max_scale_x * preview_ratio;
+            max_height = info.height * max_scale_y * preview_ratio;
 		}
 	}
 
--- a/src/QtImageReader.cpp
+++ b/src/QtImageReader.cpp
@@ -68,10 +68,14 @@ void QtImageReader::Open()
 	if (!is_open)
 	{
 		bool loaded = false;
+        QSize default_svg_size;
 
         // Check for SVG files and rasterizing them to QImages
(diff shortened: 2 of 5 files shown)


jonoomph:
Looks like I have some unit test failures... oops

src/FFmpegReader.cpp
```diff
@@ -1285,9 +1286,18 @@ void FFmpegReader::ProcessVideoPacket(int64_t requested_frame) {
...
-			// No scaling, use original image size (slower)
-			max_width = info.width;
-			max_height = info.height;
+            // Scale video to equivalent unscaled size
+            // Since the preview window can change sizes, we want to always
+            // scale against the ratio of original video size to timeline size
+            float preview_ratio = 1.0;
+            if (parent->ParentTimeline()) {
+                Timeline *t = (Timeline *) parent->ParentTimeline();
+                preview_ratio = t->preview_width / float(t->info.width);
```
JacksonRG, line 1295: Probably missunderstanding this. But I think preview_width is the preview window width, and info.width is the SVG's width.
In that case, the example of a 500px wide image on a 1000px wide preview would have preview_ratio of 2.

src/QtImageReader.cpp
```diff
@@ -100,8 +104,15 @@ void QtImageReader::Open()
...
-		info.height = image->height();
+		if (!default_svg_size.isEmpty()) {
+		    // Use default SVG size (if detected)
+            info.width = default_svg_size.width();
+            info.height = default_svg_size.height();
+		} else {
+		    // Use Qt Image size as a fallback
+            info.width = image->width();
+            info.height = image->height();
+		}
```
JacksonRG, line 115: So we use the svg's natural size if we can for scaling later. This much I get :+1:

src/QtImageReader.cpp
```diff
@@ -269,22 +297,28 @@ bool QtImageReader::load_svg_path(QString) {
...
-                image = std::make_shared<QImage>(image->scaled(
-                        svg_size.width(), svg_size.height(), Qt::KeepAspectRatio, Qt::SmoothTransformation));
+        if (loaded) {
+            // Set default SVG size
+            default_size.setWidth(image->width());
+            default_size.setHeight(image->height());
+
+            if (image->width() < current_max_size.width() || image->height() < current_max_size.height()) {
+                // Load SVG into larger/project size (so image is not blurry)
+                QSize svg_size = image->size().scaled(current_max_size.width(), current_max_size.height(), Qt::KeepAspectRatio);
```
JacksonRG, line 307: If I understand this, It's rendering the SVG at the maximum resolution, so we can scale down to the preview size... Let me know if I'm way off.

jonoomph: Yes, all of this logic is based on the concept that our preview window might be "smaller" than the project, and thus, we can cheat and use scaled down assets (videos, images, SVGs), so things go smoother during real-time preview.

src/QtImageReader.cpp
jonoomph: Exactly, and this fallback code is for non-SVG images... which still use the QtImageReader (such as PNG or JPEG)

src/FFmpegReader.cpp
jonoomph: t->info.width  is  timeline->info.width

jonoomph: So, it's `preview window width / timeline width`, which for a 1920px timeline width, and a 800px preview window width, it would be `800 / 1920 = 0.416666667` preview ratio

PR #719, opened 2021-08-24
New audio device [type] switching + Audio sync improvements

ferdnyc:
**NOTE:** This is a breaking change, and will require synchronized changes to OpenShot/openshot-qt (in `src/windows/preferences.py`) for OpenShot to continue to work correctly.

So, most of this actually started as old commits I'd done a while ago, forgot about, and never submitted.

This PR adds a new `Settings::Instance()->PLAYBACK_AUDIO_DEVICE_TYPE` string, initially empty, that can be set to target a specific JUCE `AudioIODeviceType` when there are multiple types available on the system.

If the type is not set when the `AudioDeviceManagerSingleton` is initialized, it will attempt to search the available types for a type which contains the device `Settings::Instance()->PLAYBACK_AUDIO_DEVICE_NAME`, and it will switch to that type before calling `initialise()`. But, ideally the type name should be stored alongside the device name.

The mechanism by which audio device names are retrieved is also modified, with a new `AudioDevices` class replacing the `AudioDeviceInfo` struct. `AudioDevices::getNames()` returns an STL container type (`std::vector<std::pair<std::string, std::string>>`) instead of using a custom struct. STL containers offer native Python container semantics, including iterability. For example, instead of:
```python3
value_list = []
# Loop through audio devices
value_list.append({"name": "Default", "value": ""})
for audio_device in get_app().window.preview_thread.player.GetAudioDeviceNames():
    value_list.append({
         "name": "%s: %s" % (
             audio_device.type, audio_device.name),
         "value": audio_device.name,
   })
```
an equivalent new-format list can be built with just:
```python3
devices = openshot.AudioDevices()
value_list = [{"name": "Default", "value": ("", "")})
value_list.extend([
    {"name": f"{t}: {n}", "value": (n, t)}
    for (n, t) in devices.getNames()
]
```

Addresses OpenShot/libopenshot-audio#133

**Update:**

- I've merged `develop` back into this PR (and resolved a bunch of conflicts)
- I've improved the audio sync/de-sync logic on our preview playback (to handle different cases)
- I've switched the resolution of our preview sleep to use microseconds (instead of milliseconds), for added precision. For example, on a 24 FPS video we need to sleep for 41.66 milliseconds, but we were only sleeping for 41. It was enough to cause drift issues very rapidly... which usually results in video stuttering.
- I've fixed an issue with a stale frame duration (which is a huge issue when changing profiles, but still sleeping based on a previous frame duration, causing runaway drift)


jonoomph:
@ferdnyc This one has some merge conflicts it appears

ferdnyc:
@jonoomph Yeah, it'll be affected by the JUCE 6 changes anyway so it's not the biggest priority right now. Not really worth merging ahead of the upgrade.

jonoomph:
@ferdnyc Just following up with you on the JUCE 6 support? Please let me know if I can help with that one. Thx!

jonoomph:
@ferdnyc I resolved the conflicts so I could test this out a bit. It actually fixed the Windows audio device issue I was having on my test machine, where no audio would play at all. However, I am seeing ~1 second delay between clicking play and hearing any sound. It's like the entire playback buffer is 1 second behind. Even seeking around the timeline... always 1 second behind any click or seek. 

I've been looking in the JUCE forums, to see if I can find any answers... but none yet. We are close!

jonoomph:
I'm also seeing some random slowdowns when seeking or clicking play... so this might not be directly related to JUCE. This leads me to believe we are sleeping too long in one of our playback threads, or mis-managing the syncing of our audio playback thread and our video playback thread... I'm going to add in some logging for Windows, and see what I can find.

jonoomph:
I've done quite a bit of real-life testing with this version, and in general, all audio device issues seem resolved on Windows + I'm seeing much improved playback performance, especially when changing profiles.

Issue #686, opened 2021-06-07
Unit test failure with OpenCV 4.5.2 due to failed assertion


ferdnyc:
@BrennoCaldato 

I'm seeing failures in the `CVTracker` unit test case `SaveLoad_Protobuf` when libopenshot is built using OpenCV 4.5.2 — exact same error, a failed assertion inside the OpenCV `trackerKCF.cpp` code itself, has turned up on both Windows MSYS2 and Fedora 34. Here's the output of the test runner:

```sh
% build/tests/openshot-CVTracker-test
[ INFO:0] global 
/builddir/build/BUILD/opencv-4.5.2/modules/core/src/parallel/registry_parallel.impl.hpp (96)
 ParallelBackendRegistry core(parallel): Enabled backends(2, sorted by priority):
 TBB(1000); OPENMP(990)
[ INFO:0] global 
/builddir/build/BUILD/opencv-4.5.2/modules/core/
include/opencv2/core/parallel/backend/parallel_for.tbb.hpp (54)
 ParallelForBackend Initializing TBB parallel backend: TBB_INTERFACE_VERSION=11103
[ INFO:0] global /builddir/build/BUILD/opencv-4.5.2/modules/core/src/parallel/parallel.cpp (77)
 createParallelForAPI core(parallel): using backend: TBB (priority=1000)
[ INFO:0] global /builddir/build/BUILD/opencv-4.5.2/modules/core/src/ocl.cpp (1176)
 haveOpenCL Initialize OpenCL runtime...
[ INFO:0] global /builddir/build/BUILD/opencv-4.5.2/modules/core/src/ocl.cpp (1182)
 haveOpenCL OpenCL: found 2 platforms
[ INFO:0] global /builddir/build/BUILD/opencv-4.5.2/modules/core/src/ocl.cpp (974)
 getInitializedExecutionContext OpenCL: initializing thread execution context
[ INFO:0] global /builddir/build/BUILD/opencv-4.5.2/modules/core/src/ocl.cpp (984)
 getInitializedExecutionContext OpenCL: creating new execution context...
[ INFO:0] global /builddir/build/BUILD/opencv-4.5.2/modules/core/src/ocl.cpp (1012)
 getInitializedExecutionContext OpenCL: context is not available/disabled

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #726, opened 2021-09-06
Fix CVTracker tests

ferdnyc:
This PR fixes the "SaveLoad_Protobuf" unit test so that it passes with OpenCV 4.5.2+ (where previously it was asserting due to an empty ROI, see linked issue.

Fixes: #686 

cc: @BrennoCaldato

### DIFF
--- a/tests/CVTracker.cpp
+++ b/tests/CVTracker.cpp
@@ -40,8 +40,46 @@
 
 using namespace openshot;
 
-// Just for the tracker constructor, it won't be used
-ProcessingController tracker_pc;
+TEST_CASE( "initialization", "[libopenshot][opencv][tracker]" )
+{
+    std::string bad_json = R"proto(
+    }
+        [1, 2, 3, "a"]
+    } )proto";
+    ProcessingController badPC;
+    CVTracker* badTracker;
+    CHECK_THROWS_AS(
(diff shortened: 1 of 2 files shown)


tests/CVTracker.cpp
```diff
@@ -73,10 +73,10 @@ TEST_CASE( "Track_Video", "[libopenshot][opencv][tracker]" )
...
 
     // Compare if tracked data is equal to pre-tested ones
-    CHECK(x >= 255); CHECK(x <= 257);
-    CHECK(y >= 133); CHECK(y <= 135);
-    CHECK(width >= 179); CHECK(width <= 181);
-    CHECK(height >= 165); CHECK(height <= 168);
+    CHECK(x == Approx(256).margin(1));
+    CHECK(y == Approx(134).margin(1));
+    CHECK(width == Approx(180).margin(1));
+    CHECK(height == Approx(166).margin(2));
```
ferdnyc, line 79: I wrote these ranges to approximate what the previous code was doing, but they really feel kind of broad. It seems like we should be able to be more precise than that, is there really that much variation in the results that we'll get?

ferdnyc:
Merging, to fix unit tests for anyone building with OpenCV 4.5.2+.

PR #727, opened 2021-09-08
CI: Build audio lib & cache

ferdnyc:
This is an experimental attempt to replace use of the libopenshot-audio PPA with a cached build done during CI. The cache is keyed on the OS, compiler, and a hash of the OpenShotAudio `CMakeLists.txt`; iff all of those are the same, it will skip building the audio lib.

This will allow us to add macOS and Windows CI builds for libopenshot.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,10 +13,16 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-      # Work around a codecov issue detecting commit SHAs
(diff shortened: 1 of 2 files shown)


ferdnyc:
I'm going to merge this, as I'd like to see it work under a few different scenarios before creating any Mac/Win CI configs. In theory it should be no different from depending on the PPA, which is always built against the `develop` branch of OpenShot/libopenshot-audio. I guess the only difference will be, if we manage to merge **broken** code so that `develop` doesn't build, the PPA would stay at the most recent known-good config, whereas here the broken code will be checked out and built (or attempted). But that should be pretty rare anyway, with CI protecting the repos from broken commits.

PR #729, opened 2021-09-11
Fold libopenshot_protobuf into libopenshot

ferdnyc:
This PR eliminates the building of a separate `libopenshot_protobuf.{so,dll,dylib}`, an additional complication in the build/install process (particularly on Windows, due to DLL path issues) that isn't particularly necessary. If OpenCV is enabled, the generated protocol buffers source files and headers are added to the main `openshot` library target as `PRIVATE` sources.

- The glob for `.proto` files is eliminated, all existing `.proto` files (moved from `src/protobuf_messages/` to `src/`) are explicitly listed in `src/CMakeLists.txt`. Any new `.proto` files created should be added to the `PROTOBUF_SOURCES` list in that file.
- The generated `*.pb.h` headers (listed in the CMake variable `${ProtoHeaders}` as an output of the `PROTOBUF_GENERATE_CPP()` macro) are no longer referenced in any OpenShot header files. Any existing references have been replaced with forward declarations of the appropriate classes.
- As such, those headers are now considered private implementation details, and they are no longer written to `${CMAKE_INSTALL_INCLUDEDIR}/libopenshot/` during install. Outside libopenshot API callers have no access to the internals of the protobuf-generated classes, nor should they need it.

cc: @BrennoCaldato


ferdnyc:
I'm gonna go for it, with this one. It just makes life easier.

PR #716, opened 2021-08-21
Add support for more Chroma Key methods

troyrollo:
Adds support for new, selectable, Chroma Key keying methods:

- HSV hue,
- HSV saturation,
- HSV luminosity;
- CIE LCH(ab) luminosity;
- CIE LCH(ab) chroma;
- CIE LCH(ab) hue;
- CIE colour difference based on CIEDE2000

This feature requires the babl library for colour model conversion.

PR #730, opened 2021-09-12
ChromaKey: Minimal unit testing

ferdnyc:
In #716 @troyrollo provides some interesting new keying modes for the ChromaKey effect.  I mentioned that unit testing for effects is currently lacking (abysmal might be a better word), and gave some thoughts on improving that.

This PR adds `tests/ChromaKey.cpp` which demonstrates how those tests could be structured, implementing very, very basic function tests for `ChromaKey::GetFrame()`. It can hopefully also serve as a template for anyone looking to implement tests for other effects.


jonoomph:
@ferdnyc Thanks, this looks great!

PR #731, opened 2021-09-16
Add explicit QtSvg dependency

ferdnyc:
A QtImageReader unit test failure building Fedora packages made me realize that we have a silent dependency on QtSvg. Normally it's pulled in by QtGui, so it's not an issue, but that assumes it's been _installed_ at all — which isn't a good assumption, when packaging for Linux distros. So, better to make that dependency explicit.


ferdnyc:
Ugh, Debian packaging is _BYZANTINE_. "libqt5svg5-dev", SRSLY?!?

PR #732, opened 2021-09-17
Rewrite Crop effect using QPainter

ferdnyc:
The crop effect was using some frankly byzantine pointer math and `memcpy()` calls to implement what a couple of `QRect`s and `QPainter::drawImage()` can make child's play. So, this PR rewrites it to use exactly those tools. The operations performed are the same, just without the bit-banging.

(Actually, the algorithm here is slightly improved. With the previous implementation, setting both the left and right crop sizes to `0.5` left a thin sliver of visible image in the output. Ditto top and bottom. Here, that's eliminated, and left `0.5` + right `0.5` == crop out the entire frame.)

### DIFF
--- a/src/AudioReaderSource.cpp
+++ b/src/AudioReaderSource.cpp
@@ -30,6 +30,10 @@
 
 #include "AudioReaderSource.h"
 #include "Exceptions.h"
+#include "Frame.h"
+#include "ZmqLogger.h"
+
+#include <OpenShotAudio.h>
 
 using namespace std;
 using namespace openshot;
--- a/src/AudioReaderSource.h
+++ b/src/AudioReaderSource.h
@@ -31,8 +31,8 @@
 #ifndef OPENSHOT_AUDIOREADERSOURCE_H
 #define OPENSHOT_AUDIOREADERSOURCE_H
 
-#include <iomanip>
 #include "ReaderBase.h"
+
 #include <OpenShotAudio.h>
 
 /// This namespace is the default namespace for all code in the openshot library
--- a/src/CacheBase.h
+++ b/src/CacheBase.h
@@ -32,11 +32,13 @@
 #define OPENSHOT_CACHE_BASE_H
 
 #include <memory>
-#include <cstdlib>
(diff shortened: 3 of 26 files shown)


ferdnyc:
<s>Codecov makes the excellent point that Crop should really be unit-tested, which is a reasonable demand if the tests follow the basic format I sketched out for ChromaKey in #730.</s>

Crop, now with unit tests!

jonoomph:
This looks great. It's hard to tell from the code, but hopefully this has 100% the same behavior as before. I worked really hard to match the behavior of our old clip properties (which I thought were extremely confusing to use)... But if it works the same exact way, with a better/simpler implementation, and unit tests... that is amazing!

ferdnyc:
@jonoomph The way cropping is parameterized is definitely confusing, though not as bad as it USED to be (where you had to both shift the image down/right AND add left/top cropping, to add black bars on those sides without the image appearing to move). I tested this reasonably thoroughly to match the old behavior, though I won't make any empty promises about all of its interactions with the various scaling modes and etc., because there's no chance I managed to test them all.

ferdnyc:
> (Actually, the algorithm here is slightly improved. With the previous implementation, setting both the left and right crop sizes to 0.5 left a thin sliver of visible image in the output. Ditto top and bottom. Here, that's eliminated, and left 0.5 + right 0.5 == crop out the entire frame.)

That's actually not quite correct. It DID happen, but usually not with `0.5` on each side. I pretty reliably got it to happen with `0.6` left crop and `0.4` right crop, though, which seems to point to a rounding error converting from float parameters. Regardless, it should be better now since the QPainter-based math bases the width/height values on `1.0` minus the sum of the crop sizes, and `1.0 - 0.6 - 0.4` is pretty reliably `0.0`.

(Maybe we should consider scaling up all of our percentage-based parameters by 1000, in a future revision of the project data format, so we're not dealing with values in the sub-`0.1` range quite so readily. We can even still **display** them like that, and just convert to integers -- or larger floats -- for/from the JSON data.)

PR #734, opened 2021-09-20
Windows and macOS CI builds

ferdnyc:
This PR (which is finally no longer WIP, but in the final stages of polishing and preparation to merge) builds on the recent switch from the Ubuntu daily PPA to a cached local build as the source of our libopenshot-audio dependency. Now that we're no longer tied to the Linux platform as our sole source of libraries built from the current `develop` branch source of libopenshot-audio, we're free to extend our CI offerings to the other supported platforms where libopenshot is used and expected to be tested/functional.

As such, these changes enable full libopenshot builds (including cached builds of the libopenshot-audio `develop` branch sources) _and_ unit test runs on both macOS (g++ and clang++) and on Windows (g++ only, due to issues building libopenshot-audio with MinGW clang++).

Coverage data is still not collected on either macOS or Windows, due to remaining issues running the collection tools on those platforms, so our coverage data is still entirely Linux-driven. Hopefully that can be addressed in a future PR.

P.S> (Yes, I will be squashing this branch's shamefully messy commit history prior to merge.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,98 +1,174 @@
 name: libopenshot CI Build
 on: [push, pull_request]
+
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.sys.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, ubuntu-20.04]
+        sys:
+          - { os: ubuntu-18.04, shell: bash }
+          - { os: ubuntu-latest, shell: bash }
+          - { os: macos-latest, shell: bash }
+          - { os: windows-latest, shell: 'msys2 {0}' }
         compiler:
-         - { cc: gcc, cpp: g++ }
-         - { cc: clang, cpp: clang++ }
+          - { cc: gcc, cxx: g++ }
+          - { cc: clang, cxx: clang++ }
+        exclude:
+          # Windows clang isn't being our friend,
+          # JUCE seems to think it can use _get_tzname there
+          # (it can't)
+          - sys: { os: windows-latest, shell: 'msys2 {0}' }
+            compiler: { cc: clang, cxx: clang++ }
+    defaults:
+      run:
+        shell: "${{ matrix.sys.shell }}"
     env:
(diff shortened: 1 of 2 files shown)


ferdnyc:
Collection of coverage data on macOS is still breaking, which is the _only_ thing keeping this PR in "WIP" state. As much as I'd like to get it working on _both_ Windows and macOS, I don't want to hold this up while solving that problem.

So, I've reset the configs to continue taking coverage data only during the linux builds (and only with GCC), and will work on coverage fixes for a future PR. Hopefully this next set of runs will all pass, and I can take this off WIP and get it ready for merging.

ferdnyc:
Squashed like a pancake and ready to go.

ferdnyc:
Be awesome if... _anyone_ could take a look at this... give it the ol' thumbs-up/thumbs-down... or find a problem, that'd be **great** actually, as I'd much rather find them _before_ merging.

Full build runs on all three OSes, including unit tests... pretty sweet right heah...

(Yeah, the Windows job runs for over **FORTY** minutes, which is insane — but since fully half of that time is spent installing dependencies, I don't think we're gonna be able to improve on it much.)

ferdnyc:
**ATTN:** @jonoomph et al,

At this point, the only thing holding me back from merging this (other than the fact that I need to squash the commits again), is that I'm headed out the door for the evening. But I'm hoping to do it later tonight or tomorrow. The macOS CI jobs have already revealed one bug in the code that was otherwise hidden (see #742), and my experiments in #740 are producing enlightening amounts of information about library handling on macOS. Having Windows and macOS builds generated for every PR can only improve the quality of the codebase.

So, one of those "speak now..." type situations.

eisneinechse:
From me it's a thumbs-up.

ferdnyc:
Just waitin' on Windows after a rebase onto `develop`, then I'm gonna light this candle. (The job's only been running for 30 minutes so far. "Only.")

Windows build times have sadly not improved much, though the time required to install dependencies can fluctuate some with network/server load. When the package-gulf-stream is favorable it can shave 10 minutes off the total job runtime.

PR #738, opened 2021-09-22
Protecting a caption effect with no Clip (i.e. effect added directly to timeline)

jonoomph:
This can't happen in openshot-qt, but it can definitely happen when using the OpenShot Cloud API or libopenshot directly, to add the Caption effect directly to a layer (affecting many different clips)


JacksonRG:
Guess I forgot to hit submit on that review. Better late than never :+1:

PR #742, opened 2021-09-27
Don't use right-aligned std::boolalpha

ferdnyc:
Turns out, macOS's libc++ doesn't apply fill padding / alignment the same way other libs do. [Known issue due to ambiguity in the standard.](https://timsong-cpp.github.io/lwg-issues/2703) So we have to manually transform booleans into `"true"` / `"false"` strings, in order to right-align them consistently.


ferdnyc:
Simple and obvious fix is simple, obvious.

PR #744, opened 2021-09-30
WIP: MacOS INSTALL_NAME_DIR/@rpath investigations

ferdnyc:
As the branch name notes, this is actually meant to be my exploration into using RPATH on macOS, something [CMake actually _recommends_](https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_RPATH.html#prop_tgt:MACOSX_RPATH) these days. We're still forcibly disabling it.

Instead, we set `INSTALL_NAME_DIR` to `"${CMAKE_INSTALL_PREFIX}/lib"`, which is just _suuuper_ wrong. At the very least, we should be using `"$<INSTALL_PREFIX>/lib"` so that the path is configured (the same way CMake relinks Linux libraries) at install time.

This PR (not currently meant to be merged) experimentally removes the disabling of `MACOSX_RPATH` project-wide, as well as the setting of `INSTALL_NAME_DIR` on our libraries. The goal is to test whether builds on MacOS that use the default CMake configuration (setting the library name to `@rpath/libNAME.dylib` and referencing under that same name when linking from other libs/binaries) actually gives us more reliable and flexible runtime linking. A corresponding branch in OpenShot/libopenshot-audio contains similar changes, and is built as the CI dependency for this branch.

I've created a script `/cmake/scan-mac-libs.sh` that's run after the build finishes, the output of that script is displayed by the "macOS Library checks" build step.

Some other experimental changes are also included, like switching from `lcov` to `gcovr` as the coverage-reporting tool, and using the Cobertura XML output format for the data. The hope is that it will be easier to get at least macOS (and possibly Windows) coverage working, if we use this better-supported format for reporting.


ferdnyc:
Here's what this branch is about, from my perspective. (I sent this in email to @jonoomph but I thought it was worth getting On The Record, so to speak. And a lot of this is very theoretical, so if anyone has any practical experience or advice on any of this, that'd be great and hopefully save us some wasted time going down wrong paths.)

#### MACOSX_RPATH

CMake have actually recommended using `@rpath` on macOS for a while now, apparently since around macOS 10.5 or 10.6 support matured enough that they now consider it "a more flexible and powerful alternative to `@executable_path` and `@loader_path`".

AIUI, the theory is that the best way to leverage macOS's two-stage loading system for shared libraries is to use `@rpath` in your library names. Used correctly, it makes them fully relocatable and you shouldn't need to adjust them, just the binaries that load them. The executables get built with one or more `LC_RPATH`, pointing them to the location of the libraries. And those paths can use `@executable_path` to point to a location relative to the executable, avoiding any fixed paths.

So, f'rinstance, if a library dir at `/Applications/OpenShot.app/lib/` is filled with .dylib files that have `@rpath` install names, like so....

```text
/Applications/OpenShot.app/lib/:
  - libobjc.A.dylib (INSTALL_NAME: @rpath/libobjc.A.dylib)
  - libz.1.dylib (INSTALL_NAME: @rpath/libz.1.dylib)
  - libc++.1.dylib (INSTALL_NAME: @rpath/libc++.1.dylib)
  - libSystem.B.dylib (INSTALL_NAME: @rpath/libSystem.B.dylib)
  - libopenshot-audio.8.dylib (INSTALL_NAME: @rpath/libopenshot.audio.dylib)
    otool -L dependencies:
        @rpath/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
        @rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
        @rpath/libc++.1.dylib (compatibility version 1.0.0, current version 904.4.0)
        @rpath/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.60.1)
  - libopenshot.21.dylib (INSTALL_NAME: @rpath/libopenshot.21.dylib)
    otool -L dependencies:
        @rpath/libopenshot-audio.8.dylib (compatibility version 8.0.0, current version 0.2.2)
        @rpath/libavcodec.58.dylib (compatibility version 58.0.0, current version 58.134.100)
        @rpath/libavformat.58.dylib (compatibility version 58.0.0, current version 58.76.100)
        @rpath/libavutil.56.dylib (compatibility version 56.0.0, current version 56.70.100)
        @rpath/libswscale.5.dylib (compatibility version 5.0.0, current version 5.9.100)
        @rpath/libswresample.3.dylib (compatibility version 3.0.0, current version 3.9.100)
        @rpath/libzmq.5.dylib (compatibility version 8.0.0, current version 8.4.0)
        @rpath/libopencv_tracking.4.5.dylib (compatibility version 4.5.0, current version 4.5.3)
        @rpath/libopenshot_protobuf.21.dylib (compatibility version 21.0.0, current version 0.2.7)
        @rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
   (and so on...)
```

And if `/Applications/OpenShot.app/bin/openshot-qt` contains an `LC_RPATH` entry that leads it to those libraries...

```text
/Applications/OpenShot.app/bin/:
    - openshot-qt (LC_RPATH=@executable_path/../lib/)
```

Then the openshot-qt executable can find all of its library dependencies, and the entire OpenShot.app bundle can still be located anywhere on disk because it remains fully self-contained by relying entirely on relative paths. More importantly, all of the `@rpath`-named libraries can be moved relative to the binary, or packaged in a different way with a completely different binary. The only adjustment that's required is to add/update an `LC_RPATH` entry in the executable. None of the libraries should ever need to be changed at all.

...That's the theory, anyway. I'm curious to find out whether it's correct. But left to its own devices (meaning, if we stop overriding the default behavior) CMake since version 3.0 (!) will produce libraries exactly this way: using an `INSTALL_NAME` of `@rpath/filename.dylib`.

The ultimate goal, if all this pans out, would be to create a libopenshot package for Python, which included not only the bindings but also all of their library dependencies, bundled right inside. (The same way PyQt5 contains all of the Qt libs, if you install it using pip.) That way, when cx_Freeze pulls that package into the final build, the libraries are already all included.

It shouldn't be the responsibility of any part of the OpenShot build process to deal with shared library dependencies, since it DOESN'T depend on any of those libraries. It only depends on `openshot.py` (...which depends on `_openshot.so` which depends on `libopenshot.dylib` which depends on everything else). But all of that consolidation can be done way before cx_Freeze even comes into the picture.

I'm confident the same thing can be achieved for Linux, using libraries built with ELF `RPATH`s (actually `RUNPATH`s).

Whether or not it's doable for Windows... we'll have to see about that. Windows may have to be special forever.

(But one big reason to take care of this in the context of libopenshot: CMake 3.20 and 3.21 added a whole set of new tools specifically intended for use in managing targets' runtime dependencies.) You can now do things like:

* Copy all of the Windows DLL requirements for the Python bindings into their build dir, after compiling _openshot.pyd:
    ```cmake
    add_custom_command(TARGET pyopenshot POST_BUILD
      COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_RUNTIME_DLLS:pyopenshot> $<TARGET_FILE_DIR:pyopenshot>
CMake have actually recommended using @rpath on macOS for a while now, apparently since around macOS 10.5 or 10.6 support matured enough that they now consider it "a more flexible and powerful alternative to @executable_path and @loader_path".


AIUI, the theory is that the best way to leverage macOS's two-stage loading system for shared libraries is to use @rpath in your library names. Used correctly, it makes them fully relocatable and you shouldn't need to adjust them, just the binaries that load them. The executables get built with one or more LC_RPATH, pointing them to the location of the libraries. And those paths can use @executable_path to point to a location relative to the executable, avoiding any fixed paths.


So, f'rinstance, if a library dir at /Applications/OpenShot.app/lib/ is filled with .dylib files that have @rpath install names, like so....


/Applications/OpenShot.app/lib/:

  - libobjc.A.dylib (INSTALL_NAME: @rpath/libobjc.A.dylib)
  - libz.1.dylib (INSTALL_NAME: @rpath/libz.1.dylib)

  - libc++.1.dylib (INSTALL_NAME: @rpath/libc++.1.dylib)

  - libSystem.B.dylib (INSTALL_NAME: @rpath/libSystem.B.dylib)

  - libopenshot-audio.8.dylib (INSTALL_NAME: @rpath/libopenshot.audio.dylib)
    otool -L dependencies:

        @rpath/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
        @rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
        @rpath/libc++.1.dylib (compatibility version 1.0.0, current version 904.4.0)
        @rpath/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.60.1)
  - libopenshot.21.dylib (INSTALL_NAME: @rpath/libopenshot.21.dylib)

    otool -L dependencies:

        @rpath/libopenshot-audio.8.dylib (compatibility version 8.0.0, current version 0.2.2)

        @rpath/libavcodec.58.dylib (compatibility version 58.0.0, current version 58.134.100)
        @rpath/libavformat.58.dylib (compatibility version 58.0.0, current version 58.76.100)
        @rpath/libavutil.56.dylib (compatibility version 56.0.0, current version 56.70.100)
        @rpath/libswscale.5.dylib (compatibility version 5.0.0, current version 5.9.100)
        @rpath/libswresample.3.dylib (compatibility version 3.0.0, current version 3.9.100)
        @rpath/libzmq.5.dylib (compatibility version 8.0.0, current version 8.4.0)
        @rpath/libopencv_tracking.4.5.dylib (compatibility version 4.5.0, current version 4.5.3)
        @rpath/libopenshot_protobuf.21.dylib (compatibility version 21.0.0, current version 0.2.7)
        @rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)

   (and so on...)


And if /Applications/OpenShot.app/bin/openshot-qt contains an LC_RPATH entry that leads it to those libraries...


/Applications/OpenShot.app/bin/:

    - openshot-qt (LC_RPATH=@executable_path/../lib/)


Then the openshot-qt executable can find all of its library dependencies, and the entire OpenShot.app bundle can still be located anywhere on disk because it remains fully self-contained by relying entirely on relative paths. More importantly, all of the @rpath-named libraries can be moved relative to the binary, or packaged in a different way with a completely different binary. The only adjustment that's required is to add/update an LC_RPATH entry in the executable. None of the libraries should ever need to be changed at all.


...That's the theory, anyway. I'm curious to find out whether it's correct. But left to its own devices (meaning, if we stop overriding the default behavior) CMake since version 3.0 (!) will produce libraries exactly this way: using an INSTALL_NAME of @rpath/filename.dylib.


The ultimate goal, if all this pans out, would be to create a libopenshot package for Python, which included not only the bindings but also all of their library dependencies, bundled right inside. (The same way PyQt5 contains all of the Qt libs, if you install it using pip.) That way, when cx_Freeze pulls that package into the final build, the libraries are already all included.


It shouldn't be the responsibility of any part of the OpenShot build process to deal with shared library dependencies, since it DOESN'T depend on any of those libraries. It only depends on openshot.py (...which depends on _openshot.so which depends on libopenshot.dylib which depends on everything else). But all of that consolidation can be done way before cx_Freeze even comes into the picture.


I'm confident the same thing can be achieved for Linux, using libraries built with ELF RPATHs (actually RUNPATHs).


Whether or not it's doable for Windows... we'll have to see about that. Windows may have to be special forever.


(But one big reason to take care of this in the context of libopenshot: CMake 3.20 and 3.21 added a whole set of new tools specifically intended for use in managing targets' runtime dependencies.) You can now do things like:

* Copy all of the Windows DLL requirements for the Python bindings into their build dir, after compiling `_openshot.pyd`:
    ```cmake
     add_custom_command(TARGET pyopenshot POST_BUILD
      COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_RUNTIME_DLLS:pyopenshot> $<TARGET_FILE_DIR:pyopenshot>
      COMMAND_EXPAND_LISTS
      )
    ```

* Or bundle e.g. the FFmpeg libraries at install time (assuming necessary RPATHs are set, on Linux/macOS):
    ```cmake

    install(IMPORTED_RUNTIME_ARTIFACTS
      FFmpeg::avcodec FFmpeg::avformat FFmpeg::avutil FFmpeg::swscale FFmpeg::swresample
  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
  )

      COMMAND_EXPAND_LISTS
  )
    ```
* Or bundle e.g. the FFmpeg libraries at install time (assuming necessary RPATHs are set, on Linux/macOS):
    ```cmake
    install(IMPORTED_RUNTIME_ARTIFACTS
      FFmpeg::avcodec FFmpeg::avformat FFmpeg::avutil FFmpeg::swscale FFmpeg::swresample
      LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
      )
    ```

ferdnyc:
Closing this as I now have better local methods of investigation.

PR #747, opened 2021-10-03
WIP: CI: Switch to gcovr coverage, enable on macOS

ferdnyc:
By switching our coverage reporting tool from lcov to gcovr, and using the de facto standard Cobertura XML reporting format instead of lcov's cryptic text file, we're able to collect and incorporate test coverage on macOS, in addition to Linux.

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -699,7 +699,9 @@ std::shared_ptr<Frame> Clip::GetOrCreateFrame(int64_t number)
 {
 	try {
 		// Debug output
-		ZmqLogger::Instance()->AppendDebugMethod("Clip::GetOrCreateFrame (from reader)", "number", number);
+		ZmqLogger::Instance()->AppendDebugMethod(
+			"Clip::GetOrCreateFrame (from reader)",
+			"number", number);
 
 		// Attempt to get a frame (but this could fail if a reader has just been closed)
 		auto reader_frame = reader->GetFrame(number);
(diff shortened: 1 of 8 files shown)


ferdnyc:
So the thing with gcovr is that it tracks _branch_ coverage, not just line coverage like lcov tracks. (gcovr _also_ reports line coverage, but separately includes branch coverage stats as well.)

```c++
// A simple conditional like this has four branches,
// corresponding to the four possible permutations
// of truth values for the test.
if (clip->Parent() && clip->Parent() == target_clip) { ... }

// And this function call? This one call has NINE total branches
ZmqLogger::Instance()->AppendDebugMethod(
    "Timeline::GetFrame (Add frame to cache)",
    "requested_frame", requested_frame,
    "info.width", info.width,
    "info.height", info.height);
```

Unless all branches are taken at least once by the test code, a line isn't considered fully covered. That's the source of the new "Partials" stat, which didn't exist in the older coverage reporting. Those partials correspond to the huge drop in "Hits" (which now represent only _fully_-covered lines).

Branch coverage can be somewhat misleading, though, because (to quote the [gcovr documentation](https://gcovr.com/en/stable/faq.html)) "100% branch coverage will be impossible for most programs." Not just "unrealistic" or "incredibly difficult", but provably, _mathematically impossible_.

For example, take this condition:
```c++
if (clip && clip->Parent()) { ... }
```
One of the four branches on that line can **never** be tested — it's impossible to test the `false && true` branch, because that's an impossible arrangement of values. The branch is unreachable. 

Anytime we take advantage of short-circuit evaluation, we're technically creating unreachable branches — in fact, **we're _relying_ on it**. The branch that's (intentionally) made unreachable by the short-circuit is the one that would segfault if it _were_ ever taken. Yet it still gets counted as a branch. (There **are** at least some command-line options that tell gcovr to ignore certain types of untestable branches, so they won't be counted against the final stats.)

I'm still trying to figure out if I can convince Codecov to show gcovr's line-coverage stats, instead of the branch-coverage stats — just so our coverage doesn't suddenly drop from ~50% to < 30%. Branch-based stats may initially seem like they're more accurate, but really it's just an overcompensation in the other direction. While it's true that line-based counting overreports coverage, branch-based underreports it at least as severely.

jonoomph:
@ferdnyc Is this one ready for merging? I couldn't really tell from reviewing it. Looked a bit experimental, lol

ferdnyc:
@jonoomph Yeah, it's very experimental. 

TBH I'm just not really happy with how gcovr reports coverage data, due to the branch-coverage issue. 

I definitely still think there's a _lot_ of value to be found from collecting coverage on all three OSes. (Especially considering the amount of platform-specific code we have `#ifdef`'d, so the only way to get coverage on it is to collect data separately on each OS.) But even gcovr doesn't get us all the way there (it just adds macOS), and not without its share of down sides.

PR #751, opened 2021-10-07
Prevent crash when smallest frame detected in memory object is NULL

jonoomph:
Better protect video caching thread from accessing a NULL smallest frame object. I think this is the cause of many crashes when heavy use of OpenShot (i.e. fast scrubbing, fast forward, fast backward... anything too fast which might result in a data change, and thus a clearing of the cache while the video caching thread is accessing it)

### DIFF
--- a/src/CacheMemory.cpp
+++ b/src/CacheMemory.cpp
@@ -168,7 +168,6 @@ std::shared_ptr<Frame> CacheMemory::GetSmallestFrame()
 {
 	// Create a scoped lock, to protect the cache from multiple threads
 	const GenericScopedLock<CriticalSection> lock(*cacheCriticalSection);
-	std::shared_ptr<openshot::Frame> f;
 
 	// Loop through frame numbers
 	std::deque<int64_t>::iterator itr;
@@ -179,10 +178,12 @@ std::shared_ptr<Frame> CacheMemory::GetSmallestFrame()
 			smallest_frame = *itr;
 	}
 
-	// Return frame
-	f = GetFrame(smallest_frame);
-
-	return f;
+	// Return frame (if any)
+	if (smallest_frame != -1) {
+        return GetFrame(smallest_frame);
+    } else {
+	    return NULL;
+	}
 }
 
 // Gets the maximum bytes value
--- a/src/Qt/VideoCacheThread.cpp
+++ b/src/Qt/VideoCacheThread.cpp
@@ -86,6 +86,7 @@ namespace openshot
     void VideoCacheThread::run()
     {
         // Types for storing time durations in whole and fractional milliseconds
+        std::shared_ptr<openshot::Frame> smallest_frame = NULL;
         using ms = std::chrono::milliseconds;
         using double_ms = std::chrono::duration<double, ms::period>;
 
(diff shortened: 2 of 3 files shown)


JacksonRG (approved):
Looks safe, GTM!

src/CacheMemory.cpp
```diff
@@ -179,10 +178,12 @@ std::shared_ptr<Frame> CacheMemory::GetSmallestFrame()
...
-	// Return frame
-	f = GetFrame(smallest_frame);
-
-	return f;
+	// Return frame (if any)
+	if (smallest_frame != -1) {
+        return GetFrame(smallest_frame);
+    } else {
+	    return NULL;
+	}
```
ferdnyc, line 186: Identical code here makes me think this should be in CacheBase, no? Then it only has to be written once. When will these implementations ever differ on this?

Or, if the base class is to remain pure, then a lot of the copy-paste code in the two cache implementations is really crying out for abstraction into a common class. (A `CacheFrameList` that they both use to manage their range lists, instead of implementing it directly in both?)

I mean, `CalculateRanges()`, just as an example, is literally identical between the two classes. That's not supposed to be the result of pure-abstract base classes.

It would be better to break the abstraction's purity and put it in `CacheBase`, than to maintain two copies of the exact same code<sup>1</sup>. (There are, of course, other solutions, like a common `CacheFrameList`.)

1. ...At least, exact same _for now_. History tells us that they'll slowly diverge over time, in small ways that can lead to subtle and hard-to-track-down differences based on which theoretically-interchangeable class is involved.

src/CacheMemory.cpp
```diff
@@ -179,10 +178,12 @@ std::shared_ptr<Frame> CacheMemory::GetSmallestFrame()
...
 
-	// Return frame
-	f = GetFrame(smallest_frame);
-
-	return f;
+	// Return frame (if any)
+	if (smallest_frame != -1) {
+        return GetFrame(smallest_frame);
+    } else {
+	    return NULL;
```
ferdnyc, line 185: `nullptr`, please. It's 2021. Since C++11 `NULL` _can_ be `#define`d to `nullptr`, or it could be something entirely different. But `nullptr` always means `nullptr`, and [won't let you get away with passing things like `0` or `NULL` where you're not allowed to](https://en.cppreference.com/w/cpp/language/nullptr#Example) (a good thing).

src/Qt/VideoCacheThread.cpp
```diff
@@ -86,6 +86,7 @@ namespace openshot
     void VideoCacheThread::run()
     {
         // Types for storing time durations in whole and fractional milliseconds
+        std::shared_ptr<openshot::Frame> smallest_frame = NULL;
```
ferdnyc, line 89: Also `nullptr`.

src/Qt/VideoCacheThread.cpp
```diff
@@ -107,13 +108,11 @@ namespace openshot
...
-							if (smallest_cached_frame > current_display_frame) {
-								// Cache position has gotten too far away from current display frame.
-								// Reset the position to the current display frame.
-								position = current_display_frame;
-							}
+                        smallest_frame = reader->GetCache()->GetSmallestFrame();
+                        if (smallest_frame && smallest_frame->number > current_display_frame) {
+                            // Cache position has gotten too far away from current display frame.
+                            // Reset the position to the current display frame.
+                            position = current_display_frame;
```
ferdnyc, line 115: Nit: Indentation is clearly all messed up here.

ferdnyc (commented):
This is already merged, so I'm sure there's no point in submitting this. Leave a shout-out if you read this, though (the accompanying _code comments_,  not **this** note) and be eligible to win a special surprise gift.

(Spoiler: The gift will turn out to be a lie. Surprise.)

PR #752, opened 2021-10-07
Better detection of duration, especially for audio-only files

jonoomph:
Better detection of duration, especially for audio-only files. Also, improved PTS offset detection (for when start_time is used), mainly things recorded from streams. Better bitrate detection as well.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -645,14 +645,29 @@ void FFmpegReader::UpdateAudioInfo() {
 	info.channel_layout = (ChannelLayout) AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->channel_layout;
 	info.sample_rate = AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->sample_rate;
 	info.audio_bit_rate = AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->bit_rate;
+	if (info.audio_bit_rate <= 0) {
+	    // Get bitrate from format
+        info.audio_bit_rate = pFormatCtx->bit_rate;
+	}
 
 	// Set audio timebase
 	info.audio_timebase.num = aStream->time_base.num;
 	info.audio_timebase.den = aStream->time_base.den;
 
 	// Get timebase of audio stream (if valid) and greater than the current duration
-	if (aStream->duration > 0.0f && aStream->duration > info.duration)
-		info.duration = aStream->duration * info.audio_timebase.ToDouble();
+	if (aStream->duration > 0 && aStream->duration > info.duration) {
+	    // Get duration from audio stream
+        info.duration = aStream->duration * info.audio_timebase.ToDouble();
+    } else if (pFormatCtx->duration > 0 && info.duration <= 0.0f) {
+        // Use the format's duration
+        info.duration = float(pFormatCtx->duration) / AV_TIME_BASE;
+    }
+
+    // Calculate duration from filesize and bitrate (if any)
+    if (info.duration <= 0.0f && info.video_bit_rate > 0 && info.file_size > 0) {
+        // Estimate from bitrate, total bytes, and framerate
+        info.duration = float(info.file_size) / info.video_bit_rate;
+    }
 
 	// Check for an invalid video length
 	if (info.has_video && info.video_length <= 0) {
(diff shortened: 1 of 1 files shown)


src/FFmpegReader.cpp
```diff
@@ -645,14 +645,29 @@ void FFmpegReader::UpdateAudioInfo() {
...
+	}
 
 	// Set audio timebase
 	info.audio_timebase.num = aStream->time_base.num;
 	info.audio_timebase.den = aStream->time_base.den;
 
 	// Get timebase of audio stream (if valid) and greater than the current duration
-	if (aStream->duration > 0.0f && aStream->duration > info.duration)
-		info.duration = aStream->duration * info.audio_timebase.ToDouble();
+	if (aStream->duration > 0 && aStream->duration > info.duration) {
```
JacksonRG, line 658: Just for my info, what's the info object we're comparing to?

src/FFmpegReader.cpp
```diff
@@ -1843,20 +1860,25 @@ void FFmpegReader::UpdatePTSOffset(bool is_video) {
...
-			// TODO: Please see https://github.com/OpenShot/libopenshot/pull/565#issuecomment-690985272
-			// for ideas to improve this logic.
-			int64_t max_offset = info.video_timebase.Reciprocal().ToFloat();
-			if (video_pts_offset < -max_offset || video_pts_offset > max_offset) {
-				// Ignore PTS, it seems invalid
-				video_pts_offset = 0;
-			}
+            if (pStream->start_time != AV_NOPTS_VALUE && pStream->start_time > 0) {
+                // Adjust all PTS by start_time (if available)
+                video_pts_offset = 0 - pStream->start_time;
```
JacksonRG, line 1865: Another question, just for my info, what are we offsetting the PTS relative to?

JacksonRG (approved):
Couple questions, but don't see issues

Edit b/c that wasn't very enthusiastic:  Nice work!

src/FFmpegReader.cpp
jonoomph: The info struct is on base classes for Reader and Writers. It usually contains the known or requested features of a file.

For example, the ReaderBase
```
	struct ReaderInfo
	{
		bool has_video;				///< Determines if this file has a video stream
		bool has_audio;				///< Determines if this file has an audio stream
		bool has_single_image;		///< Determines if this file only contains a single image
		float duration;				///< Length of time (in seconds)
		int64_t file_size;		///< Size of file (in bytes)
		int height;					///< The height of the video (in pixels)
		int width;					///< The width of the video (in pixesl)
		int pixel_format;			///< The pixel format (i.e. YUV420P, RGB24, etc...)
		openshot::Fraction fps;				///< Frames per second, as a fraction (i.e. 24/1 = 24 fps)
		int video_bit_rate;			///< The bit rate of the video stream (in bytes)
		openshot::Fraction pixel_ratio;		///< The pixel ratio of the video stream as a fraction (i.e. some pixels are not square)
		openshot::Fraction display_ratio;		///< The ratio of width to height of the video stream (i.e. 640x480 has a ratio of 4/3)
		std::string vcodec;				///< The name of the video codec used to encode / decode the video stream
		int64_t video_length;		///< The number of frames in the video stream
		int video_stream_index;		///< The index of the video stream
		openshot::Fraction video_timebase;	///< The video timebase determines how long each frame stays on the screen
		bool interlaced_frame;		// Are the contents of this frame interlaced
		bool top_field_first;		// Which interlaced field should be displayed first
		std::string acodec;				///< The name of the audio codec used to encode / decode the video stream
		int audio_bit_rate;			///< The bit rate of the audio stream (in bytes)
		int sample_rate;			///< The number of audio samples per second (44100 is a common sample rate)
		int channels;				///< The number of audio channels used in the audio stream
		openshot::ChannelLayout channel_layout;	///< The channel layout (mono, stereo, 5 point surround, etc...)
		int audio_stream_index;		///< The index of the audio stream
		openshot::Fraction audio_timebase;	///< The audio timebase determines how long each audio packet should be played
		std::map<std::string, std::string> metadata;	///< An optional map/dictionary of metadata for this reader
	};
``

jonoomph: Essentially, it's a convenient place to store all this data, both for our FFmpegReader, and FFmpegWriter

src/FFmpegReader.cpp
jonoomph: So, the PTS values can be anything... negative, positive, small, huge, etc...  The timestamps are also relative to a specific timebase (1/1000 of a second, 1/10000000 of a second, etc...). The timebase can be anything also. Ultimately, we don't care exactly what the 1st PTS value is... but we want that to be the beginning of our playback data (i.e. in our editor, all playback starts at timestamp 0)

jonoomph: For example, let's say you are watching a stream, and it's on second 99999999 of the stream currently, and you record a little bit of the stream. We want to offset the "start_time" and initial PTS value, and not wait for second 99999999 to eventually happen.

PR #753, opened 2021-10-07
Support for anamorphic profiles, i.e. non-square pixels (including waveform fixes)

jonoomph:
Support for anamorphic profiles, i.e. non-square pixels. Fixes include Frame::Display, Frame::Save, and Timeline::SetMaxSize, all which had various issues. Added sample_aspect_ratio to FFmpegWriter, so the exported videos will continue to be anamorphic.

Also fixed some waveform issues, which were not appearing, or appearing squished, and sometimes containing gaps.


JacksonRG (approved):
LGTM!

PR #755, opened 2021-10-08
Fix SVG Cmake regression (we don't need QtSvg)

jonoomph:
Fix SVG cmake regression (we don't need QtSvg to be installed for libopenshot to work), and our Linux build server does not include it. @ferdnyc This one is related to https://github.com/OpenShot/libopenshot/pull/731.

**Maybe this needs some additional thought...** 
libopenshot + QtSvg - Resvg == GOOD
libopenshot - QtSvg + Resvg == GOOD
libopenshot - QtSvg - Resvg == BAD :man_shrugging:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -284,7 +284,7 @@ endif ()
 
 ################# QT5 ###################
 # Find QT5 libraries
-set(_qt_components Core Gui Widgets Svg)
+set(_qt_components Core Gui Widgets)
 find_package(Qt5 COMPONENTS ${_qt_components} REQUIRED)
 
 foreach(_qt_comp IN LISTS _qt_components)


ferdnyc:
@jonoomph 

Ooh, you're right, I didn't think about Resvg — it should be one _or_ the other. (It's not quite true that libopenshot doesn't need QtSvg, though... it does, to be able to render SVGs, unless it has Resvg.)

But it should absolutely be conditional on Resvg **not** being installed, I can fix that pretty easily.

ferdnyc:
@jonoomph See #756.

ferdnyc:
This reminds me, though, that I really should add a build (or builds) to our GitHub CI matrix that includes Resvg, to test this sort of thing.

jonoomph:
Thanks a bunch @ferdnyc! Closing this PR

PR #756, opened 2021-10-08
Include either Resvg or QtSVG, not both

ferdnyc:
This should fix things so QtSvg isn't required if Resvg is installed, which I broke in #731.


jonoomph:
That is perfect! Feel free to smash the merge button once it's ready! :smile:

ferdnyc:
Done! In exactly 30 minutes. Not bad, for building two Mac versions and one Windows version (in addition to four Linux permutations). Merging.

PR #757, opened 2021-10-08
Resvg: Add support for v0.11.0+, enable in GitHub CI

ferdnyc:
This PR adds in support for Resvg version `v0.11.0` and later (up to the current `v0.19.0`), which were previously  unsupported due to changes in the API. This makes Resvg much easier to integrate, as the old versions we _have_ been relying on require patching to be compatible with the current Qt5 LTS (5.15.x series).

In addition, the Resvg Find module is enhanced with version detection and reporting (if available, Resvg v0.5.0 and earlier didn't report the version number in the headers so it isn't available).

The Linux clang++ builds in our CI matrix are extended to perform a cached build of Resvg, and to use it when setting up libopenshot so that we can unit-test the integration.

### DIFF
--- a/cmake/Modules/FindResvg.cmake
+++ b/cmake/Modules/FindResvg.cmake
@@ -41,6 +41,28 @@ Copyright (c) 2020, FeRD (Frank Dana) <ferdnyc@gmail.com>
 #]=======================================================================]
 include(FindPackageHandleStandardArgs)
 
+### Macro: parse_resvg_version
+#
+# Read the resvg.h file and extract the definition
+# for RESVG_VERSION, to use as our version string.
+macro (parse_resvg_version)
+  set(_header "${Resvg_INCLUDE_DIRS}/resvg.h")
+  if(EXISTS "${_header}")
+    #message(STATUS "Parsing Resvg version from ${_header}")
+    file(STRINGS "${_header}" _version_def
+      REGEX "^#define[ \t]+RESVG_VERSION[ \t]+\".*\"[ \t]*$")
+    string(REGEX REPLACE
+      "^.*RESVG_VERSION[ \t]+\"(.*)\".*$"
+      "\\1"
+      Resvg_VERSION "${_version_def}")
+    #message(STATUS "Found Resvg version: ${Resvg_VERSION}")
+  endif()
+endmacro()
+
+###
+### Begin discovery
+###
+
 # CMake 3.4+ only: Convert relative paths to absolute
 if(DEFINED RESVGDIR AND CMAKE_VERSION VERSION_GREATER 3.4)
   get_filename_component(RESVGDIR "${RESVGDIR}" ABSOLUTE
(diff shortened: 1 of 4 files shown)


ferdnyc:
@jonoomph Speaking of SVG rendering... 

Now we can finally upgrade that gimpy, ancient version of Resvg from like 2018 that the official builds are still using. (Newer version(s) — which are all based on a rust reimplementation of the Skia rendering backend called tinyskia, to reduce external dependencies and eliminate the complexity of the different backends —  even have the ability to load TrueType fonts into the internal font manager. I haven't exposed that in the libopenshot API yet but it's something to consider for installations where the Ubuntu font is bundled with OpenShot, and not part of the system font library.)

cmake/Modules/FindResvg.cmake
```diff
@@ -41,6 +41,28 @@ Copyright (c) 2020, FeRD (Frank Dana) <ferdnyc@gmail.com>
...
+# Read the resvg.h file and extract the definition
+# for RESVG_VERSION, to use as our version string.
+macro (parse_resvg_version)
+  set(_header "${Resvg_INCLUDE_DIRS}/resvg.h")
+  if(EXISTS "${_header}")
+    #message(STATUS "Parsing Resvg version from ${_header}")
+    file(STRINGS "${_header}" _version_def
+      REGEX "^#define[ \t]+RESVG_VERSION[ \t]+\".*\"[ \t]*$")
+    string(REGEX REPLACE
+      "^.*RESVG_VERSION[ \t]+\"(.*)\".*$"
```
jonoomph, line 55: Nice! :smile:

cmake/Modules/FindResvg.cmake
```diff
@@ -82,20 +107,24 @@ find_library(Resvg_LIBRARIES
...
-  "Could NOT find Resvg, using Qt SVG parsing instead"
-  Resvg_LIBRARIES Resvg_INCLUDE_DIRS )
+  REQUIRED_VARS Resvg_LIBRARIES Resvg_INCLUDE_DIRS
+  VERSION_VAR Resvg_VERSION
+)
 
 # Export target
 if(Resvg_FOUND AND NOT TARGET Resvg::Resvg)
-  message(STATUS "Creating IMPORTED target Resvg::Resvg")
+  #message(STATUS "Creating IMPORTED target Resvg::Resvg")
```
jonoomph, line 127: Can we delete this comment? Or is it still useful?

jonoomph:
LGTM, just a nitpick on a comment

jonoomph:
Gonna merge, and then I'll remove the comment in a separate PR

jonoomph:
@ferdnyc I'm upgrading the resvg versions on all our builders tonight. :crossed_fingers:

jonoomph:
Done updating builders with latest version of resvg. Just did a quick test, and no build errors! Please let me know if you see any issues with the new resvg libraries.

PR #758, opened 2021-10-16
CI: Validate repository licensing with Reuse

ferdnyc:
Using a pre-packaged action from the Marketplace, the workflow will validate that the repo remains in compliance with the [Reuse standards](https://reuse.software/) for license management.

All new files should be tagged with both copyright and license identifier metadata, either directly via in-source comments, or for binary files, with a metadata block inserted into the `.reuse/dep5` file in the repo.

For more details see the existing source files, as well as the `.reuse/` and `LICENSES/` subdirs. Or just ping me for more info.

### DIFF
--- a/.github/workflows/reuse.yml
+++ b/.github/workflows/reuse.yml
@@ -1,3 +1,7 @@
+# © OpenShot Studios, LLC
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
 # Workflow to run 'reuse lint' via the reuse-action script,
 # which will flag any files added or changed in the repo
 # that don't have valid, verifiable license data associated
--- a/.github/workflows/reuse.yml
+++ b/.github/workflows/reuse.yml
@@ -1,0 +1,24 @@
+# Workflow to run 'reuse lint' via the reuse-action script,
+# which will flag any files added or changed in the repo
+# that don't have valid, verifiable license data associated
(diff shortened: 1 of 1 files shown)


ferdnyc:
HAHAHAHAHAHAHAHAHA! :rofl: 

You see, when I added the workflow file for Reuse, I neglected to tag it with license and copyright metadata. So it immediately caused the repo to fail its own checks.

Oops. (Hey, this stuff takes some getting used to. That's why it's so important to  have automated monitoring in place.)

I'll just go tag that file now.

ferdnyc:
@jonoomph 

I'm going to merge this as just a regular CI job workflow now, so its checks will run on every PR from now on. But to get the most out of Reuse, personally I'd like to see the check be a PR-merge blocker. 

_ALL_ that's required to pass its validations is that every file in the repo (_**every** file_) have valid license terms associated with it, and that all of the license texts for all of those licenses be located in the `/LICENSES/` directory at the root of the repo.

That's not even remotely an onerous requirement to impose, though (as I've already demonstrated in this very PR) it's super easy to forget, which is why having a digital nanny that keeps us honest is a good thing.

Up to you, though. You alone have the power to bless the `reuse.yml` workflow as a merge requirement for PRs. (Which any of us with commit access can always manually override, anyway — unless that power is also separately _dis_abled for the target branch(es), of course.)

PR #760, opened 2021-10-16
Remove BlackMagic DeckLink support entirely

ferdnyc:
So, earlier while testing some other changes, I had opportunity (and need) to download the BlackMagic SDK and attempt to build with `ENABLE_BLACKMAGIC` set.

And what  I discovered was that the code is completely broken. Like It's broken for _our_ classes. One of the methods in `DecklinkInput` tries to call `.Exists()` on a CacheMemory object, a method that — ironically — does _not_ `Exists()` in the `CacheMemory` or `CacheBase` class _at all_.

It's also coded around an ancient version of the BlackMagic SDK that I suspect isn't even compatible with current ones, but that's really beside the point. If it isn't even valid _**libopenshot**_ code, there's no reason to keep it in the library. It could never work even with the expected version of the SDK unless its OpenShot-side interfaces were fixed, meaning it's unusable in its current form.

So, this PR removes all of the `Decklink*` source files and the `examples/ExampleBlackmagic.cpp` file, deletes the headers from `OpenShot.h` and the `openshot.i` language bindings, deletes `FindBlackMagic.cmake`, and removes all mentions of BlackMagic from all of the `CMakeLists.txt` files. (It also originally attempted to deprecate the DecklinkError exception, but I screwed that code up somehow and had to back it out. The exception doesn't really hurt anything just sitting there silently unused, anyway.)

If someone ever wants to revive it and fix the code, they can always pull it out of the repo's commit history.

cc: @jonoomph


ferdnyc:
I'm going to merge this, since I've demonstrated that the Decklink code in its current state cannot be compiled with any version of their SDK. It even tries to use _libopenshot_ APIs that don't exist.

PR #763, opened 2021-10-27
Clip: Fix has_video, add unit test

ferdnyc:
Turns out, the ability to set `has_video` keyframes on Clip objects has been broken since the most recent compositing changes — it's ignored and the Clip always outputs video if the Reader provides any.

This PR fixes `has_video` by having `Clip::GetOrCreateFrame()` re-fill the copied Frame's image with Qt::transparent, if `has_video` evaluates to `0` for that frame.

It also fills the audio with silence, if `has_audio` evaluates to `0`, though Timeline was probably already taking care of that.

To accommodate this fix, a new overload is added to `Frame.cpp`: `Frame::AddColor(const QColor&)`. The existing `Frame::AddColor(int w, int h, std::string color)` is rewritten as a thin wrapper around `Frame::AddColor(const QColor&)`, to avoid code duplication.

A `has_video` unit test is added for the Clip class, which verifies that the frame image is transparent when `has_video` is `0`.

I'm not **100%** sure this DTRT for all permutations of readers that output video-only, readers that output audio-only, etc. It could use more testing. I know it works for video+audio clips, rendering the video invisible when `has_video == 0`.

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -23,6 +23,8 @@
 #include "DummyReader.h"
 #include "Timeline.h"
 
+#include <Qt>
+
 using namespace openshot;
 
 // Init default settings for a clip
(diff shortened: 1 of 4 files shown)


jonoomph:
LGTM! Nice work on this one, and good catch!

PR #767, opened 2021-10-31
Avoid a bug in ImageMagick or cassert headers

troyrollo:
At least on Debian bullseye as at today, the ImageMagick headers
include &lt;cassert&gt; with an ImageMagick namespace current. <cassert>
and the headers it then pulls in also do not set the namespace.
Those headers then declare functions such as __assert_fail, which
end up in the ImageMagick namespace. This causes failures in other
parts of the build, such as the ZeroMQ code. This seems to be a
bug in either the ImageMagick headers or the &lt;cassert&gt; header, but
can be easily worked around by including &lt;cassert&gt; before any
ImageMagick header. This change does so.

PR #771, opened 2021-11-02
Move ImageMagick code out of Frame class

ferdnyc:
#767 got me thinking.

In that PR, building on @troyrollo 's fix for MagickUtilities, I replaced the ImageMagick headers in Frame.h with a forward declaration of `Magick::Image`, which is all it needed.

But that made the point that really, the _only reason_ there's any ImageMagick code in the Frame class is to convert between QImage and Magick::Image in both directions, when necessary. There's absolutely no reason that has to be done _by_ the Frame class. In fact, it makes a lot more sense if it isn't.

So, this PR adds a new `MagickUtilities.cpp` to go with `MagickUtilities.h`. It implements two functions, `Magick2QImage` and `QImage2Magick`. The ImageMagick readers/writers can use those to convert their images as needed, and Frame only has to deal with QImage.

As part of this change I also moved the `cleanUpQbuffer` function to `QtUtilities.h`, and removed the `qbuffer` member from `Frame`. There's no reason that needs to be there.

All of the existing unit tests pass with flying colors ,whether building with or without ImageMagick enabled. That's _probably_ a sign this is good to go, but those tests aren't as comprehensive as I'd really like so I may try to add some more before merging this. Primarily, I want to make sure I haven't introduced any memory leaks or corruptions.

@troyrollo I'd love your thoughts on this as well, if you've got a chance to look it over.


ferdnyc:
image


Huh! I guess the unit tests are a little more comprehensive than I'd realized, after all. Still, it would be better to get that to 100% at least for the diff.

src/ImageReader.cpp
```diff
@@ -100,17 +97,19 @@ std::shared_ptr<Frame> ImageReader::GetFrame(int64_t requested_frame)
 {
 	// Check for open reader (or throw exception)
 	if (!is_open)
-		throw ReaderClosed("The FFmpegReader is closed.  Call Open() before calling this method.", path);
+		throw ReaderClosed(
+            "The FFmpegReader is closed. "
+            "Call Open() before calling this method.", path);
```
ferdnyc: Ahhh, crap. Not only did I screw up the indentation here, but I failed to notice it's calling itself the _wrong_ class.

src/ImageReader.cpp
```diff
@@ -100,17 +97,19 @@ std::shared_ptr<Frame> ImageReader::GetFrame(int64_t requested_frame)
 {
 	// Check for open reader (or throw exception)
```
ferdnyc: And I have no idea why I left this inane comment.

src/ImageReader.cpp
```diff
@@ -100,17 +97,19 @@ std::shared_ptr<Frame> ImageReader::GetFrame(int64_t requested_frame)
...
-		throw ReaderClosed("The FFmpegReader is closed.  Call Open() before calling this method.", path);
+		throw ReaderClosed(
+            "The FFmpegReader is closed. "
+            "Call Open() before calling this method.", path);
 
 	// Create or get frame object
 	auto image_frame = std::make_shared<Frame>(
-		requested_frame, image->size().width(), image->size().height(),
+		requested_frame,
+        image->size().width(), image->size().height(),
```
ferdnyc: Indentation, again.

src/MagickUtilities.h
```diff
@@ -30,8 +33,8 @@
     // Determine ImageMagick version, as IM7 isn't fully
     // backwards compatible
     #ifndef NEW_MAGICK
-	   #define NEW_MAGICK (MagickLibVersion >= 0x700)
-	#endif
+    #define NEW_MAGICK (MagickLibVersion >= 0x700)
+    #endif
```
ferdnyc: Indents  here, too. Battin' 1000.

ferdnyc (commented):
Needs some source cleanup.

ferdnyc:
Hmm. From the CI failure, it seems I also missed a call to `Frame::cleanUpBuffer`. I wonder why only macOS noticed that?

ferdnyc:
> Hmm. From the CI failure, it seems I also missed a call to `Frame::cleanUpBuffer`. I wonder why only macOS noticed that?

Ah, OK. The issue was that I forgot to delete the declaration from Frame.h.

Stil weird that Linux was OK with that. But, fixed, as well as (I think) all the indentation issues and the wrongly-identified class.

ferdnyc:
Happened to spot a note in the Mac build of the Python bindings that `delete` was being called on `TimelineBase` despite it  not  having a virtual destructor. 

Base classes [must always have a virtual destructor](https://isocpp.org/wiki/faq/virtual-functions#virtual-dtors) if objects of type `DerivedClass` will ever be deleted by calling `delete *BaseClassPtr;`. The virtual destructor can be _defaulted_, but it has to be explicitly declared because [destructors are never virtual by default](https://isocpp.org/wiki/faq/virtual-functions#virtual-dtor-rationale). So, I added a `virtual ~TimelineBase() = default;`. Easy-peasy.

src/MagickUtilities.cpp
```diff
@@ -0,0 +1,73 @@
...
+    MagickCore::ExportImagePixels(
+        image->constImage(), 0, 0,
+        image->columns(), image->rows(),
+        "RGBA", Magick::CharPixel,
+        qbuffer, &exception);
+
+    auto qimage = std::make_shared<QImage>(
+        qbuffer, image->columns(), image->rows(),
+        image->columns() * BPP,
+        QImage::Format_RGBA8888_Premultiplied,
```
ferdnyc, line 67: I cannot figure out  why Codecov is calling out this line in particular as not covered. It's in the middle of a function call! And the whole file _should_ be covered now, since I added a "conversions" case to the ImageWriter tests.

jonoomph:
LGTM, but we have some conflicts on it. 👍

jonoomph (approved):

ferdnyc:
@jonoomph Resolved. I believe it's GTG, but I'll let the CI builds decide for sure.

ferdnyc:
Well, I was close. _One_ missing include.

jonoomph:
LGTM!

PR #772, opened 2021-11-03
CI: Try to avoid always recompiling Resvg

ferdnyc:
Currently the CI scripting is causing it to always recompile Resvg, when it's enabled. This PR attempts to fix that, so that Resvg won't be rebuilt on a cache hit.


ferdnyc:
Yay, that worked. Merging.

PR #774, opened 2021-11-09
Code changes to prepare for JUCE upgrade, speed up builds

ferdnyc:
This PR makes changes to the way libopenshot uses JUCE that prepare the codebase for the forthcoming upgrade to JUCE 6.1.2. Many of these are not strictly _necessary_, but they eliminate long-standing bad/outdated practices and/or inefficiencies that have become ingrained in how we use the various JUCE modules.

### Major changes

1. Replace all references to `juce::AudioSampleBuffer` with `juce::AudioBuffer<float>`. `juce::AudioSampleBuffer` has been a legacy typedef for `AudioBuffer<float>` for several years now.
2. Eliminate all use of `juce::CriticalSection` and `juce::GenericScopedLock`, replacing all `CriticalSections` with `std::recursive_mutex` and locking in the standard fashion (using `std::lock_guard`). Unlike `std::mutex` (which becomes a deadlock factory if it's used the way we use `CriticalSection`), `std::recursive_mutex` is explicitly designed (as was `CriticalSection`) to support multiple locks from the same thread  when code that uses it calls other code that uses it. Jules has even mentioned at times that he's considered deprecating `CriticalSection` in favor of `std::recursive_mutex` now that the latter exists, as they have identical semantics. 
    The primary benefit of this change is that it allows us to remove uses of the JUCE headers in classes that have no business including them, like `ZmqLogger`. The only reason `#include <OpenShotAudio.h>` appeared in `ZmqLogger.cpp` was for `juce::CriticalSection`, now it's been replaced with a much lighter-weight `#include <mutex>`.
3. Speaking of `OpenShotAudio.h`, all uses of `#include <OpenShotAudio.h>` have been eliminated from the code. Now, code that makes use of e.g. `juce::AudioBuffer<float>` will instead do the following:
    ```c++
    #include <AppConfig.h>
    #include <juce_audio_basics/juce_audio_basics.h>
    ```
    This avoids pulling in the _entirety_ of JUCE into every class that needs even just a simple audio buffer.
4. Class declarations that only make use of JUCE classes via pointers have further had the includes replaced with forward declarations, to lighten the compilation weight of those headers. (The same has also been increasingly done for `openshot::` classes as well, where appropriate. A _lot_ of `#include "Frame.h"` in headers is now `namespace openshot { class Frame; }`, for a significant reduction in header size and complexity.)

### Compilation time savings
Taken all together, the changes in this PR — even compiling with JUCE 5.4.7 — had a consistent impact on my local machine's runtime for the command `time cmake --build build_noccache --target pyopenshot` (from a cold `cmake -B build_noccache -S .`, into a formerly-empty directory and with `ccache` disabled). 

| branch          | CPU cycles                              | total runtime      |  speedup   |
|-----------------|---------------------------------------|----------------------|---------------|
| develop        | `641.65s user`<br/>`21.18s system` | `11:14.93 total` |                   |
| pre-juce612 | `437.21s user`<br/>`11.17s system` | `7:32.07 total`    | **33%**     |

### Other changes

In addition to the preceding, this PR contains (as separate commits):
1. An un-indenting of the entire `Enums.h` header by one level, to avoid wasting an indent on the namespace.
2. A fix for the new `ChromaKey.cpp` code so that it uses `std::`-prefixed `<cmath>` classes for math functions like `std::sin`, `std::cos`, `std::sqrt`, etc. 
     Turns out, the existing code was accidentally making use of the JUCE math functions. (Most obviously `juce::square()`, which has no corresponding STL version. The closest equivalent is `std::pow()`, which takes two arguments.) With the header changes, ChromaKey's namespace is no longer polluted with the JUCE classes,

### DIFF
--- a/bindings/python/openshot.i
+++ b/bindings/python/openshot.i
@@ -38,7 +38,7 @@
 #ifdef USE_IMAGEMAGICK
 	%shared_ptr(Magick::Image)
 #endif
-%shared_ptr(juce::AudioSampleBuffer)
+%shared_ptr(juce::AudioBuffer<float>)
 %shared_ptr(openshot::Frame)
 
 %{
--- a/bindings/ruby/openshot.i
+++ b/bindings/ruby/openshot.i
@@ -38,7 +38,7 @@
 #ifdef USE_IMAGEMAGICK
 	%shared_ptr(Magick::Image)
 #endif
-%shared_ptr(juce::AudioSampleBuffer)
+%shared_ptr(juce::AudioBuffer<float>)
 %shared_ptr(openshot::Frame)
 
 /* Template specializations */
--- a/src/AudioBufferSource.cpp
+++ b/src/AudioBufferSource.cpp
@@ -16,14 +16,14 @@ using namespace std;
 using namespace openshot;
 
 // Default constructor
-AudioBufferSource::AudioBufferSource(juce::AudioSampleBuffer *audio_buffer)
+AudioBufferSource::AudioBufferSource(juce::AudioBuffer<float> *audio_buffer)
 		: position(0), repeat(false), buffer(audio_buffer)
 { }
 
 // Destructor
 AudioBufferSource::~AudioBufferSource()
 {
-	// forget the AudioSampleBuffer. It still exists; we just don't know about it.
+	// forget the AudioBuffer<float>. It still exists; we just don't know about it.
 	buffer = NULL;
 }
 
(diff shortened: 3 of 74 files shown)


ferdnyc:
I'd say these numbers speak for themselves.

### Most recent `develop` branch build
image

### This branch
image

jonoomph (approved):
Thanks @ferdnyc! This looks great, and it all makes sense. Nice work cutting down on build times also!!! 💯

jonoomph:
Merging

PR #777, opened 2021-11-11
tests/Fraction: Approx() floating-point for 32bit

ferdnyc:
This should fix the failing unit tests on 32-bit Windows.


jonoomph:
@ferdnyc Awesome, that fixed the build servers! Thx

PR #778, opened 2021-11-11
(Second attempt) Switch from UseDoxygen to doxygen_add_docs

ferdnyc:
A re-submission of #613, which caused build failures due to a stupid mistake: I accidentally renamed the output dir from `doc` to `docs`. Fixed in this updated PR.

I also fixed a bunch of _other_ boneheaded mistakes I made (like excluding the `OpenShot.h` file that contains the documentation `\mainpage`!) — in my defense, it was a very old PR.

Also:
* Cleaned up the documentation comment issues in the `audio_effects` classes that Doxygen was complaining about (had been with the old `UseDoxygen` generator as well). There are still a bunch of undocumented parameters in most of the constructors, though, which I couldn't document because I honestly have no idea what they do. What's a `q_factor`??
* A big part of that cleanup was _deleting_ copy-pasted doc comments for `override` functions, because I learned that Doxygen has a lovely feature: If a function in a derived class has no documentation, the documentation from the base class implementation will _automatically_ be reused! So, there's no reason for us to be repeating all of that documentation in every derived class, we just have to document the base class implementations and we're set.
* Speaking of audio_effects, I renamed all of the constructors' parameters from <code>new_<var>foo</var></code> to just <code><var>foo</var></code>, while I was in there. Because a constructor with parameters named <code>new_<var>whatever</var></code> is just weird... there was never an **old** <code><var>whatever</var></code> value to be modifying!)
* Expanded the `PREDEFINED` symbols for Doxygen, to include more documentation even if certain libraries aren't compiled in on the system generating the docs.
* Fixed a couple of ambiguous function signatures Doxygen was whining about
* Took advantage of the fact that JUCE DSP forced us to the C++14 standard, to replace the previous deprecation annotations with C++14-standard `[[deprecated]]`, which Doxygen seems to handle better than `__attribute__ (((deprecated)))` (because... yeah).


jonoomph:
@ferdnyc This looks great! Thanks for the quick turn-around on it. Shall we merge? :crossed_fingers:

jonoomph:
Merging now

ferdnyc:
@jonoomph Sure, I'll give it a whirl. Hopefully nothing blows up the second time. It's pretty tested, but I did touch the audio_effects which have no unit tests, so... slightly perilous. But, hey, we can always re-re-revert. :roll_eyes:

ferdnyc:
@jonoomph So, that was more a rhetorical question, eh? :wink:

jonoomph:
Like you said, we can always re-re-re-revert, lol

jonoomph:
So far so good on the build servers!

ferdnyc:
Yeah, the only reason it failed the first time was because I stupidly changed the name of the output dir from `build/doc/html` to `build/docs/html` (unintentionally). But even if it'd loaded the files from `build/docs/html/`, they would've been hella broken. Mostly because of the excluded `OpenShot.h` file. I _think_ they'll come out OK now, tho.

PR #779, opened 2021-11-13
Disable FFmpeg hwaccel when APPIMAGE_BUILD is set

### LINKED ISSUE #4538: Linux AppImage: Possible issues importing video with certain GPUs
It appears the OpenShot 2.6.1 release AppImage (and all recent daily builds before/after that release) _may_ fail to import certain video formats on systems containing a GPU which has hardware decoding support for the video format.

The issue occurs when the user's existing preferences have a hardware-decoding mode selected. (For instance, from previously running an installed native build of OpenShot.) This setting is not properly ignored by the AppImage builds of OpenShot, which do not currently support hardware video decoding. This results in a valid video file being rejected by OpenShot when importing:

image

As a workaround, quit all instances of OpenShot, then delete or rename the file `$HOME/.openshot_qt/openshot.settings` to reset your OpenShot user preferences. The next time OpenShot is launched, that file will be created with default settings, and the AppImage build will successfully import the affected files.

ferdnyc:
As discussed in OpenShot/openshot-qt#4538, the hwaccel code in FFmpegReader/Writer should be completely disabled, when building for the Linux AppImages. This PR respects the CMake option 'APPIMAGE_BUILD', which we are already setting in the project server builds, and defines the symbol `USE_HW_ACCEL=0` for the build. The `#ifdef`s around the hwaccel code will cause it to be omitted from the compiled classes entirely.

Fixes OpenShot/openshot-qt#4538

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -359,7 +359,10 @@ add_feature_info("FFmpeg ${resample_lib}" TRUE "Audio resampling uses ${resample
 set(FFMPEG_USE_SWRESAMPLE ${USE_SW} CACHE BOOL "libswresample used for audio resampling" FORCE)
 mark_as_advanced(FFMPEG_USE_SWRESAMPLE)
 
-# Version check for hardware-acceleration code
+# Version check / enablement for hardware-acceleration code
+if(APPIMAGE_BUILD)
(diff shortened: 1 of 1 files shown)


JacksonRG:
LGTM!

PR #781, opened 2021-11-18
Resolve a build failure for missing <memory> include

jonoomph:
Resolve a build failure for missing <memory> include. This is needed for `shared_ptr` to build on certain compilers. My local system was failing to build (including our Cloud API template servers).  No idea why this doesn't fail on our GitLab or GitHub builders though... so I'm assuming this is somehow related to the version of our compilers.

```
error: 'shared_ptr' was not declared in this scope
```

### DIFF
--- a/src/WriterBase.cpp
+++ b/src/WriterBase.cpp
@@ -12,6 +12,7 @@
 
 #include <iostream>
 #include <iomanip>
+#include <memory>
 
 #include "WriterBase.h"
 #include "Exceptions.h"


jonoomph:
Merging this quickly though... since I need to get some builds back online

JacksonRG:
I actually was rebuilding my libopenshot yesterday, and running into issues. I assumed [PEBKAC](https://www.computerhope.com/jargon/p/pebkac.htm). But I just pulled and did the same thing, and make installed successfully. :+1:

ferdnyc:
Apologize that I hadn't seen this untiil now — my only comment would've probably been that, while I'm _normally_ a big proponent of putting `#include`s in the _implementation_ files where they're needed, in **this particular** case there's no actual use of `std::shared_ptr` **anywhere** in `WriterBase.cpp`. It's **only** used in `WriterBase.h`. So, arguably that would be the place to `#include <memory>`. But if it works this way too, doesn't really make much difference to me.

Most likely, until my latest cleanup `WriterBase.h` had a ton of _other_ OpenShot `#include`s in it, one of which surely contained `#include <memory>` since most do. I must've missed that I'd removed all of the indirect includes.

(Odd that it didn't show up in the CI though. It seems like some systems/compilers make `<memory>` a default include — `<string>` too, for that matter — while others need it explicitly included.)

PR #784, opened 2021-12-10
Replace av_init_packet in FFmpeg 3.2+

ferdnyc:
`av_init_packet` is now deprecated in the newest FFmpeg, mostly because we shouldn't be using it anyway. In all FFmpeg versions since 3.0, we've had `av_packet_alloc()` instead, which is the preferred method of initializing a packet pointer.

Note that I said packet _pointer_, not a packet — making this change required changing all `pkt` instances from `AVPacket` to `AVPacket*`, and adding the necessary pointer dereferences to all uses. (Or, as often as not, changing a library call to _stop_ taking the address of the locally-allocated struct, and instead just pass it the pointer.)

FFmpeg 2.8 — which we still depressingly support (in theory) — doesn't have an `av_packet_alloc()`, so its use is sequestered behind an `#if IS_FFMPEG_3_2` check. If that fails the old `av_init_packet()` method is used instead, but it now initializes an `AVPacket*` instead. Hopefully that will work; if not, we don't have any CI that checks FFmpeg < 3.4 anymore, so I guess we'll find out if an Ubuntu 16.04 user complains. Of course, any remaining Ubuntu 16.04 users are running an unsupported system that's past its End-of-Life date, so I'm not really worried.

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -828,17 +828,17 @@ void FFmpegWriter::flush_encoders() {
 #endif
 
 	// FLUSH VIDEO ENCODER
-	if (info.has_video)
+	if (info.has_video) {
 		for (;;) {
 
 			// Increment PTS (in frames and scaled to the codec's timebase)
-            video_timestamp += av_rescale_q(1, av_make_q(info.fps.den, info.fps.num), video_codec_ctx->time_base);
+			video_timestamp += av_rescale_q(1, av_make_q(info.fps.den, info.fps.num), video_codec_ctx->time_base);
 
 #if IS_FFMPEG_3_2
-            AVPacket* pkt = av_packet_alloc();
+			AVPacket* pkt = av_packet_alloc();
 #else
-            AVPacket* pkt;
-            av_init_packet(pkt);
+			AVPacket* pkt;
+			av_init_packet(pkt);
 #endif
 			pkt->data = NULL;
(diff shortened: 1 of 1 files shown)


ferdnyc:
> FFmpeg 2.8 — which we still depressingly support (in theory) — doesn't have an `av_packet_alloc()`, so its use is sequestered behind an `#if IS_FFMPEG_3_2` check. If that fails the old `av_init_packet()` method is used instead, but it now initializes an `AVPacket*` instead. Hopefully that will work; if not, we don't have any CI that checks FFmpeg < 3.4 anymore, so I guess we'll find out if an Ubuntu 16.04 user complains. Of course, any remaining Ubuntu 16.04 users are running an unsupported system that's past its End-of-Life date, so I'm not really worried.

(This is my passive-aggressive way of saying, "We REALLY need to get all the crap FFmpeg 2.x code out of the library".) It's a whole lot of ugly code that may or may not still _work_ — we  have no idea, we haven't tested it in years.

FFmpeg 3.0 ushered in the era of the packet-centric pipeline, which radically changed the flow of code written with the library. Changes since then have been relatively minor, mostly deprecating or swapping around individual function calls in ways that don't change the program flow. But trying to keep FFmpeg 2.x and 3.x+ code together in the same codebase means _loooong_ stretches of `#ifdef`'d code that do completely different things, while trying to pretend they're the same.

PR #790, opened 2021-12-28
Simplify frame timing corrections

JacksonRG:
We've been getting a few issues regarding the video, and audio not matching up. (in one case the video paused, and then jumped forward several seconds)

@jonoomph Suggested this comes from our timing correction code, which had multiple levels of response depending on how out of sync the audio and video became. The goal is a single approach which "rubber-bands" the video playback to the audio. Speeding up when behind and slowing down when ahead.

My approach is this: have the video preview thread sleep for
equation
additional seconds where
equation

The reason I'm cubing the error, is so that the sign `+/-` is preserved. And Ideally, the correction is smaller the closer we are to synched.

For this to work, I believe I need to divide before cubing like this:
equation^3)

Which will take advantage of how cubing fractions should prevent us from over-correcting.

### Testing:
If you feel like testing it, I used this [video](https://www.youtube.com/watch?v=ucZl6vQ_8Uo) (via youtube-dl) to test audio-video synch.

### DIFF
--- a/src/Qt/PlayerPrivate.cpp
+++ b/src/Qt/PlayerPrivate.cpp
@@ -16,12 +16,11 @@
 #include "ZmqLogger.h"
 
 #include <thread>    // for std::this_thread::sleep_for
-#include <chrono>    // for std::chrono microseconds, high_resolution_clock
-#include <cmath>
+#include <chrono>    // for std::chrono milliseconds, high_resolution_clock
 
 namespace openshot
 {
-    const auto max_sleep = std::chrono::seconds(1);
+    int close_to_sync = 5;
     // Constructor
     PlayerPrivate::PlayerPrivate(openshot::RendererBase *rb)
     : renderer(rb), Thread("player"), video_position(1), audio_position(0)
@@ -62,9 +61,16 @@ namespace openshot
         using double_micro_sec = std::chrono::duration<double, micro_sec::period>;
 
         while (!threadShouldExit()) {
+            // Get the start time (to track how long a frame takes to render)
             const auto time1 = std::chrono::high_resolution_clock::now();
+            // Calculate on-screen time for a single frame in microseconds
             const auto frame_duration = double_micro_sec(1000000.0 / reader->info.fps.ToDouble());
+            const auto max_sleep = frame_duration * 1.5;
+
+            // Get the current video frame (if it's different)
             frame = getFrame();
+
+            // Experimental Pausing Code (if frame has not changed)
             if ((speed == 0 && video_position == last_video_position)
                 || (video_position > reader->info.video_length)
                ) {
@@ -72,9 +78,12 @@ namespace openshot
                 std::this_thread::sleep_for(frame_duration);
                 continue;
             }
-            //Display the frame
+
+            // Set the video frame on the video thread and render frame
             videoPlayback->frame = frame;
             videoPlayback->render.signal();
+
+            // Keep track of the last displayed frame
             last_video_position = video_position;
 
             // How many frames ahead or behind is the video thread?
@@ -89,30 +98,28 @@ namespace openshot
                 video_frame_diff = video_position - audio_position;
             }
 
-            // how far ahead or behind (negative) the video is
-            // relative to the audio
-            // (Cubed so that the higher the error, the more extreme the correction
-            auto correction = micro_sec((long)frame_duration.count() * (long)powf(video_frame_diff,3));
-
-            // If we're more than a whole frame behind, skip frames
-            int frames_behind = correction.count() / frame_duration.count();
-            if (frames_behind > 0) {
-                videoPlayback->frame->number += frames_behind;
-            }
-
-            /* SLEEP CALCULATIONS */
+            // Get the end time (to track how long a frame takes to render)
             const auto time2 = std::chrono::high_resolution_clock::now();
-            const auto render_time = double_micro_sec(time2-time1);
-            const auto sleep_time = duration_cast<micro_sec>(frame_duration - render_time) + (correction);
 
-            // Debug
+            // Determine how many microseconds it took to render the frame
+            const auto render_time = double_micro_sec(time2 - time1);
+
+            // Calculate the amount of time to sleep (by subtracting the render time)
+            auto sleep_time = duration_cast<micro_sec>(frame_duration - render_time);
+            // Raising video_frame_diff to the power of 3
+                // Preserves the sign (+/-) and corrects more aggressively depending on how
+                // out of sync the video is.
+            sleep_time = sleep_time + std::chrono::microseconds((long(powf(100 * video_frame_diff, 3) )));
+
             ZmqLogger::Instance()->AppendDebugMethod("PlayerPrivate::run (determine sleep)", "video_frame_diff", video_frame_diff, "video_position", video_position, "audio_position", audio_position, "speed", speed, "render_time(ms)", render_time.count(), "sleep_time(ms)", sleep_time.count());
 
-            if (sleep_time.count() > 0) {
-                if (sleep_time > max_sleep ) {
-                    std::this_thread::sleep_for(max_sleep);
-                } else {
+            // Sleep if sleep_time is greater than zero after correction
+            // (leaving the video frame on the screen for the correct amount of time)
+            if (sleep_time > sleep_time.zero() ) {
+                if (sleep_time < max_sleep) {
                     std::this_thread::sleep_for(sleep_time);
+                } else {
+                    std::this_thread::sleep_for(max_sleep);
                 }
             }
         }
--- a/src/Qt/PlayerPrivate.cpp
+++ b/src/Qt/PlayerPrivate.cpp
@@ -16,10 +16,11 @@
 #include "ZmqLogger.h"
 
 #include <thread>    // for std::this_thread::sleep_for
-#include <chrono>    // for std::chrono milliseconds, high_resolution_clock
+#include <chrono>    // for std::chrono microseconds, high_resolution_clock
 
 namespace openshot
 {
+    int skipFrames = 0;
     // Constructor
     PlayerPrivate::PlayerPrivate(openshot::RendererBase *rb)
     : renderer(rb), Thread("player"), video_position(1), audio_position(0)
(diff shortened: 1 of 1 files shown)


src/Qt/PlayerPrivate.cpp
```diff
@@ -109,15 +109,14 @@ namespace openshot
             const auto render_time = double_micro_sec(time2 - time1);
 
             // Calculate the amount of time to sleep (by subtracting the render time)
-//            auto sleep_time = duration_cast<micro_sec>( (frame_duration - render_time));
-            auto correction_factor = video_frame_diff * video_frame_diff * video_frame_diff * video_frame_diff * video_frame_diff;
-//            auto sleep_time_no_correction = duration_cast<micro_sec>((frame_duration - render_time));
+            auto correction_factor = video_frame_diff * video_frame_diff * video_frame_diff;
```
jonoomph: Can we use `std::powf` instead? https://en.cppreference.com/w/cpp/numeric/math/pow

src/Qt/PlayerPrivate.cpp
```diff
@@ -109,15 +109,14 @@ namespace openshot
...
             auto sleep_micro_seconds = (frame_duration - render_time) + double_micro_sec (correction_factor);
             auto sleep_time = (sleep_micro_seconds.count() > 0) ? duration_cast<micro_sec>(double_micro_sec(sleep_micro_seconds)) : duration_cast<micro_sec>(double_micro_sec(0.0));
-            if (- (sleep_micro_seconds) > render_time) {
-                skipFrames = floor((-sleep_micro_seconds)/render_time);
-                if (skipFrames * 1.0 > reader->info.fps.ToDouble())
-                    skipFrames = floor(reader->info.fps.ToDouble());
+            if ( - (sleep_micro_seconds.count()) > frame_duration.count()) {
```
jonoomph: Let's add some comments to this, to explain what is happening. Also, let's remove the extra `()` and spaces between them.

src/Qt/PlayerPrivate.cpp
```diff
@@ -109,15 +109,14 @@ namespace openshot
...
+                skipFrames = floor( (-sleep_micro_seconds.count()) / frame_duration.count());
+                skipFrames = (skipFrames / 2) + 1;
+                skipFrames = (float)skipFrames > reader->info.fps.ToDouble() ? reader->info.fps.ToDouble() : skipFrames;
```
jonoomph: Should we be casting these double values to float? Should we be rounding to an INT?

src/Qt/PlayerPrivate.cpp
```diff
@@ -55,12 +56,16 @@ namespace openshot
...
         using std::chrono::duration_cast;
 
-        // Types for storing time durations in whole and fractional milliseconds
+        // Types for storing time durations in whole and fractional microseconds
         using micro_sec = std::chrono::microseconds;
         using double_micro_sec = std::chrono::duration<double, micro_sec::period>;
 
         while (!threadShouldExit()) {
-            // Calculate on-screen time for a single frame in milliseconds
+            if (skipFrames > 0) {
```
jonoomph: I don't think this actually does anything. It just quickly subtracts whatever you put in this variable back to 0 (almost instantaneously). It does not call `getFrame()` or increment the video position.

src/Qt/PlayerPrivate.cpp
jonoomph: Also, this logic looks very suspicious to me. Why would negative `sleep_micro_seconds` ever be larger than `frame_duration`, which I assume is always positive?

JacksonRG: Right. I should definitely explain... The idea is if we want to sleep a negative amount of time. Skip `(-sleep_time)/frame_duration` frames

I'll make sure that's clearly explained

src/Qt/PlayerPrivate.cpp
JacksonRG: Updated, so it should be updating the frame number, as a way to jump ahead.

ferdnyc:
@jonoomph @JacksonRG 

I  don't know if anyone is interested in my take on this, but here it is.

> We've been getting a few issues regarding the video, and audio not matching up. (in one case the video paused, and then jumped forward several seconds)
> 
> @jonoomph Suggested this comes from our timing correction code, 

It definitely does.

> My approach is this: have the video preview thread sleep for [...]

Let me stop you right there.

IMHO, the entire problem is that we have multiple threads, running without synchronization, relying on mutually sleeping in order to sync up their timing.

That is never going to work. It's not a question of adding additional complexity to the logic.

I'm almost done with the JUCE 6.1.4 update to libopenshot-audio (actually, I _am_ done (ferdnyc/libopenshot-audio branch `juce614` [passes its CI builds](https://github.com/ferdnyc/libopenshot-audio/actions/runs/1642208616)), but I have some changes to make to make here first) — regardless, even the current code works exactly the same.

Take a look at how JUCE's [`AudioTransportSource`](https://docs.juce.com/develop/classAudioTransportSource.html) (which plays back from a [`PositionableAudioSource`](https://docs.juce.com/develop/classPositionableAudioSource.html)) and [`BufferingAudioSource`](https://docs.juce.com/develop/classBufferingAudioSource.html) (which buffers a `PositionableAudioSource`) are constructed.

They both take a [`TimeSliceThread`](https://docs.juce.com/develop/classTimeSliceThread.html) as a constructor argument. And if you look in their code, both `juce_audio_basics/sources/juce_BufferingAudioSource.cpp` and `juce_audio_devices/sources/juce_AudioTransportSource.cpp` contain exactly **one** call to `Thread::Sleep()` — with a  hardcoded duration, no less — and in both cases it's only when they're _not_ playing audio. (AudioTransportSource sleeps for 2ms after stopping playback, BufferingAudioSource sleeps for 5ms on a loop in `prepareToPlay` while its buffer is filling.)

The reason they work like that is that all of JUCE's timing code is event-driven. For playback, it's controlled by that `TimeSliceThread` (TST), which acts as a scheduler. The sources register themselves with the TST when they're constructed. During playback, the TST — which holds a queue of registered classes — handles the job of balancing the other threads and deciding which ones need to be called next. The threads use [`WaitableEvent`](https://docs.juce.com/develop/classWaitableEvent.html)s to signal when they're ready to be called, and when they're holding for data from other threads.

> A thread can call `WaitableEvent::wait()` to suspend the calling thread until another thread wakes it up by calling the `WaitableEvent::signal()` method.

I'm not saying we should be using `juce::TimeSliceThread` for scheduling, per se. In fact, I don't think we should. But having all of these threads operate in isolation, trying to keep in sync by clock-watching, is IMHO never going to work. Or if it does, it will have become too heavy to run in real-time on all but the most overpowered systems out there. 

Synchronization is not something that can happen by accident, it _requires_ coordination between the moving parts.

ferdnyc:
> I'm not saying we should be using `juce::TimeSliceThread` for scheduling, per se. In fact, I don't think we should.

We ARE right now, though, effectively. I should point that out. currently all of our playback classes are subclasses of `juce::Thread`. We're just using their scheduling completely _wrong_.

All of our playback classes also purport to be `Qt` classes (even though most of them aren't) — it would be worth looking at how QtMultimedia does synchronization.

Using Qt signals and slots, instead of `juce::WaitableEvents`, **miiiight** be workable... I'm honestly not sure how they work for performance-critical timing. I know they do have both queued and direct connection modes, and that slots are supposed to automatically compact duplicated signals (either on queued connections, or all connections?) when the system gets bogged down.

(Which is why it's a bit silly that we make such a big deal out of delaying responses to GUI signals, in the OpenShot Python code. They're _built_ to handle storms of repeated events, and the only reason it's a problem for us is that **our** code _isn't_ built to handle it.)

So, using Qt signals to synchronize playback might be enough to solve the timing issues all by itself.

To some extent, if we're using JUCE for audio output we _have_ to use `TimeSliceThread`, or at least our own scheduler may have to implement that interface in order to manage the JUCE parts of the system. But it shouldn't be that difficult to implement. Here are the `cloc` (Count Lines Of Code) results for the JUCE thread classes:

```sh
$ cloc --by-file juce_Thread.* juce_TimeSliceThread.* juce_WaitableEvent.*
       6 text files.
       6 unique files.                              
       0 files ignored.

github.com/AlDanial/cloc v 1.90  T=0.01 s (456.0 files/s, 110722.4 lines/s)
--------------------------------------------------------------------------------------
File                                    blank        comment           code
--------------------------------------------------------------------------------------
juce_Thread.cpp                           119             50            370
juce_TimeSliceThread.cpp                   38             20            126
juce_Thread.h                             105            237             73
juce_WaitableEvent.cpp                     15             16             39
juce_TimeSliceThread.h                     37             79             34
juce_WaitableEvent.h                       26             56             17
--------------------------------------------------------------------------------------
SUM:                                      340            458            659
--------------------------------------------------------------------------------------
```

If we're doing much more than that, we're overcomplicating things.

ferdnyc:
OK, last few points and then I've said my piece.

1. A basic principle of A/V playback: From the user perspective, it is much, MUCH better to delay the start of playback until it can be done smoothly, and to insert the occasional long pause during playback to let the system catch up, than to charge ahead and exhibit constant stuttering or jerky performance throughout. The latter can make the video basically unwatchable, but "buffering..." pauses can be tolerated.

1. Playback threads should be created when playback starts, and destroyed when it stops. Possibly even when it _pauses_. Having long-running threads that sit there and furiously sleep until they're needed is a sign of something wrong.

    Cheap, lightweight threads are a fundamental requirement of all modern computing platforms, and making good use of them has a lot of advantages. For starters, if your thread is task-oriented then you don't have to spend lots of time re-computing parameters. What's the output resolution of my video, what's the frame rate, etc? You compute that _ONCE_ in the constructor and then you never have to worry about it again, because if it changes the thread will be destroyed and replaced with a new one. That allows the thread to be far more efficient.

1. The code to pull up and show a single frame in the preview and the code to display the video as it plays may need to be completely separate code, even though they result in the same image. Seeking playback to a certain point and pausing it is a hellishly expensive way to display a video frame. And doing the audio processing for that frame, when it won't be used for anything until the user hits "Play" (which they may never do, for that frame), is just wasted time. It may be better to hold off on all of that, and instead have to deal with a half-second delay when the "Play" button is pressed while the system spins up the playback machinery.

1. It's important to remember that in Qt, as in most other systems, any graphics code **has** to run on the main loop. That includes the code that manages the video preview window in OpenShot. We can pretend to be running video on another thread all we like, but in the end it's not really true. The **main** thread has to be signaled to actually display the image, and playback performance is inextricably linked to the performance of the `VideoWidget` class in Python. It _may_ be possible to mitigate that some by directly using an OpenGL surface to display the video.

    (Right now, Qt will use OpenGL if possible, but it's still accelerating raster display of an RGB `QImage`, not of a YUV video surface. Ideally we'd generate the video directly in the GPU memory and never pass it back to the code. But that will have a major impact on things like the transform mode.)

JacksonRG:
@ferdnyc I like the idea of lightweight threads. And scrubbing the playhead back and forth could probably be a lot smoother if we have a quick "display that frame" function.

But I think that will be my next PR on this topic. I learned a great deal about the cache, and have some ideas to improve performance. The critical blocker is the fact that audio and video are out of sync.

 To the best of my knowledge, this synchronizes them, and performance will be the same until we give that some focus.

PR #792, opened 2022-01-10
Rename GetAudioSample to GetPeakSampleLevel, and fix... everything about it

ferdnyc:
While working on audio code updates for  JUCE 6, I discovered that `Frame::GetAudioSample`...

1. Was wildly misnamed, relative to its function
1. Was _confusingly_ named, since `Frame::GetAudioSamples` does something completely different
1. Had a confusingly-named set of largely-undocumented parameters. The behavior of `channel == -1` was mentioned, but no mention was made of what `magnitude_range` is or how to specify it. (I renamed it `sample_count`, which hopefully _does_ tell you what it actually signifies. But just in case, it's also got a `@param` line in the doc comment that explains in more detail.)
1. Was documented to do something completely different than what it _actually_ does (or, rather, what the JUCE function underlying it actually does).
1. Had a bug when passed a `channel` parameter of `0`, and would return the wrong values. (It was supposed to take `channel = -1` to mean all channels, but the test in the code was written `if (channel > 0) { <do channel-specific stuff> }`, meaning that when `channel == 0` it would take the other branch and act as if `channel == -1`.) <sup>1</sup>
1. Is never actually called anywhere in the libopenshot code. (It is used by OpenShot, to graph audio waveforms.) <sup>2</sup>

Details on the fixes for these issues forthcoming, as inline review comments.

#### Notes
1. The implication of this is that OpenShot's waveforms are currently incorrect for a lone/separated channel 0. Based on the existing code, they will always be identical to the combined `channel == -1` waveforms. The corrected code will OpenShot to correctly display the audio waveform for _only_ the first channel.

2. libopenshot's own waveform code is entirely unlike OpenShot's. libopenshot uses a completely different (and incredibly inefficient) method involving direct examination of the audio buffer's individual sample values. That should probably be improved.

### DIFF
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -298,7 +298,7 @@ void Frame::DisplayWaveform()
 }
 
 // Get highest sample value in range
-float Frame::GetPeakLevel(int channel, int sample, int sample_count)
+float Frame::GetPeakSampleLevel(int channel, int sample, int sample_count)
 {
 	if (channel < 0) {
 		// Return magnitude across all channels
@@ -309,7 +309,7 @@ float Frame::GetPeakLevel(int channel, int sample, int sample_count)
 
 // Alias (Deprecated)
 float Frame::GetAudioSample(int channel, int sample, int sample_count) {
-    return GetPeakLevel(channel, sample, sample_count);
+    return GetPeakSampleLevel(channel, sample, sample_count);
 }
 
 // Get an array of sample data
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -297,17 +297,19 @@ void Frame::DisplayWaveform()
 	ClearWaveform();
 }
 
-// Get magnitude of range of samples (if channel is -1, return average of all channels for that sample)
-float Frame::GetAudioSample(int channel, int sample, int magnitude_range)
+// Get highest sample value in range
+float Frame::GetPeakLevel(int channel, int sample, int sample_count)
 {
-	if (channel > 0) {
-		// return average magnitude for a specific channel/sample range
-		return audio->getMagnitude(channel, sample, magnitude_range);
-
-	} else {
-		// Return average magnitude for all channels
-		return audio->getMagnitude(sample, magnitude_range);
+	if (channel < 0) {
+		// Return magnitude across all channels
+		return audio->getMagnitude(sample, sample_count);
 	}
+	return audio->getMagnitude(channel, sample, sample_count);
+}
+
+// Alias (Deprecated)
+float Frame::GetAudioSample(int channel, int sample, int sample_count) {
+    return GetPeakLevel(channel, sample, sample_count);
 }
 
 // Get an array of sample data
--- a/src/Frame.h
+++ b/src/Frame.h
@@ -183,8 +183,17 @@ namespace openshot
 		/// Display the wave form
 		void DisplayWaveform();
 
-		/// Get magnitude of range of samples (if channel is -1, return average of all channels for that sample)
-		float GetAudioSample(int channel, int sample, int magnitude_range);
+		/// Get highest value within a range of samples
+		/// 
+		/// @param channel Channel number to scan, or -1 for all channels
+		/// @param sample The starting sample for the range
+		/// @param sample_count The length of the sample range
+		/// @returns The maximum sample value within the range of samples on
+		///          the specified channel, or all channels if channel == -1
+		float GetPeakLevel(int channel, int sample, int sample_count);
+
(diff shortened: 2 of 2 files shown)


src/Frame.h
```diff
@@ -183,8 +183,17 @@ namespace openshot
...
-		/// Get magnitude of range of samples (if channel is -1, return average of all channels for that sample)
-		float GetAudioSample(int channel, int sample, int magnitude_range);
+		/// Get highest value within a range of samples
+		/// 
+		/// @param channel Channel number to scan, or -1 for all channels
+		/// @param sample The starting sample for the range
+		/// @param sample_count The length of the sample range
+		/// @returns The maximum sample value within the range of samples on
+		///          the specified channel, or all channels if channel == -1
+		float GetPeakLevel(int channel, int sample, int sample_count);
```
ferdnyc: I started by renaming the method to `Frame::GetPeakLevel`. (Though I'm considering a change to `GetPeakAudioLevel` or `GetPeakSampleLevel`, now that I remember this is in `Frame` and not an audio class.) The new name better represents both its real purpose and its actual function. Its parameters are renamed, as mentioned, from `(channel, sample, magnitude_range)` to `(channel, sample, sample_count)`.

src/Frame.h
```diff
@@ -183,8 +183,17 @@ namespace openshot
...
 
-		/// Get magnitude of range of samples (if channel is -1, return average of all channels for that sample)
-		float GetAudioSample(int channel, int sample, int magnitude_range);
+		/// Get highest value within a range of samples
+		/// 
+		/// @param channel Channel number to scan, or -1 for all channels
+		/// @param sample The starting sample for the range
+		/// @param sample_count The length of the sample range
+		/// @returns The maximum sample value within the range of samples on
+		///          the specified channel, or all channels if channel == -1
```
ferdnyc: I removed all mention of "average"-anything from the doc comment, because no averaging is _ever_ performed by `juce::AudioBuffer::getMagnitude()`. What it _actually_ does is return the largest sample value from within a range of samples. If a channel is specified, it's from among the samples on that channel. If a channel value isn't specified (or is `-1`, in our case) it's from among the samples on _all_ channels in that range. But the value returned is still the **absolute** maximum, without averaging.

ferdnyc: Each parameter is explicitly documented, as is the return value.

src/Frame.cpp
```diff
@@ -297,17 +297,19 @@ void Frame::DisplayWaveform()
...
-	if (channel > 0) {
-		// return average magnitude for a specific channel/sample range
-		return audio->getMagnitude(channel, sample, magnitude_range);
-
-	} else {
-		// Return average magnitude for all channels
-		return audio->getMagnitude(sample, magnitude_range);
+	if (channel < 0) {
+		// Return magnitude across all channels
+		return audio->getMagnitude(sample, sample_count);
```
ferdnyc, line 305: This fixes the channel-test bug, by flipping the branches along with the test.

src/Frame.h
```diff
@@ -183,8 +183,17 @@ namespace openshot
...
+		float GetPeakLevel(int channel, int sample, int sample_count);
+
+		/// Deprecated alias for GetPeakLevel
+		float GetAudioSample(int channel, int sample, int sample_count);
```
ferdnyc: I left `Frame::GetAudioSample()` as a deprecated alias to avoid breaking OpenShot in the short term. Its parameter names are changed to match `GetPeakLevel()`, but that's OK because OpenShot only passes its arguments positionally, never by name.

src/Frame.h
ferdnyc: (I did pull the trigger on renaming the method to `GetPeakSampleLevel`, which is why the code from this comment is now Outdated.)

src/Frame.h
```diff
@@ -183,10 +187,27 @@ namespace openshot
 		/// Display the wave form
 		void DisplayWaveform();
 
-		/// Get magnitude of range of samples (if channel is -1, return average of all channels for that sample)
-		float GetAudioSample(int channel, int sample, int magnitude_range);
+		/// Get highest value within a range of samples
+		//
+		/// @since 0.2.8
```
ferdnyc, line 192: This PR also marks our first 'since' annotation in the docs. 'since' is a handy tool for API authors to let consumers know when something was introduced to the library, so that they don't have to _guess_ whether they're using code that's not compatible with older versions.

I browbeat the CMake developers into adding since annotations to the CMake docs (and submitted PRs to add plenty of them myself), and it's made a world of difference in the usability of both the docs, and CMake's config language itself.

PR #793, opened 2022-01-12
Reformat ZmqLogger calls

ferdnyc:
I have submitted _parts_ of this in more PRs than I care to count, but most of them ended up abandoned or not merged. So, I have learned my lesson.

This PR does ONE THING<sup>1</sup> only.

It reformats nearly all of the `ZmqLogger` calls throughout the codebase — many, single lines of code in excess of 300, 400 characters long (!!) — to follow a consistent, well-defined, multi-line format.

```c++
// Except for all but the shortest of log calls, which I left alone...
ZmqLogger::Instance()->AppendDebugMethod("Short message");

// And with only one or two other, well-considered exceptions
// (which I'll get to), the standard format is this:
ZmqLogger::Instance()->AppendDebugMethod(
    "Initial message",
    "label", variable,
    "label", variable,
    "label", variable,
    "label", variable);

// The only two exceptions are:
// (1): Calls with especially long labels/variables, which became this...
ZmqLogger::Instance()->AppendDebugMethod(
    "Initial message",
    "label", variable,
    "Unnecessarily expository label that's just going to clutter the logs",
         usually_an_insanely_long->variable_name + math + to_go_with_it,
    "label", variable,
    "label", variable);
// (2): Intrinsically paired variables, like (direct from Clip.cpp):
ZmqLogger::Instance()->AppendDebugMethod(
    "Clip::get_transform (Build QTransform - if needed)",
    "frame->number", frame->number,
    "x", x, "y", y,
    "r", r,
    "sx", sx, "sy", sy);
```

I have tried to match the prevailing indentation of the surrounding function as much as possible.
* Where it was spaces, I used (four) spaces. 
* Where it was tabs, I closed my eyes, thought of England, and used tabs. 
* Where it was a complete disaster _**already**_, I reformatted the entire function to 4-space indents and never looked back.
* Where it was **wrong** (like, some misleading tabs or spaces thrown in that didn't line up unless your editor settings matched the author's), I fixed it. (But only if there was a logging call there I was changing anyway.)

And that's all<sup>1</sup>.

#### Notes
1. Only a very slight exaggeration, It also moves around the FFmpegWriter includes so that the `#include "ZmqLogger.h"` call (and any other source-only includes) happens in the implementation file, instead of the header. I thought I'd already done that for all of the classes, but I guess I missed that one.

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -699,7 +699,9 @@ std::shared_ptr<Frame> Clip::GetOrCreateFrame(int64_t number)
 {
 	try {
 		// Debug output
-		ZmqLogger::Instance()->AppendDebugMethod("Clip::GetOrCreateFrame (from reader)", "number", number);
+		ZmqLogger::Instance()->AppendDebugMethod(
+			"Clip::GetOrCreateFrame (from reader)",
+			"number", number);
 
 		// Attempt to get a frame (but this could fail if a reader has just been closed)
 		auto reader_frame = reader->GetFrame(number);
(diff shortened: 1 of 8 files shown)


ferdnyc:
Merging, because this (a) touches a lot of code, and will be painful if it gets out of sync, and (b) is purely code-cosmetic, other than cleaning up the FFmpegWriter includes.

PR #796, opened 2022-01-14
Caching & Playback Timing Improvements

jonoomph:
This PR is a huge refactor related to video and audio **playback** in OpenShot. Details below.

**Improvements and Bug Fixes in this PR:**
 - Replaced very bad video timing code with a much **improved video timing code**. We now track the "time" playback starts, and we can accurately predict the "End time" of any frame # (`start_time + (frame_counter_since_start * frame_duration)`. So, instead of the video thread keeping a "relative" sleep per loop with no way of accurately synchronizing with the audio thread, we are always taking the current time minus the calculated end time, for a 100% perfect video timing. This is the solution I have been searching for. Seems obvious now, lol.
 - **Refactored video cache**, which now works much more reliably than before. It now caches in both directions (forward in time, and backwards), filling in missing frames when detected. It now only uses 1/2 the cache size, so 50% behind the playhead & 50% in front of the playhead. This allows users to easily scrub forward and backwards in the cache, without needing to re-generate frames. 
 - **Fixed a bad audio buffer** design, where we had 2 buffers for no reason. This fixes a bug where audio would stay in one of our audio buffers, and not empty out when seeking to a new position (i.e. playing a tiny bit of sound from the last position).
 - **Lower audio latency** (sending 500 samples per callback now, usually around 88 - 96 callbacks per second). No more pre-buffering in JUCE.
 - **Pre-roll support** before playback. Instead of instantly starting playback, we always wait for our pre-roll to cache a hand-full of frames before playback begins.
 - Improved fast forward / rewind to use correct timing
 - Utilizing **higher resolution timers** (no more milliseconds... which caused issues on certain framerates that needed 1/2 milliseconds.
 - **Fixed big memory leak** in our use of Vectors (in FrameMapper, Cache, Keyframe classes). This was most noticeable on very large Clips (i.e. 60+ minutes duration), plus usage of hundreds of these clips (and hundreds of FrameMappers).... thus creating many, many GB of Vector data, and never really fully releasing the memory until we quit OpenShot.
 - **Improved sample rate detection** (especially on Windows). We output a warning if we see a mismatching sample rate. This is a bit misleading, as on Linux and Mac this mismatch does not really cause any issues (since the audio drivers resample all sounds anyway... if needed). But on Windows, some audio drivers simply don't resample correctly, so we output this message to help troubleshoot if your audio is playing back faster/slower pitch.
 - **Fixed a crash on shutdown** of audio thread

~~This PR also includes some new, experimental timing code to improve our video playback. Historically, our audio thread does not report an accurate position (i.e. in frame position). It is sometimes a few frames ahead, and sometimes not. It is variable, and depends on how many samples are requested from JUCE (and ultimately the audio drivers). Because of this "inaccuracy", our timing logic can not react immediately to audio/video sync... as it might be incorrect. Thus, we are now averaging the sync over many frames, to calculate if we are far enough away to trigger an adjustment. The adjustment now takes place in tiny increments, and over many frames to bring the sync back near zero frames difference.~~

~~This adjustment can cause a subtle oscillation over time, as the adjustment drifts down and up, always chasing zero. But due to the small size of adjustment, it should be undetectable. Also, once we reach zero drift, we disable the adjustment code until it is triggered again... hopefully ignoring the random jitter of the audio position inaccuracies. However, if over time, the average diff keeps increasing/decreasing, it will trigger another adjustment, and so on.~~

~~I have tried dozens of different approaches for correct audio/video sync during playback, and graphed them all for analysis. This approach, by far, has the best results so far... with minimal audio/video desync, and minimal oscillation. So I think it's time for more eyeballs to experiment with playback using this code... for a real-world test.~~


JacksonRG:
My bad! made a change and meant to push it to my personal repo to access it tomorrow. I'll undo that commit

src/Qt/PlayerPrivate.cpp
```diff
@@ -55,94 +57,75 @@ namespace openshot
...
+            // Calculate on-screen time for a single frame
+            int frame_speed = std::max(abs(speed), 1);
+            const auto frame_duration = double_micro_sec(1000000.0 / (reader->info.fps.ToDouble() * frame_speed));
+            const auto max_sleep = frame_duration * 4; ///< Don't sleep longer than X times a frame duration
+
+            // Pausing Code (if frame has not changed)
+            // Also pause at end of timeline, and pause if 'playing' and
+            // the pre-roll is not yet ready
+            if ((speed == 0 && video_position == last_video_position) ||
+                (speed != 0 && last_speed != speed) ||
```
JacksonRG, line 78: Noticed we dropped the `(video_position > reader->info.video_length)` check here. Do we no longer need that?

jonoomph:
This branch has finally passed our initial user testing (internally), and is ready for wider testing on the daily build.

PR #799, opened 2022-01-18
CI: Use windows-2022, start with a clean-slate MSYS2 image

ferdnyc:
Turns out that when setting up to build on Windows, it's almost 35% faster (**#WhoKnew?**; see msys2/MSYS2-packages#2788) if we start from a clean slate and directly install all 204 packages that make up our build environment. (That's the total count after dependency resolution.) Using a preinstalled MSYS2 package set, we have to update all of those packages _first_ (a slow process), before we can even start our own installs.

(When I say 35% faster, I mean the _package setup_ phase on Windows drops from ~ `9m 10s` to ~ `6m 10s`. The entire job runtime is still around `22m - 25m`, down from `25m - 30m`, so this only represents a 10-15% speedup overall. But still, that's 15% less thumb-twiddling time waiting for CI runs to complete.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
           - { os: ubuntu-18.04, shell: bash }
           - { os: ubuntu-latest, shell: bash }
           - { os: macos-latest, shell: bash }
-          - { os: windows-latest, shell: 'msys2 {0}' }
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging, because who doesn't love faster CI builds?

PR #800, opened 2022-01-18
Win CI: Prune Qt5 package set

ferdnyc:
And, hot on the heels of #799, the MSYS2 team informed me that they've broken up their Qt5 monopackage into separate, per-component packages. (`mingw-w64-x86_64-qt5` is now a metapackage that pulls them all in.)

Which means that, since libopenshot only needs `Qt5{Core, GUI, Widgets}` along with possibly `Qt5Svg`, we can skip installing all the rest of the modules, _and_ any of their dependencies that were only being pulled in for modules we don't use.

<s>(The `mingw-w64-x86_64-python3-pyqt5` package now only requires Qt modules as _optional_ dependencies, so a similar pruning should be possible for the OpenShot build as well! We'll have to pull in a few more modules there (Webkit, WebChannel, WebView, translations...) but still far fewer than **all** of them.)</s>

**Edit:** Oh, right... I forgot. What I meant was, "Someday, if we actually _have_ OpenShot CI jobs running on Windows, we'll be able to use a reduced package set there as well." **#DareToDream**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,10 @@ jobs:
               mingw-w64-x86_64-make
               mingw-w64-x86_64-cmake
               mingw-w64-x86_64-pkgconf
-              mingw-w64-x86_64-qt5
+              mingw-w64-x86_64-qt5-base
+              mingw-w64-x86_64-qt5-svg
+              mingw-w64-x86_64-qt5-winextras
+              mingw-w64-x86_64-qt5-tools
               mingw-w64-x86_64-libvpx
               mingw-w64-x86_64-ffmpeg
               mingw-w64-x86_64-zeromq


ferdnyc:
Updated because I left out `qt5-imageformats`, and while I'm not _sure_ it's strictly **necessary** for a libopenshot build, I feel better having those plugins installed. ISTR we've had problems with some of the unit tests failing if they're missing.

ferdnyc:
Nice! That puts the package install time for Windows runs at under 5 minutes, when a package cache available. (MSYS2 automatically caches its package set between runs, if there haven't been any changes upstream the cached set will be reused.)

That's assuming probably close-to-ideal conditions, sometimes it'll be longer. But with a good tailwind, the Windows CI run can now complete in under 20 minutes, which is something I wasn't sure I'd ever see.

(If the ffmpeg and opencv packages were pruned down so we could cast off some of _those_ unnecessary dependencies, we could potentially inch even closer to the holy grail of a 15-minute Windows build.)

ferdnyc:
Merging because, again, who's gonna say "no" to faster CI builds?

PR #804, opened 2022-01-21
Make coverage work on macOS, enable in CI

ferdnyc:
I finally seem to have beaten `lcov` into submission on macOS (only with `clang++`), so this will experimentally enable it in our CI runs, and add those results to Codecov's reporting along with the linux data.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
       - uses: haya14busa/action-cond@v1
         id: coverage
         with:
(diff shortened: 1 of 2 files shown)


ferdnyc:
(The huge change in coverage hits and misses, despite the relatively small change in covered percentage, is due to the fact that there are no macOS coverage reports for the comparison's BASE commit — aka, the `develop` branch HEAD. After merging, the `develop` CI builds will be publishing macOS reports as well, so that'll even out.)

PR #807, opened 2022-02-22
WIP: Expect caption timestamps relative to the clips position

JacksonRG:
**Closing, because this code was already accounting for the position of it's clip.**

# Issue:
Timestamps took absolute times for when to start and stop displaying.

This meant that they had to be adjusted any time a clip was moved.

# Solution:
Subtract the `position` (the time at the left edge of a clip) before creating the timestamp, and add it back in when calculating when to display the text.

I also created a [PR](https://github.com/OpenShot/openshot-qt/pull/4691) in openshot-qt for subtracting the clip's position before creating the timestamps. (These two PR's need each other to work properly)


jonoomph:
This is not required, as the Caption effect is supposed to be relative to the clip position.

PR #808, opened 2022-02-24
Don't force the sample rate to 44100, then complain about it

ferdnyc:
I noticed that OpenShot was now complaining about the 48kHz default sample rate I use, because it was under the impression that my card's default is 44.1kHz. Which I know for a fact is incorrect, but seems like _this_ probably had something to do with it:

https://github.com/OpenShot/libopenshot/blob/f2b89e3254d3db177ad99cef9034eea978d95678/src/Qt/AudioPlaybackThread.cpp#L33-L37

Now, I question the utility of those rate-checks _at all_, since clearly we're not actually checking it against the device's selected rate, we're checking it against the rate **WE'RE** defaulting to. But, we might as well at least default to the same rate everywhere, so we aren't contradicting ourselves.

### DIFF
--- a/src/Qt/AudioPlaybackThread.cpp
+++ b/src/Qt/AudioPlaybackThread.cpp
@@ -30,10 +30,10 @@ namespace openshot
 	// Global reference to device manager
 	AudioDeviceManagerSingleton *AudioDeviceManagerSingleton::m_pInstance = NULL;
 
-    // Create or Get audio device singleton with default settings (44100, 2)
+    // Create or Get audio device singleton with default settings (48000, 2)
     AudioDeviceManagerSingleton *AudioDeviceManagerSingleton::Instance()
     {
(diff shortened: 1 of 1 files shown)


ferdnyc:
So, even with the CI builds fixed (which they will be, by #809), this doesn't solve the problem.

And I think I know why. Jules pretty much spelled it out, in the `AudioIODevice` documentation:

juce

When the audio device isn't in use, `getCurrentSampleRate()` doesn't return the _default_ rate, it just returns _A_ rate. In fact, on Linux, it appears to simply return the **first** rate on the list. So, even though my audio output explicitly defaults to `48000 Hz`, OpenShot is now detecting `44,100 Hz` (a _supported_ rate, sure) and claiming that's the "default" rate. It's not, at all.

This check is... probably not good.

PR #810, opened 2022-02-24
Make Cache Less Aggressive

jonoomph:
Do not cache when paused (to allow for transform tool and timeline scrubbing to only render the current frame needed, for speed). This results in very smooth transform tool editing, very smooth timeline scrubbing, and I'm still seeing very smooth video playback when not paused. Needs more testing though.

The caching thread (before this PR) was really destroying the real-time transform tool performance, and the scrubbing performance. Not to mention, various other Timeline mutex locks, causing lots of crazy blocking issues all over the place. We were spamming Timeline::GetFrame from the cache thread non-stop, and when trying to make many Timeline changes in a short time (i.e. editing a transform), it just destroys performance.

**Need to test this on a variety of systems before merging.**

### DIFF
--- a/src/Qt/VideoCacheThread.cpp
+++ b/src/Qt/VideoCacheThread.cpp
@@ -27,7 +27,7 @@ namespace openshot
 	VideoCacheThread::VideoCacheThread()
 	: Thread("video-cache"), speed(0), last_speed(1), is_playing(false),
 	reader(NULL), current_display_frame(1), cached_frame_count(0),
-	min_frames_ahead(24), max_frames_ahead(OPEN_MP_NUM_PROCESSORS * 6)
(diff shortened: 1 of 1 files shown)


JacksonRG:
Testing this led me to realize I've had my cache set to disk. (Whoops)

Playback is smooth when I switched to memory cache. (as is develop) But I'm getting a constant stutter when it's set to disk. Should we not worry since memory is the default?

jonoomph:
Let me also test disk cache... to see the effects of this change

jonoomph:
I am getting smooth playback on Disk cache, but I also have an SSD. We have done some additional internal testing on this PR, and are seeing smooth playback on multiple systems and OSes. Merging this now to improve "transform" and "scrubbing" performance on our daily builds.

PR #811, opened 2022-02-28
Allow Caching on Initial 'Pause' (based on Timeline cache size)

jonoomph:
Allow cache on initial pause, to use the Timeline cache size to initially cache a bunch more frames... and then fully pause. This allows for a balance between not-caching during pause (for Transform and Scrubbing performance), but still caching frames on pause so the user can wait for a smoother playback experience.

**Use Cases Supported:**
- When real-time playback is not possible (due to CPU or project complexity), pausing playback will cache a bunch of frames, allowing for very smooth playback once the caching is completed
- Transforming a clip in the preview window (i.e. select a clip, and drag the blue handles) should be very smooth, once initial pause caching has finished.
- Scrubbing the timeline should be very fast (and not laggy)

### DIFF
--- a/src/Qt/VideoCacheThread.cpp
+++ b/src/Qt/VideoCacheThread.cpp
@@ -74,7 +74,7 @@ namespace openshot
         // Types for storing time durations in whole and fractional microseconds
         using micro_sec = std::chrono::microseconds;
         using double_micro_sec = std::chrono::duration<double, micro_sec::period>;
-        bool should_pause = false;
+        bool should_pause_cache = false;
 
 		while (!threadShouldExit() && is_playing) {
             // Calculate on-screen time for a single frame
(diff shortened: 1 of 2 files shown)


src/Qt/VideoCacheThread.cpp
```diff
@@ -74,6 +74,7 @@ namespace openshot
         // Types for storing time durations in whole and fractional microseconds
         using micro_sec = std::chrono::microseconds;
         using double_micro_sec = std::chrono::duration<double, micro_sec::period>;
+        bool should_pause = false;
```
JacksonRG: Little issue, but for clarity this could be called `should_pause_cache` or `halt_cache` (avoiding the word pause altogether)

Just to distinguish the _cache_ pausing from the _player_ pausing.

JacksonRG (approved):
Just a nitpick. Logic looks great, and dragging transforms is quick.

PR #812, opened 2022-03-02
Silence Clip's audio when requesting a Frame past the end of the reader

jonoomph:
Silence Clip's audio when requesting a Frame past the end of the reader. For example, if a Clip has 1000 frames, and the user requests frame 1001, we will return the last valid openshot::Frame object (repeating the last image), but we don't want to repeat the audio samples (causing a stutter). Any frame past the end of the reader, should always return audio silence. Also, fixed a few invalid comments, and added a Unit test.

This also required removing a unit test which would not allow invalid frames to be requested from a DummyReader object. Since most readers support requesting any frames #'s, regardless of validity, this change actually makes the DummyReader much more similar to our other readers.

This is addressing an issue that many customers have faced on OpenShot Cloud API, which allows customers to easily create a Clip object longer than a File (especially when using templates, and replacing a clip's source asset with a user uploaded file, that might be shorter). This results in the last frame of the video being repeated (stuttering the audio).

**Testing Instructions:**
- If all unit tests pass (since we are verifying the audio silence after the end of the reader)
- Create a project in openshot-qt, drop a video that has a solid tone (needs loud audio on all frames), save the *.osp project, edit the project in a text editor (update the "End" and "Duration" of the clip to be longer than the video file), re-open the *.osp project, and playback the project. Audio should go silent after the last valid frame, but the picture will remain/repeat.

### DIFF
--- a/tests/Clip.cpp
+++ b/tests/Clip.cpp
@@ -20,6 +20,7 @@
 #include <QSize>
 
 #include "Clip.h"
+#include "DummyReader.h"
 #include "Enums.h"
 #include "Exceptions.h"
 #include "Frame.h"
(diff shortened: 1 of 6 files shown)


tests/Clip.cpp
```diff
@@ -277,3 +278,62 @@ TEST_CASE( "has_video", "[libopenshot][clip]" )
...
+    CHECK(frame->GetAudioSamples(0)[600] == Approx(0.0).margin(0.00001));
+    CHECK(frame->GetAudioSamples(0)[1200] == Approx(0.0).margin(0.00001));
+
+    // Get the +2 past the end of the reader (should be audio silence)
+    frame = c1.GetFrame(32);
+
+    CHECK(frame->GetAudioSamples(0)[0] == Approx(0.0).margin(0.00001));
+    CHECK(frame->GetAudioSamples(0)[600] == Approx(0.0).margin(0.00001));
+    CHECK(frame->GetAudioSamples(0)[1200] == Approx(0.0).margin(0.00001));
+}
```
JacksonRG, line 339: Test looks great!

JacksonRG (approved):
LGTM!

PR #819, opened 2022-03-26
Apply time keyframes to the correct clip frame number

JacksonRG:
# Issue
1) Drag a video file onto the timeline
2) Reverse the video (`Right click` > `Time` > `Normal` > `Backwards 1x`)
3) At the beginning of the clip drag the preview to the left
4) At the end of the clip drag the preview to the right
5) Preview, and see the clip moving in the wrong direction
(Also test a 2x speed clip. The clip will only get half way across the screen)

# Test
Repeat the issue above, and observe the preview moving in the correct direction

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -411,6 +411,7 @@ std::shared_ptr<Frame> Clip::GetFrame(std::shared_ptr<openshot::Frame> backgroun
 		// Get time mapped frame number (used to increase speed, change direction, etc...)
 		// TODO: Handle variable # of samples, since this resamples audio for different speeds (only when time curve is set)
 		get_time_mapped_frame(original_frame, new_frame_number);
+		original_frame->number = frame_number;
 
 		// Apply local effects to the frame (if any)
 		apply_effects(original_frame);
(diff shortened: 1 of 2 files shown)


jonoomph:
@JacksonRG It seems like the "real" issue here is that we replace the "frame" object in the `apply_keyframes()` method, and thus the `frame->number` is replaced with the time mapped frame number. Perhaps the `apply_keyframes` method should never replace the `frame->number`, and thus, this PR gets a whole lot simpler, and we don't need the `unmapped_frame` variable (which is the wrong type by the way - should be int64_t).  How does that sound?

JacksonRG:
That makes sense. (I think I went unsigned because it shouldn't be able to be negative.) (though maybe that's bad flawed logic if the timing data is real weird)

jonoomph (approved):
LGTM

jonoomph:
@JacksonRG We should probably add a unit test for this one, since we've already seen a regression on this feature. For example, adding both a `scale_x` and `time` keyframe, and verifying keyframe values... What do you think?

JacksonRG:
Just rebased to re-run the checks. Merging pending success

PR #820, opened 2022-04-02
python/openshot.i: Convert to 4-space indents

ferdnyc:
Dumb housekeeping: Convert the `openshot.i` swig interface file in `bindings/python/` from tab indents to four-space indents, because it turns out the tabs were making it into the generated `openshot.py` file and making a mess of it. We all know how Python likes its indents Just So™.


ferdnyc:
Yet another painfully uncontroversial tweak.

PR #825, opened 2022-05-13
Fix CVTracker to use INT values for the Rect2d bounding box

jonoomph:
Ensure CVTracker bbox uses Integer values for it's Rect2d bounding box, and not float. The floats are killing the BOOSTING detector. Very sneaky bug... not sure why Rect2d allows for floats to be passed in.


jonoomph:
[Tracker Effect crashes with BOOSTING type](https://trello.com/c/tpOmaEpk/33-tracker-effect-crashes-with-boosting-type)

JacksonRG:
LGTM. What's the reason for the change from 134 to 132?

jonoomph:
I believe this is a rounding error from truncating the floats to ints.

PR #827, opened 2022-06-17
Quick Fix: Correct `class_name`s for Tracker and ObjectDetection

ferdnyc:
Effect class name is used when introspecting the available effects for a given build of libopenshot, for example when extracting translation strings for OpenShot. The `EffectInfo::class_name` member needs to match the name of the class it describes, which two of ours did not.

- `ObjectDetection` had its `class_name` set to "Object Detector", which obviously isn't the class name.
- `Tracker`, worse, was allowing `class_name` to be **SET** from the JSON parameters passed to its `SetJSON()` method, making the class name mutable at the whim of the class's callers.

### DIFF
--- a/src/effects/ObjectDetection.cpp
+++ b/src/effects/ObjectDetection.cpp
@@ -57,7 +57,7 @@ void ObjectDetection::init_effect_details()
     InitEffectInfo();
 
     /// Set the effect info
-    info.class_name = "Object Detector";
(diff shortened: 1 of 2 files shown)


ferdnyc:
Working on the macOS CI now (Homebrew has upgraded to Catch2 v3 which works very differently), but that's no reason to hold this up.

PR #828, opened 2022-06-18
Make unit tests work with Catch2v3 (for macOS CI)

ferdnyc:
Homebrew has upgraded Catch2 to version 3.0.1, which is structured significantly different from the previous version 2. This PR adds logic to check which version of Catch is loaded, and configures one of either `tests/catch2v2.h.in` or `tests/catch2v3.h.in` as the runtime header `openshot_catch.h` which is what all of our unit tests now `#include`. The different headers, along with some CMake logic to build the test runners differently, papers over the differences in the Catch configuration.

That papering-over includes the use of `using namespace Catch;` in the v3 header, which normally I would be the first one to  oppose. I'm still not happy about it, but because it's an internal header only ever intended to be included by our internal unit test code, I've found it within me to rationalize it as a necessary evil.

The changes here don't yet harness the full power of Catch2v3 to speed up compilation times, which they accomplished by splitting up their code into multiple headers so that you only import what you absolutely need. It may or may not be possible to take full, or better, advantage of that without sacrificing Catch2v2 compatibility, but for the moment I've punted to just using `#include <catch2/catch_all.hpp>` which emulates the previous monolithic `#include <catch2/catch.hpp>` setup from v2. There's definitely room for improvement in followup PRs, but the immediate goal is to fix macOS CI.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -68,20 +68,37 @@ if(NOT BUILD_TESTING)
   return()
 endif()
 
-# Create object library for test executable main(),
-# to avoid recompiling for every test
-add_library(catch-main OBJECT catch_main.cpp)
-target_link_libraries(catch-main PUBLIC Catch2::Catch2)
+add_library(openshot_catch2 INTERFACE)
+target_include_directories(openshot_catch2 INTERFACE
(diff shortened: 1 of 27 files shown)


ferdnyc:
Success! I've submitted #830 to track possible improvements, but this at least stops macOS jobs from failing.

PR #831, opened 2022-06-18
WIP: Add FPS, a Fraction subclass with converting powers

ferdnyc:
#### Motivation
To do basically any type of timing conversion in libopenshot, you need to know the FPS you're working with. We pass it into nearly every conversion function, in one way or another, because without it you're dead in the water.

In fact, FPS is so central to the process of conversions like frame<->time or frame<->sample that it makes sense for the object holding the FPS value to be able to _perform_ those conversions itself!

#### Solution (so far)

So, this (very much work-in-progress) PR is an attempt at empowering our FPS values, by subclassing `Fraction` into a new class, `FPS`. The `FPS` class doesn't carry any more **information** than `Fraction`, it just has additional methods useful for common conversions. (In fact, the classes are freely convertible in either direction. Which is why all function signatures that take timing values still pass them as `Fraction`s, not `FPS`s.)

Where I _have_ made use of `FPS`, so far, is in  the class member objects stored by things like `ReaderBase`, `WriterBase`, `FrameMapper`, etc. Any members holding `fps`, `video_timebase`, or `audio_timebase` values, which used to be type `Fraction`, are now type `FPS`.

The code doesn't (yet) make any more _use_ of those objects than it did when they were `Fraction`s, but all of the unit tests still pass and the change appears to have had zero effect on functionality, which was the goal for this stage.

#### Demonstration

The utility of `FPS` can be seen in the Python bindings (though the same operations are available to the C++ code directly):

```python
>>> import openshot
>>> # If we have an FPS value...
>>> fps = openshot.FPS(24000, 1001)
>>> # We can convert frame numbers to (start) times: 
>>> fps.time(1)
0.0
>>> fps.time(25)
1.001
>>> # We can convert time values to frame numbers
>>> fps.frame(10.0)
240
>>> fps.frame(10.01)
241
>>> # We can convert frame# to start sample, for a given samplerate
>>> fps.sample(1, 44100)
0
>>> fps.sample(2, 44100)
1839
>>> fps.sample(3, 44100)
3678
>>> fps.sample(2, 48000)
2002
>>> fps.sample(3, 48000)
4004
>>> # Want to know the sample count for the first 30 frames of video?
>>> sr = 44100
>>> from pprint import pp
>>> pp([fps.sample(f+1, sr) - fps.sample(f, sr)
...     for f in range(1, 31)], compact=True)
[1839, 1839, 1840, 1839, 1839, 1840, 1839, 1839, 1840, 1839, 1839, 1840, 1839,
 1839, 1840, 1839, 1839, 1840, 1839, 1839, 1840, 1839, 1839, 1840, 1839, 1839,
 1840, 1839, 1839, 1840]
```

#### A case of naming

You may notice that the new method names on `FPS` are lowercase. Yeah, sorry. I just couldn't, with the capitalized method names. `TitleCase` is for classes, not for functions. `SomeClass::lowerTitleCase` (aka `antiSentenceCase`) is fine, but not `SomeClass::FullTitleCase`.

You may also notice that `FPS` is fully capitalized, rather than the class name being `Fps`. Yes, yes it is. I'm not even remotely sorry, on that one.

Though I'll have to consider whether a hypothetical method to retrieve a Frame's FPS should be called `Frame::FPS()`  or `Frame::fps()`. (`Frame::fPS()` is no more on the table than `Frame::Fps()` is.) I think I'm leaning towards the latter, if for no other reason than to avoid clashes with the class name. (...And now I suddenly appreciate just how critical things like `decltype()` and `typename` are to writing template code.)

...Even more squirrelly, would the Frame class's method to self-convert into other frame rates be called `Frame::toFPS()` or `Frame::toFps()`? That's kind of a hard one!

FPS also has some other adjustments compared to Fraction. Like, where the Python string representation of a Fraction is <code><var>num</var>:<var>den</var></code> (used by the OpenShot Profile Selector dialog when displaying Profile details), for FPS I've changed that to <code><var>num</var>/<var>den</var></code>. Though, there are arguments for and against displaying _either_ type of value in either format, so I'm not really married to this change yet.

#### Next steps

There's plenty more to add (aside from, like, actually _using_ this in the code itself), but that's the basic idea. 

- [ ] **Additional conversions:**: More types of conversions could also be added, if they'd be useful. (I thought about adding time<->sample, but I couldn't really think of anyplace where that would actually be useful for anything.)
    Some sort of ability to convert _between_ different sample rates in some fashion, OTOH, might actually be handy. It could theoretically centralize the calculations necessary to configure resamplers and size buffers, which are currently scattered, in whole or in part, across ad-hoc implementations found in at least `FFmpegReader`, `AudioResampler`, and I believe `FFmpegWriter` as well.

- [ ] **Storing FPS in Frame?:** Another thing I've been thinking about is storing an `FPS` value _inside_ each Frame object that gets created. Frames basically _do_ have an inherent FPS, since things like their sample count and corresponding time duration (also not currently tracked within Frame itself) are completely dependent on the FPS from which it was created, and Frames moving between FPS values need to be converted (hence FrameMapper).
    But if Frame carried its FPS internally, then it could produce values like start sample and sample duration / time duration on demand, and determining when conversions are required would be simpler.

- [ ] **Unit tests:** `FPS` needs them for itself, and test coverage of anywhere else it's used should ideally be increased to 100% as well.

- [ ] **Documentation:** Filling in the class's missing docstrings, at the very least. (But not the ones inherited from `Fraction`, because it's a _feature_ of Doxygen that you don't have to repeat documentation in every subclass that inherits a member/method. One we should make better use if.)

### DIFF
--- a/bindings/ruby/openshot.i
+++ b/bindings/ruby/openshot.i
@@ -59,7 +59,7 @@
   #define RB_RSHIFT(a, b) RSHIFT(a, b)
   #undef RSHIFT
 #endif
-include "OpenShotVersion.h"
+#include "OpenShotVersion.h"
 #include "ReaderBase.h"
 #include "WriterBase.h"
     #include "ChunkReader.h"
--- a/bindings/ruby/openshot.i
+++ b/bindings/ruby/openshot.i
@@ -59,43 +59,44 @@
   #define RB_RSHIFT(a, b) RSHIFT(a, b)
   #undef RSHIFT
 #endif
-#include "OpenShotVersion.h"
+include "OpenShotVersion.h"
 #include "ReaderBase.h"
 #include "WriterBase.h"
-#include "AudioDevices.h"
-#include "CacheBase.h"
-#include "CacheDisk.h"
-#include "CacheMemory.h"
-#include "ChannelLayouts.h"
-#include "ChunkReader.h"
-#include "ChunkWriter.h"
+    #include "ChunkReader.h"
+    #include "ChunkWriter.h"
+    #include "DummyReader.h"
+    #include "FFmpegReader.h"
+    #include "FFmpegWriter.h"
+    #include "FrameMapper.h"
+    #include "QtHtmlReader.h"
+    #include "QtImageReader.h"
+    #include "QtTextReader.h"
+#include "Fraction.h"
+    #include "FPS.h"
 #include "ClipBase.h"
-#include "Clip.h"
+    #include "Clip.h"
+#include "CacheBase.h"
+    #include "CacheDisk.h"
+    #include "CacheMemory.h"
+#include "AudioDevices.h"
+#include "ChannelLayouts.h"
 #include "Coordinate.h"
 #include "Color.h"
-#include "DummyReader.h"
 #include "EffectBase.h"
-#include "Effects.h"
-#include "EffectInfo.h"
+    #include "Effects.h"
+    #include "EffectInfo.h"
 #include "Enums.h"
 #include "Exceptions.h"
-#include "FFmpegReader.h"
-#include "FFmpegWriter.h"
-#include "Fraction.h"
 #include "Frame.h"
-#include "FrameMapper.h"
 #include "PlayerBase.h"
+    #include "QtPlayer.h"
 #include "Point.h"
 #include "Profiles.h"
-#include "QtHtmlReader.h"
-#include "QtImageReader.h"
-#include "QtPlayer.h"
-#include "QtTextReader.h"
 #include "KeyFrame.h"
 #include "RendererBase.h"
 #include "Settings.h"
 #include "TimelineBase.h"
-#include "Timeline.h"
+    #include "Timeline.h"
 #include "ZmqLogger.h"
 
 /* Move FFmpeg's RSHIFT to FF_RSHIFT, if present */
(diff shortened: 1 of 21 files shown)


bindings/ruby/openshot.i
```diff
@@ -59,43 +59,44 @@
   #define RB_RSHIFT(a, b) RSHIFT(a, b)
   #undef RSHIFT
 #endif
-#include "OpenShotVersion.h"
+include "OpenShotVersion.h"
```
ferdnyc: Well, that's a typo. :-/
```suggestion
#include "OpenShotVersion.h"
```

ferdnyc:
> Where I _have_ made use of `FPS`, so far, is in the class member objects stored by things like `ReaderBase`, `WriterBase`, `FrameMapper`, etc. Any members holding `fps`, `video_timebase`, or `audio_timebase` values, which used to be type `Fraction`, are now type `FPS`.

Actually, I'm reconsidering that for `video_timebase` and `audio_timebase`, because the methods won't be _correct_ — they assume that the data stored in the `FPS` object is a fraction where `num` represents frames and `den` represents seconds. If it's inverted, calling (for example) `info.video_timebase.time(60)` will produce an insane value that's definitely **not** the time in seconds corresponding to frame 60.

I'm actually tempted to say we shouldn't even store those values as _Fractions_. There is only **ONE** instance in the code (outside of values housekeeping / data duplication / JSON [de]serialization) where either of those variables is ever accessed as a Fraction, rather than by calling the `.ToDouble()`, `.Reciprocal().ToFloat()`, or etc. methods to convert it to  a numeric value. And that's this one:
https://github.com/OpenShot/libopenshot/blob/7c6fcb0c5149f4b34bd078f37a4e546ea1dc755d/src/FFmpegWriter.cpp#L1299-L1307

Aside from that one instance of the code setting the `time_base` values for the stream and context, for all intents and purposes both `audio_timebase` and `video_timebase` are already treated as `double`.

PR #834, opened 2022-06-21
Fix regression caused by Object Detector renamed to Object Detection

jonoomph:
Fixed regression in https://github.com/OpenShot/libopenshot/pull/827. Many places in libopenshot and openshot-qt refer to the class name of this effect, causing errors on the daily build, and breakages in various places.


jonoomph:
[Object Detector throwing error in daily build](https://trello.com/c/nEUDFp1W/143-object-detector-throwing-error-in-daily-build)

JacksonRG (approved):
LGTM

PR #837, opened 2022-07-21
Improved PTS support (presentation time stamps)

jonoomph:
This is a large refactor of how FFmpegReader class utilizes the timestamps, to make better/quicker decisions about audio/video data.

**Improvements:**
- Improved PTS offset logic (to offset our streams so that at least one of them starts at position 0.0)
- Removed missing frame detection (not needed anymore)
- Use timestamps to reason about missing frames
- Fixed many async decoding issues (jumbled frames, missing beginning or ending frames)
- Much improved support for video/audio files with missing chunks/gaps (i.e. missing timestamps)
- Improved CheckWorkingFrames() method, to determine when working frames are ready to be finalized
- Improved FPS detection for invalid or missing FPS
- Fixed a crash (only reproducible on Windows), caused by 2 calls to AddImage() for the same frame

**What does all this mean?**
Video/audio files contain packets. Packets have timestamps to help us keep them in order, and know **when** to use them. However, packets must be decoded into actual usable data first (compressed bytes => image/audio data). Modern decoders (FFmpeg v3+) utilize threads to decode this data, and handle things asynchronously. This refactor mainly focuses on this async code, to dramatically improve the reliability, speed, and accuracy of the frame data we are accessing with FFmpegReader. OpenShot still had lots of legacy code, which expected the current packet to be processed synchronously from the decoder, causing all sorts of mis-timing events, or flat out missing frames. 

**Bottom line:** This PR will improve the performance for most users, prevent crashes for some users, and improve compatibility with many, many video files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,6 @@ jobs:
         sys:
           - { os: ubuntu-18.04, shell: bash }
           - { os: ubuntu-latest, shell: bash }
-          - { os: macos-latest, shell: bash }
           - { os: windows-2022, shell: 'msys2 {0}' }
         compiler:
           - { cc: gcc, cxx: g++ }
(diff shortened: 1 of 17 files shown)


jonoomph:
[Jumbled frames and audio lag (PTS Offset issue)](https://trello.com/c/OZbY1sQA/45-jumbled-frames-and-audio-lag-pts-offset-issue)

PR #843, opened 2022-08-24
Improving Caption Effect (centering, background color, line spacing...)

jonoomph:
Improving **Caption Effect** with many bug fixes and improvements:

- background color now sticks to the text
- background border now scales correctly (at different resolutions)
- background padding now scales correctly
- stroke size now scales correctly
- margins are now defaulted to 15% on left/right
- text caption area now centers between available left/right margins
- background color can now be faded in and out, animated
- new property: line_spacing, a % from 1.0 (100%) to affect how much space are between lines
- Use each font's default lineSpacing, instead of a fixed distance between lines. This property can be adjusted with the new line_spacing keyframe.

Caption-align-area-center

### DIFF
--- a/src/effects/Caption.cpp
+++ b/src/effects/Caption.cpp
@@ -26,9 +26,9 @@
 using namespace openshot;
 
 /// Blank constructor, useful when using Json to load the effect properties
-Caption::Caption() : color("#ffffff"), stroke("#a9a9a9"), background("#ff000000"), background_alpha(0.0), left(0.25), top(0.7), right(0.1),
+Caption::Caption() : color("#ffffff"), stroke("#a9a9a9"), background("#ff000000"), background_alpha(0.0), left(0.15), top(0.7), right(0.15),
 					 stroke_width(0.5), font_size(30.0), font_alpha(1.0), is_dirty(true), font_name("sans"), font(NULL), metrics(NULL),
-					 fade_in(0.35), fade_out(0.35), background_corner(10.0), background_padding(20.0)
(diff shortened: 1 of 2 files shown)


jonoomph:
[Caption Improvements (line spacing, background color, centering)](https://trello.com/c/WIYHEvyK/145-caption-improvements-line-spacing-background-color-centering)

src/effects/Caption.cpp
```diff
@@ -153,46 +153,28 @@ std::shared_ptr<openshot::Frame> Caption::GetFrame(std::shared_ptr<openshot::Fra
 	double fade_in_value = fade_in.GetValue(frame_number) * fps.ToDouble();
 	double fade_out_value = fade_out.GetValue(frame_number) * fps.ToDouble();
 	double right_value = right.GetValue(frame_number);
-	double background_corner_value = background_corner.GetValue(frame_number);
-	double padding_value = background_padding.GetValue(frame_number);
+	double background_corner_value = background_corner.GetValue(frame_number) * scale_factor;
+	double padding_value = background_padding.GetValue(frame_number) * scale_factor;
+	double stroke_width_value = stroke_width.GetValue(frame_number) * scale_factor;
+	double line_spacing_value = line_spacing.GetValue(frame_number);
+	double metrics_line_spacing = metrics.lineSpacing();
```
jonoomph, line 160: metrics.lineSpacing() is already based on the FontSize (which has already been scaled to the preview window size) - thus, this metric is also scaled

JacksonRG: Makes sense! And looks like that was the original issue making spacing too large.

JacksonRG (approved):
LGTM. The background box in particular looks much simpler

PR #844, opened 2022-08-28
Fixing transition end frame calculation

jonoomph:
The timeline was calculating the right edge of transitions incorrectly (by adding an extra frame). This fixes that mis-calculation, and brings it inline with our clip intersection logic.


jonoomph:
Merging to fix this bug quickly in production. Will review this post merge.

jonoomph:
[Clip/Transition Snapping is not precise enough (causing misalignments)](https://trello.com/c/heJ2rDOf/160-clip-transition-snapping-is-not-precise-enough-causing-misalignments)

JacksonRG (commented):
Looks good, should remove the extra frames we’ve been seeing.

JacksonRG:
[Cut Clips have Wrong Frame Numbers](https://trello.com/c/mRSw0Pq4/165-cut-clips-have-wrong-frame-numbers)

PR #845, opened 2022-09-02
Detect video length (i.e. # of video frames) from Video Metadata

jonoomph:
We were calculating the `video_length` from `duration * fps`. While this works pretty well, to approximate the # of frames encoded in a video, it is occasionally different by a small amount, and in my testing, the metadata `Stream->nb_frames` seems more accurate.

I've tested this with a bunch of test videos, and this seems to work great. :crossed_fingers:

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -822,7 +822,8 @@ void FFmpegReader::UpdateVideoInfo() {
 	}
 
 	// Get the # of video frames (if found in stream)
-	if (pStream->nb_frames > 0) {
+	// Only set this 1 time (this method can be called multiple times)
+	if (pStream->nb_frames > 0 && info.video_length <= 0) {
 		info.video_length = pStream->nb_frames;
 	}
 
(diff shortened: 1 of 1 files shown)


jonoomph:
[Invalid # of Frames detected in Certain Videos](https://trello.com/c/IW5ggDFw/163-invalid-of-frames-detected-in-certain-videos)

src/FFmpegReader.cpp
```diff
@@ -831,8 +837,11 @@ void FFmpegReader::UpdateVideoInfo() {
 		// Yes, a duration was found
 		is_duration_known = true;
 
-		// Calculate number of frames
-		info.video_length = round(info.duration * info.fps.ToDouble());
+		// Calculate number of frames (if not already found in metadata)
+		// Only set this 1 time (this method can be called multiple times)
+		if (info.video_length <= 0) {
+			info.video_length = round(info.duration * info.fps.ToDouble());
```
JacksonRG, line 843: So we only calculate `video_length` from `duration` if we have `duration` and we're missing a value for `video_length`. Because we prefer the data from `nb_frames`? Makes sense.

JacksonRG (approved):
LGTM

PR #848, opened 2022-09-22
Allow intentional line breaks

JacksonRG:
We were only adding to the `current_y` if there were words left on the current line of text. However any lines of text until the next timestamp show up.

This change allows users to specify line breaks in captions.


JacksonRG:
[Allow multi-line captions](https://trello.com/c/OWN5pv0r/169-allow-multi-line-captions)

jonoomph:
LGTM

PR #849, opened 2022-09-24
Fixing infinite loop caused by override of has_video/has_audio

jonoomph:
Fixing behavior of manual override of `has_video` or `has_audio`, to remove unused packets, and not get stuck in an endless loop due to 1 or 2 pending packets that are no longer needed after the override.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -595,7 +595,8 @@ void FFmpegReader::Close() {
 		// Drain any packets from the decoder
 		packet = NULL;
 		int attempts = 0;
-		while (packets_decoded < packets_read && attempts < 256) {
+		int max_attempts = 128;
+		while (packets_decoded < packets_read && attempts < max_attempts) {
 			ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::Close (Drain decoder loop)",
 													 "packets_read", packets_read,
 													 "packets_decoded", packets_decoded,
@@ -925,12 +926,14 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
 	// Allocate video frame
 	bool check_seek = false;
 	int packet_error = -1;
+	int attempts = 0;
+	int max_attempts = 128;
 
 	// Debug output
 	ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::ReadStream", "requested_frame", requested_frame, "max_concurrent_frames", max_concurrent_frames);
 
 	// Loop through the stream until the correct frame is found
-	while (true) {
+	while (attempts < max_attempts) {
 		// Check if working frames are 'finished'
 		if (!is_seeking) {
 			// Check for final frames
@@ -978,6 +981,15 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
 			// Process Audio Packet
 			ProcessAudioPacket(requested_frame);
 		}
+
+		// Remove unused packets (sometimes we purposely ignore video or audio packets,
+		// if the has_video or has_audio properties are manually overridden)
+        if ((!info.has_video && packet && packet->stream_index == videoStream) ||
+            (!info.has_audio && packet && packet->stream_index == audioStream)) {
+            RemoveAVPacket(packet);
+            packet = NULL;
+            packets_decoded++;
+        }
 		
 		// Determine end-of-stream (waiting until final decoder threads finish)
 		// Force end-of-stream in some situations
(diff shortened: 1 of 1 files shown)


jonoomph:
[Infinite loop caused by manual override of `has_video` or `has_audio`](https://trello.com/c/xkoOOX4v/172-infinite-loop-caused-by-manual-override-of-hasvideo-or-hasaudio)

src/FFmpegReader.cpp
```diff
@@ -595,7 +595,8 @@ void FFmpegReader::Close() {
 		// Drain any packets from the decoder
 		packet = NULL;
 		int attempts = 0;
-		while (packets_decoded < packets_read && attempts < 256) {
+		int max_attempts = 128;
+		while (packets_decoded < packets_read && attempts < max_attempts) {
```
jonoomph, line 599: Typically this loop never approaches the `max_attempts`, and instead breaks out of the loop as soon as the requested frame is found, typically within a few iterations. However, there are rare cases where an infinite loop can arise, and we don't want this to loop forever.

src/FFmpegReader.cpp
```diff
@@ -926,11 +927,18 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
...
+	// no more packets can be found though.
+	int attempts = 0;
+	int max_attempts = 30 * 30 * 60;
+
 	// Debug output
 	ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::ReadStream", "requested_frame", requested_frame, "max_concurrent_frames", max_concurrent_frames);
 
 	// Loop through the stream until the correct frame is found
-	while (true) {
+	while (attempts < max_attempts) {
```
jonoomph, line 941: Typically this loop never approaches the max_attempts, and instead breaks out of the loop as soon as the requested frame is found, or as soon as all packets have been decoded, typically within a few iterations. However, there are rare cases where an infinite loop can arise, and we don't want this to loop forever.

jonoomph: Mostly caused by manual override to the `has_video` or `has_audio` properties after the reader has called Open(), leaving a few read packets undecoded.

jonoomph:
Merging now since we have live freezes in our daily build I'm trying to resolve. Review to come soon.

src/FFmpegReader.cpp
```diff
@@ -926,11 +927,18 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
 	bool check_seek = false;
 	int packet_error = -1;
 
+	// Seeking can sometimes cause us to be quite far away from the requested
+	// frame, and thus, we need to iterate a large number of packets to find
+	// the requested position. This is mostly to prevent infinite loops when
+	// no more packets can be found though.
+	int attempts = 0;
+	int max_attempts = 30 * 30 * 60;
```
JacksonRG, line 935: Why `30 * 30 * 60` Should it be a number derived from the FPS?

src/FFmpegReader.cpp
```diff
@@ -978,7 +986,16 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
...
 		}
-		
+
+		// Remove unused packets (sometimes we purposely ignore video or audio packets,
+		// if the has_video or has_audio properties are manually overridden)
+		if ((!info.has_video && packet && packet->stream_index == videoStream) ||
+			(!info.has_audio && packet && packet->stream_index == audioStream)) {
+			RemoveAVPacket(packet);
+			packet = NULL;
+			packets_decoded++;
```
JacksonRG, line 996: This makes a lot of sense that we're discarding packets from streams we want to ignore.

src/FFmpegReader.cpp
```diff
@@ -995,6 +1012,7 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
 			end_of_file = true;
 			break;
 		}
+        attempts++;
```
JacksonRG, line 1015: I saw one `continue` statement in this loop. Would it be better to increment attempts at the top of the loop to guarantee we're moving toward breaking out of the loop?

JacksonRG (commented):
Looks good. Should prevent infinite loops. (Unless the `continue` gets hit a bunch of times.)

src/FFmpegReader.cpp
jonoomph: This is kind of a hack. In reality, we should be counting the packets read for `video` and `audio`, and keeping separate counts. Because, if we read a `video` packet, we need to keep calling `ProcessVideoPacket` until the counts match, for example, regardless of the `has_video` property.

Maybe this is a good time to fix that, because we should be able to delete this hacky code.

jonoomph: Refactored this hack out of the code: https://github.com/OpenShot/libopenshot/pull/851

PR #851, opened 2022-09-27
Refactor of Packet tracking & EOF detection

jonoomph:
Refactor of packet counting and EOF detection. Instead of tracking overall packets read & decoded, we now track per stream (video and audio), so we can more accurately know when a packet is still pending for one of those streams (i.e. async decoding).

Also, refactored variables into a struct, to keep them together (and perhaps, more understandable)

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -73,9 +73,8 @@ FFmpegReader::FFmpegReader(const std::string &path, bool inspect_reader)
 		  path(path), is_video_seek(true), check_interlace(false), check_fps(false), enable_seek(true), is_open(false),
 		  seek_audio_frame_found(0), seek_video_frame_found(0),is_duration_known(false), largest_frame_processed(0),
 		  current_video_frame(0), packet(NULL), max_concurrent_frames(OPEN_MP_NUM_PROCESSORS), audio_pts(0),
-		  video_pts(0), pFormatCtx(NULL), packets_read(0), packets_decoded(0), videoStream(-1), audioStream(-1),
-		  pCodecCtx(NULL), aCodecCtx(NULL), pStream(NULL), aStream(NULL), pFrame(NULL), previous_packet_location{-1,0},
-		  video_eof(false), audio_eof(false), packets_eof(false), end_of_file(false) {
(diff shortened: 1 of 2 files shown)


jonoomph:
[Infinite loop caused by manual override of `has_video` or `has_audio`](https://trello.com/c/xkoOOX4v/172-infinite-loop-caused-by-manual-override-of-hasvideo-or-hasaudio)

src/FFmpegReader.cpp
```diff
@@ -596,15 +590,15 @@ void FFmpegReader::Close() {
...
-		while (packets_decoded < packets_read && attempts < max_attempts) {
+		while (packet_status.packets_decoded() < packet_status.packets_read() && attempts < max_attempts) {
 			ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::Close (Drain decoder loop)",
-													 "packets_read", packets_read,
-													 "packets_decoded", packets_decoded,
+													 "packets_read", packet_status.packets_read(),
+													 "packets_decoded", packet_status.packets_decoded(),
 													 "attempts", attempts);
-			if (info.has_video) {
+			if (packet_status.video_decoded < packet_status.video_read) {
```
JacksonRG, line 598: looks like this will save some compute time :+1:

src/FFmpegReader.cpp
```diff
@@ -991,44 +978,51 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
...
 			packet = NULL;
-			packets_decoded++;
 		}
 
 		// Determine end-of-stream (waiting until final decoder threads finish)
 		// Force end-of-stream in some situations
-		end_of_file = packets_eof && video_eof && audio_eof;
-		if ((packets_eof && packets_read == packets_decoded) || end_of_file) {
+		packet_status.end_of_file = packet_status.packets_eof && packet_status.video_eof && packet_status.audio_eof;
+		if ((packet_status.packets_eof && packet_status.packets_read() == packet_status.packets_decoded()) || packet_status.end_of_file) {
```
JacksonRG, line 997: Looks like this should always break, since we decode as long as there are packets read.

JacksonRG (approved):
LGTM. I was confused by, but then understood why we took away the hardcoded limit on maximum attempts.

PR #856, opened 2022-10-13
Detect audio EOF when all packets are decoded

jonoomph:
Detect audio EOF when all packets are decoded - since not all audio streams get a NULL packet at the end of stream


jonoomph:
[Audio decoding bug can cause OpenShot to freeze forever](https://trello.com/c/6ZUPh6mS/183-audio-decoding-bug-can-cause-openshot-to-freeze-forever)

PR #857, opened 2022-10-15
Fix Regression causing Blank audio in many files

jonoomph:
Remove regression causing blank audio on many audio files. Replacing the broken logic in a different way, to force a NULL packet to the ProcessAudioPacket() function at the end of stream / EOF.


jonoomph:
[Audio decoding bug can cause OpenShot to freeze forever](https://trello.com/c/6ZUPh6mS/183-audio-decoding-bug-can-cause-openshot-to-freeze-forever)

PR #858, opened 2022-10-15
Break out of cache loop when caching is disabled

jonoomph:
Break out of cache loop when caching is temporarily disabled (i.e. when transforming, editing keyframes, rotating, moving, resizing, etc...). This creates a much smoother editing workflow, and removes an initial delay from the transform tool caused by the cache loop still running.


jonoomph:
[Initial delay when using "Transform" tool caused by Cache Thread](https://trello.com/c/NGCY8loB/184-initial-delay-when-using-transform-tool-caused-by-cache-thread)

PR #859, opened 2022-10-16
Removing cache clearing from Timeline::Clear method - unneeded

jonoomph:
This tends to be unstable and is causing crashes, not to mention we already Clear all cache in the destructor, so this is simply not needed.


jonoomph:
[Crash on Exit (while previewing)](https://trello.com/c/aSX6jtA0/185-crash-on-exit-while-previewing)

PR #862, opened 2022-10-24
Always cache starting 1 frame before the playhead's position

jonoomph:
Always cache starting 1 frame previous to the playhead - to prevent the Seek method from clearing all cache. Each time a clip is added to the timeline, we call Seek() again, which checks to see if the previous frame is cached. This logic is designed to catch the case where we are seeking to a new position on the timeline, and clear the cache.


jonoomph:
[Improve `DeepCopy` Performance](https://trello.com/c/u6x8DzZD/192-improve-deepcopy-performance)

PR #863, opened 2022-10-24
Support many new image formats with our FFmpegReader

jonoomph:
Support many new image formats with our FFmpegReader, by safely protecting empty packets in `GetPacketPTS`, and generating a duration and correctly setting the `has_single_image` property.

Essentially, this is a 1 frame video, but we want OpenShot to pretend it's a 60 minute long image.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -819,6 +819,14 @@ void FFmpegReader::UpdateVideoInfo() {
 		info.duration = float(info.file_size) / info.video_bit_rate;
 	}
 
+	// Certain "image" formats do not have a valid duration
+	if (info.duration <= 0.0f && pStream->duration == AV_NOPTS_VALUE && pFormatCtx->duration == AV_NOPTS_VALUE) {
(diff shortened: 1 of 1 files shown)


jonoomph:
[Webp (and other image formats) not correctly supported](https://trello.com/c/8ADuW4mA/193-webp-and-other-image-formats-not-correctly-supported)

src/FFmpegReader.cpp
```diff
@@ -1849,12 +1857,17 @@ void FFmpegReader::Seek(int64_t requested_frame) {
...
 // Get the PTS for the current video packet
 int64_t FFmpegReader::GetPacketPTS() {
-	int64_t current_pts = packet->pts;
-	if (current_pts == AV_NOPTS_VALUE && packet->dts != AV_NOPTS_VALUE)
-		current_pts = packet->dts;
-
-	// Return adjusted PTS
-	return current_pts;
+    if (packet) {
+        int64_t current_pts = packet->pts;
```
jonoomph: Protect from NULL packet

PR #867, opened 2022-10-31
Add fallback for AVFrame.pkt_dts (certain codecs fail to find a valid video PTS)

jonoomph:
This is a regression caused by recent PTS changes, and apparently certain video codecs, such as AV1, do not correctly set the AVFrame->pts field, and instead rely on the AVFrame->pkt_dts. Strange, but the fix is a simple fallback.


jonoomph:
[Certain AVI / AV1 files do not playback (appear as static image)](https://trello.com/c/q8Yektzm/196-certain-avi-av1-files-do-not-playback-appear-as-static-image)

PR #868, opened 2022-11-01
New AudioWaveformer Class (for generating fast, graph-friendly audio datasets) 

jonoomph:
Use Root Mean Square to calculate both the "average" and "max" samples per period. Allow user to specific # of samples per second, and specify a particular "channel" or "all channels". Also, an option for scaling the waveform to a max value of 1.0, so we can ensure all waveforms are easy to view. Also, including new unit tests to verify this class actually works.


jonoomph:
[Waveform + Volume keyframe issues](https://trello.com/c/P88Tv4kW/194-waveform-volume-keyframe-issues)

bindings/python/openshot.i
```diff
@@ -9,7 +9,7 @@
 //
 // SPDX-License-Identifier: LGPL-3.0-or-later
 
-%module openshot
+%module("threads"=1) openshot
```
jonoomph, line 12: This allows all libopenshot methods to work async and not block the global Python lock

src/AudioWaveformer.cpp
```diff
@@ -0,0 +1,124 @@
...
+                        chunk_squared_sum += rms_sample_value;
+                        chunk_max = std::max(chunk_max, rms_sample_value);
+                    }
+                }
+
+                sample_index += 1;
+
+                // Cut-off reached
+                if (sample_index % sample_divisor == 0) {
+                    float avg_squared_sum = chunk_squared_sum / (sample_divisor * channel_count);
```
uglycoder99, line 95: Potential for divide by Zero?

src/AudioWaveformer.h
```diff
@@ -0,0 +1,101 @@
...
+
+        /// Resize both datasets
+        void resize(int total_samples) {
+            max_samples.resize(total_samples);
+            rms_samples.resize(total_samples);
+        }
+
+        /// Zero out # of values in both datasets
+        void zero(int total_samples) {
+            for (auto s = 0; s < total_samples; s++) {
```
uglycoder99, line 42: Is this the most efficient way to zero out a vector?  I don't know just asking

src/AudioWaveformer.h
```diff
@@ -0,0 +1,101 @@
...
+                max_samples[s] = 0.0;
+                rms_samples[s] = 0.0;
+            }
+        }
+
+        /// Scale # of values by some factor
+        void scale(int total_samples, float factor) {
```
uglycoder99, line 49: Same comment here about vector operations.  There may be a standard CPP way of doing this

uglycoder99 (commented):
Everything looked good but I saw a couple of spots where division by zero seemed like a possibility and vector operations looked like they could be more efficient or at least use standard methods for apply operations across vectors.

src/AudioWaveformer.h
jonoomph: Good question! Looks like `std::fill(v.begin(), v.end(), 0);` might be a better method, which apparently copies more bytes at the same time than I'm currently doing.

src/AudioWaveformer.cpp
jonoomph: Good suggestion, and I'm going to add a unit test which passes in a file with 0 audio channels, to flush out any errors around this point.

PR #872, opened 2022-11-15
Refactor FFmpegReader GetAVFrame (for AV1 & async decoding)

jonoomph:
Some decoders, for example `dav1d`, process packets asynchronously and are multi-threaded. OpenShot was not correctly handling these async decoders, which resulted in dropping packets and freezing videos. This PR adds an AV1 decoding unit test as a verification, and correctly retries packets that fail to be sent to the decoder.

For example, if a decoder can accept many packets before returning an AVFrame, it might also stop accepting new packets... until it's done processing the previous ones. In those cases, we need to halt getting the next packet, until we can successfully send the current packet.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -1114,97 +1114,112 @@ bool FFmpegReader::GetAVFrame() {
 		if (send_packet_err < 0 && send_packet_err != AVERROR_EOF) {
             ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::GetAVFrame (Packet not sent)", "send_packet_err", send_packet_err);
             if (send_packet_err == AVERROR(EAGAIN)) {
-                ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::GetAVFrame (AVERROR(EAGAIN) user must read output with avcodec_receive_frame()");
+                ZmqLogger::Instance()->AppendDebugMethod("FFmpegReader::GetAVFrame (AVERROR(EAGAIN): user must read output with avcodec_receive_frame()");
             }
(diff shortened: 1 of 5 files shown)


jonoomph:
[`test-video.mp4` not decoding video on Windows 10 (libaom-av1 codec)](https://trello.com/c/BWlyxPzp/210-test-videomp4-not-decoding-video-on-windows-10-libaom-av1-codec)

src/FFmpegReader.cpp
```diff
@@ -952,11 +955,13 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
 			break;
 		}
 
-		// Get the next packet
-		packet_error = GetNextPacket();
-		if (packet_error < 0 && !packet) {
-			// No more packets to be found
-			packet_status.packets_eof = true;
+		if (!resend_packet || !packet) {
```
jonoomph: Only get next packet if we are NOT resending the current packet

src/FFmpegReader.cpp
```diff
@@ -977,7 +982,7 @@ std::shared_ptr<Frame> FFmpegReader::ReadStream(int64_t requested_frame) {
 
 		// Video packet
 		if ((info.has_video && packet && packet->stream_index == videoStream) ||
-			(info.has_video && !packet && packet_status.video_decoded < packet_status.video_read) ||
+			(info.has_video && packet_status.video_decoded < packet_status.video_read) ||
```
jonoomph, line 985: Call `ProcessVideoPacket()` anytime `video_decoded` < `video_read`... so we can potentially resend a packet if needed, or try and retrieve a processed AVFrame from the decoder.

PR #875, opened 2022-11-23
Prevent seeking past EOF

jonoomph:
Prevent seeking past end of stream, which causes a huge # of Seeks once EOF is reached, if we try and request frame #s too large for the file. For example, if a video file is encoded with the metadata it has 1600 frames, but in reality, only 1500 frames are found, those last 100 frames are not possible to find, and we keep Seek()ing looking for the frame which doesn't exist... this is so slow, it might as well be a freeze.

### DIFF
--- a/examples/Example.cpp
+++ b/examples/Example.cpp
@@ -24,7 +24,7 @@ using namespace openshot;
 int main(int argc, char* argv[]) {
 
     // FFmpeg Reader performance test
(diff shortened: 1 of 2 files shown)


jonoomph:
[Freeze when Seeking past EOF](https://trello.com/c/Z7JNimhm/218-freeze-when-seeking-past-eof)

PR #876, opened 2022-11-27
Fix a huge regression which invokes the avcodec_open() twice

jonoomph:
Fix a huge regression which invokes the `avcodec_open()` twice for every call to `FFmpegReader::Open()`, leaving many threads running and never joined/closed. Also, adding some additional error message when attempting to call `avcodec_open()`.

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -573,8 +573,8 @@ void FFmpegReader::Open() {
 			CheckFPS();
 		}
 
-        // Mark as "open"
-        is_open = true;
+		// Mark as "open"
+		is_open = true;
 
 		// Seek back to beginning of file (if not already seeking)
 		if (!is_seeking) {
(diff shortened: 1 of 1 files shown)


src/FFmpegReader.cpp
```diff
@@ -569,13 +573,13 @@ void FFmpegReader::Open() {
 			CheckFPS();
 		}
 
+		// Mark as "open"
+		is_open = true;
+
 		// Seek back to beginning of file (if not already seeking)
 		if (!is_seeking) {
 			Seek(1);
```
jonoomph, line 581: Since the Seek() method can actually Close() and re Open() this reader, it's important we mark the reader as `is_open = True` before doing this... so our Close method can actually clean up the open reader.

jonoomph:
[OpenShot Threads Always Grow - CRITICAL](https://trello.com/c/AAJIC8dT/236-openshot-threads-always-grow-critical)

PR #877, opened 2022-11-27
fix div zero when `ExtractSamples`

fufesou:
Signed-off-by: fufesou <shuanglongchen@yeah.net>

https://github.com/OpenShot/openshot-qt/issues/5003#issuecomment-1328219623

### DIFF
--- a/src/AudioWaveformer.cpp
+++ b/src/AudioWaveformer.cpp
@@ -120,7 +120,7 @@ AudioWaveformData AudioWaveformer::ExtractSamples(int channel, int num_per_secon
 
         // Scale all values to the -1 to +1 range (regardless of how small or how large the
         // original audio sample values are)
-        if (normalize) {
+        if (normalize && samples_max > 0) {
             float scale = 1.0f / samples_max;
             data.scale(total_samples, scale);
         }


src/AudioWaveformer.cpp
```diff
@@ -120,7 +120,7 @@ AudioWaveformData AudioWaveformer::ExtractSamples(int channel, int num_per_secon
 
         // Scale all values to the -1 to +1 range (regardless of how small or how large the
         // original audio sample values are)
-        if (normalize) {
+        if (normalize && samples_max > 0) {
```
ferdnyc, line 123: I'm not really involved here anymore, but that doesn't look right to me. As the comment above says, it wants to "scale all values to the **-1 to +1 range**, so negative values should still be scaled. You'd literally want to check that `samples_max != 0.0`, as anything else is fair game.

ferdnyc: (Though it *may* be correct to use the absolute value of `samples_max` in the division, so as to not produce a negative scaling factor. I don't know, the audio code was never my thing.)

fufesou: https://github.com/fufesou/libopenshot/blob/432917e177770f42c58a9c812ec5cf7e6204fbc4/src/AudioWaveformer.cpp#L66

fufesou: `sample_max >= 0` because
...
[`float samples_max = 0.0;`](https://github.com/fufesou/libopenshot/blob/432917e177770f42c58a9c812ec5cf7e6204fbc4/src/AudioWaveformer.cpp#L66)
...
[`samples_max = std::max(samples_max, chunk_max);`](https://github.com/fufesou/libopenshot/blob/432917e177770f42c58a9c812ec5cf7e6204fbc4/src/AudioWaveformer.cpp#L111)

PR #878, opened 2022-11-29
Support for New Linux Build Server + glibc Wrapping

jonoomph:
Related to: https://github.com/OpenShot/libopenshot-audio/pull/146

Adding support for a new Linux build server, which has updated dependencies, but also a newer glibc version. Experimental support for wrapping glibc for older distros: https://github.com/wheybags/glibc_version_header

### DIFF
--- a/src/Fraction.cpp
+++ b/src/Fraction.cpp
@@ -65,6 +65,9 @@ int Fraction::GreatestCommonDenominator() {
 void Fraction::Reduce() {
 	// Get the greatest common denominator
 	int GCD = GreatestCommonDenominator();
+	if (GCD == 0) {
+	    return;
+	}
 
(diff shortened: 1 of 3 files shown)


src/Fraction.cpp
```diff
@@ -65,6 +65,9 @@ int Fraction::GreatestCommonDenominator() {
 void Fraction::Reduce() {
 	// Get the greatest common denominator
 	int GCD = GreatestCommonDenominator();
+	if (GCD == 0) {
```
jonoomph, line 68: Prevent divide by zero error

jonoomph:
[New AppImage + New Linux Builder](https://trello.com/c/u8jCeev8/240-new-appimage-new-linux-builder)

PR #881, opened 2022-12-15
Add Support for Windows 7/8.1

jonoomph:
Currently, OpenShot 3.0 crashes on Windows 7 and Windows 8.1 during launch (or immediately after launch). I can repeat this in a VM, but I'm not sure if this happens in non-VM environments. We fail to find the default audio device when trying to detect `sample rate`, and need some added protection.

### DIFF
--- a/src/Qt/AudioPlaybackThread.cpp
+++ b/src/Qt/AudioPlaybackThread.cpp
@@ -21,8 +21,8 @@
 #include "../Settings.h"
 
 #include <mutex>
-#include <thread>    // for std::this_thread::sleep_for
-#include <chrono>    // for std::chrono::milliseconds
+#include <thread>	// for std::this_thread::sleep_for
+#include <chrono>	// for std::chrono::milliseconds
 
 using namespace juce;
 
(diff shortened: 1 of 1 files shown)


src/Qt/AudioPlaybackThread.cpp
```diff
@@ -64,27 +68,35 @@ namespace openshot
...
+			deviceSetup.outputChannels = channels;
+
+			// Detect default sample rate (of default device)
+			m_pInstance->audioDeviceManager.initialiseWithDefaultDevices (0, 2);
+
+			// Set current sample rate (from audio device)
+			double currentSampleRate = 44100;
+			AudioIODevice* currentDevice = m_pInstance->audioDeviceManager.getCurrentAudioDevice();
+			if (currentDevice) {
+				currentSampleRate = currentDevice->getCurrentSampleRate();
```
jonoomph, line 88: Protecting from a NULL currentDevice - which apparently happens on Windows 7/8.1 VMs

src/Qt/AudioPlaybackThread.cpp
```diff
@@ -30,15 +31,18 @@ namespace openshot
...
+	// Create or Get audio device singleton with default settings (44100, 2)
+	AudioDeviceManagerSingleton *AudioDeviceManagerSingleton::Instance()
+	{
+		return AudioDeviceManagerSingleton::Instance(44100, 2);
+	}
 
 	// Create or Get an instance of the device manager singleton (with custom sample rate & channels)
 	AudioDeviceManagerSingleton *AudioDeviceManagerSingleton::Instance(int rate, int channels)
 	{
+		static std::mutex mutex;
```
jonoomph, line 43: Adding mutex to prevent concurrent calls to the `Instance()` method - which seemed to be happening during debugging

PR #883, opened 2022-12-18
Refactoring Audio Device detection (don't crash on Windows, find actual sample rate and device details)

jonoomph:
- Fixing a crash on Windows 7/8.1
- Refactor of Audio Device Manager
- Find the 1st valid audio device, if requested device is not valid or not available
- New method to return current active audio device


jonoomph:
[Windows 7 / 8.1 Crash during launch (or soon after)](https://trello.com/c/tBLjvsCj/17-windows-7-81-crash-during-launch-or-soon-after)

PR #887, opened 2022-12-20
Fix ColorShift Y Keyframe Logic

jonoomph:
Fix ColorShift bug when only adjusting the Y keyframes. The Y keyframes would only work correctly when also using the X axis keyframes, which was a bug.


jonoomph:
[Color Shift effect not working (only Y keyframes)](https://trello.com/c/x9rGDtTS/12-color-shift-effect-not-working-only-y-keyframes)

PR #888, opened 2022-12-20
Fix Transition/Mask `replace_image` Property

jonoomph:
This fixes a regression from a recent change to the Mask effect, which broke the `replace_image` property.


jonoomph:
[Fix 'replace_image' property on Transitions / Masks](https://trello.com/c/TfJJsrwx/18-fix-replaceimage-property-on-transitions-masks)

PR #899, opened 2023-02-14
Improved Caption Effect

jonoomph:
- Improved Caption Effect (updated Regex which no longer requires a blank line between captions, fixed long word wrapping, fixed word-wrapping for languages which don't use a space character, captions can now start with a number)
- Added unit tests for Caption effect (windows and non-windows tests - since fonts render differently in windows)
- Allow audio files to generate "Waveform" image before effects are processed - which allows for Caption effect on top of animated waveforms (more interesting, especially for audio-only transcriptions)
- When generating audio animated waveforms, use the entire `Timeline` frame size - instead of 720x480.
- Fixed bug on Clip initialization, to detect file paths containing a `%` to use a `FFmpegReader`


jonoomph:
[Caption Effect Bugs (Wrapping, Blank lines)](https://trello.com/c/DQ5xaHOe/40-caption-effect-bugs-wrapping-blank-lines)

jonoomph:
[Audio-Only Clips + Waveform + Caption Effect](https://trello.com/c/TXubwaEq/41-audio-only-clips-waveform-caption-effect)

PR #909, opened 2023-03-20
Fix Memory leaks in FFmpegReader, Clip, and FrameMapper

jonoomph:
Due to a recent regression, and some investigating into memory leaks (thanks to Valgrind), we found and fixed 3 serious memory leaks in libopenshot. 

- AVFRAME in FFmpegReader::ProcessVideoPacket - initializing struct, but not deleting it if the packet was not ready
- AVFRAME in FFmpegReader::ProcessAudioPacket - initializing struct, but not deleting it if the packet was not ready
- Clip::final_cache was not being initialized correctly, and was defaulting to **infinite** Clip cache - always growing as large as the # of frames in the clip - oops - large leak

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -95,9 +95,6 @@ void Clip::init_settings()
 
 	// Init reader info struct
 	init_reader_settings();
-
-	// Init cache
-	final_cache.SetMaxBytesFromInfo(10, info.width, info.height, info.sample_rate, info.channels);
 }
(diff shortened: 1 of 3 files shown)


jonoomph:
[Memory Leak caused by Regression in Time Re-mapping](https://trello.com/c/vkFqt4AB/48-memory-leak-caused-by-regression-in-time-re-mapping)

PR #915, opened 2023-04-13
Prevent crash in ObjectDetection (due to NULL parentClip)

jonoomph:
Protect `child_clip_id` from accessing NULL `parentClip()` pointer - which causes a crash when OpenShot attempts to update the JSON of an Object Detection effect with a child ID. Also fixing a GitHub CI issue, which is breaking the GitHub Checks/Actions.


jonoomph:
[`Child_Clip_ID` in the Tracker effect causes OpenShot to crash](https://trello.com/c/vFGMVC9t/20-childclipid-in-the-tracker-effect-causes-openshot-to-crash)

PR #916, opened 2023-04-14
Do not divide by zero when normalizing audio waveforms

jonoomph:
Do not divide by zero when normalizing audio waveforms full of 0.0 values. The bug caused a vector full of `NaN` instead of `0.0` values.

### DIFF
--- a/src/AudioWaveformer.cpp
+++ b/src/AudioWaveformer.cpp
@@ -120,7 +120,7 @@ AudioWaveformData AudioWaveformer::ExtractSamples(int channel, int num_per_secon
 
         // Scale all values to the -1 to +1 range (regardless of how small or how large the
         // original audio sample values are)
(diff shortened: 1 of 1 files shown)


jonoomph:
[Webm causes OpenShot to either crash or become unresponsive when you use the "Display | Show Waveform" function.](https://trello.com/c/xnFuis5I/22-webm-causes-openshot-to-either-crash-or-become-unresponsive-when-you-use-the-display-show-waveform-function)

jonoomph:
[Webm causes OpenShot to either crash or become unresponsive when you use the "Display | Show Waveform" function.](https://trello.com/c/xnFuis5I/22-webm-causes-openshot-to-either-crash-or-become-unresponsive-when-you-use-the-display-show-waveform-function)

PR #917, opened 2023-04-14
Fix WEBM packet retry (sending | receiving)

jonoomph:
Fix webm packet reading, to correctly retry packets. Sometimes we must **send** more packets before receiving decoded data, and sometimes we must **receive** more decoded data before sending additional packets. We had a logic error in this code. This is a delicate dance, and we had an infinite loop... oops.


jonoomph:
[WEBM file causes Openshot to Crash in both Windows and Ubuntu.](https://trello.com/c/5smONirX/21-webm-file-causes-openshot-to-crash-in-both-windows-and-ubuntu)

PR #928, opened 2023-05-22
Increase FrameMapper Resampler Input Buffer (to prevent initial audio pop)

jonoomph:
When EXTRA_INPUT_SAMPLES is less than 64, we get a pop/crackle when downsampling from 48000 to 44100 sample rate.


jonoomph:
[Initial Audio Pop when downsampling from 48000 to 44100](https://trello.com/c/YyiSXTYM/43-initial-audio-pop-when-downsampling-from-48000-to-44100)

PR #930, opened 2023-05-30
Fix Frame::GetSamplesPerFrame when channels = 0

kxxt:
This function is relying on the implementation defined behavior of x86_64 to work properly.

On x86_64,  `fmod` returns -NaN when `channels = 0`, it is handled by  `if (samples_per_frame < 0)`.

But on other architectures(like riscv64), `fmod` may return a positive NaN which will cause this function to return a bad value.

It is causing several tests to spin forever on riscv64.

This PR fixes it by directly return 0 when `channels == 0` so that we do not need to deal with NaNs.


jonoomph:
Thanks! LGTM!

PR #932, opened 2023-06-05
Allow Effects to be applied BEFORE or AFTER a clip's keyframes

jonoomph:
Effects are currently processed before we apply keyframes on each clip object. This is usually the preferred sequence of events, for example, we might want to apply a crop or mask before we position and resize the clip (i.e. process the clip’s keyframes).

However, there are times when it makes sense to apply an effect after the clip keyframes, for example, we might want to animate some scrolling credits moving down the screen, and apply a mask AFTER we have applied the clip keyframes. In this example, the scrolling credits might fade away mid-screen due to a mask… but because we apply this mask after the clip keyframes, it would appear to animate down the screen and fade away at a static location.

Mask-Credits-experimental

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -440,20 +440,17 @@ std::shared_ptr<Frame> Clip::GetFrame(std::shared_ptr<openshot::Frame> backgroun
 		// Apply waveform image (if any)
 		apply_waveform(frame, background_frame);
 
-		// Apply local effects to the frame (if any)
-		apply_effects(frame);
+		// Apply effects BEFORE applying keyframes (if any local or global effects are used)
+		apply_effects(frame, background_frame, options, true);
 
-		// Apply global timeline effects (i.e. transitions & masks... if any)
-		if (timeline != NULL && options != NULL) {
-			if (options->is_top_clip) {
-				// Apply global timeline effects (only to top clip... if overlapping, pass in timeline frame number)
-				Timeline* timeline_instance = static_cast<Timeline*>(timeline);
-				frame = timeline_instance->apply_effects(frame, background_frame->number, Layer());
-			}
-		}
-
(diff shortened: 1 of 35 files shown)


jonoomph:
[Improved Effect Sequencing (allow BEFORE or AFTER clip keyframes have been processed)](https://trello.com/c/vBgUwFau/50-improved-effect-sequencing-allow-before-or-after-clip-keyframes-have-been-processed)

jonoomph:
[Improve Crop Effect to allow "Resize Image"](https://trello.com/c/XAQpA4Ok/51-improve-crop-effect-to-allow-resize-image)

jeffski:
Not sure if I was affected by this just recently. The issue I had was applying a mask to a clip that had been stretched due to SCALE_CROP. The clip was an image that was very wide but thin, i.e. 1000px wide by 20px high, it had a mask effect applied and a scale keyframe added.

I think what happens is the clip is scaled due to SCALE_CROP so it becomes really wide to fit in a 1920 x 1080 video, the width will be > 100000's of pixels wide by 1080px high. The mask is then applied to the huge area, which makes it very very slow. Then a scale was added to shrink it down again to fit in 1920px x 1080px.

This might seem like a bit of an edge case and of course you would be better off using SCALE_NONE, but it did make me think if the scaling is done first and then the mask applied then you would be applying the effect on the original shape and not the massive scaled shape.

Anyway I am keen to see if this fixes that scenario because it would be huge performance boost.

jonoomph:
@jeffski Applying the mask effect after the clip keyframes is much more efficient. I experienced the exact issue you mentioned in the example video above, where I have a tall, skinny SVG image of all credits. Applying the mask before, is very slow per frame, but applying the mask after, is only masking the visible portion of the SVG credits image, and thus, is super fast.

PR #938, opened 2023-07-18
Simplify Audio Processing (remove ugly loops)

jonoomph:
Convert sample values into floats (AV_SAMPLE_FMT_FLTP), each channel in it's own data buffer, which makes it easy to pass to Frame::AddAudio. Then, let's utilize the FLTP format as much as possible, to avoid ugly nested loops converting audio sample formats. The primary goal of this PR is to massively simplify audio processing.


jonoomph:
[Simplify Audio Processing (reduce complexity)](https://trello.com/c/ADrABdBY/53-simplify-audio-processing-reduce-complexity)

PR #942, opened 2023-10-07
Ignore NULL nodes in JSON arrays (clips, effects). 

jonoomph:
This can happen sometimes (for an unknown reason), and it currently crashes OpenShot when attempting to Export a video.


jonoomph:
[Export Dialog ERROR: JSON is invalid (missing keys or invalid data types)](https://trello.com/c/gSzQeZvL/58-export-dialog-error-json-is-invalid-missing-keys-or-invalid-data-types)

PR #950, opened 2024-02-23
Improved Tracker & Object Detector Effects (Faster Drawing, Corner Radius, New Style)

jonoomph:
Related to this PR: https://github.com/OpenShot/openshot-qt/pull/5430

**Improvements:**
- Added "Parent" badge to any clip with a parent set (on timeline for clarity)
- Fixed `class_filter` in property window to allow for:
  - empty (to reset the filter)
  - commas between classes work better (i.e. "truck, person")
  - allow spaces on either side of the class names (i.e. "  truck  ,  person  ")
  - Ignore case of class names entered by user
- Background Corner Radius: allow up to 150
- Updated Object Detector to output "class_name: object id" as the format (instead of confidence decimal) - easier to identify and select object in properties.
- Removed "Child Clip ID" from Tracker and Object Detector Effect. "Parent" is now the only supported method of attaching clips to tracked objects.
- Made certain properties read-only (X1, X2, Y1, Y2, etc...)
- Fixed invalid background_alpha and stroke_alpha properties (reversed range... it was backwards)
- Added new Display All Boxes property, which affects all Object Detector tracked boxes with a single property
- Lots of refactoring and simplifying (i've removed 2 times as much code as I've added in this PR)

- Refactored clip caching:
  - Prevent black backgrounds (however, when using "Parent" to a tracked object, the timeline Track order does matter. Example, if you want something on top of the tracked object, it must be on a layer above the tracked object. If you have a clip below a tracked object, it will be drawn under it (and might be hidden). 
  - Simplified some caching code (much easier to follow)
  - Clips still cache their Frame object, but no longer cache them flattened with the previous timeline layers

**Refactor Object Detector** with QPainter (support for corner radius, faster drawing performance)
OpenShot-Object-Detector-Radius

**Refactor Tracker** with QPainter (support for corner radius, faster drawing performance)
OpenShot-Tracker-Radius

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -479,6 +479,24 @@ openshot::EffectBase* Clip::GetEffect(const std::string& id)
 	return nullptr;
 }
 
+// Return the associated ParentClip (if any)
+openshot::Clip* Clip::GetParentClip() {
+    if (!parentObjectId.empty() && (!parentClipObject && !parentTrackedObject)) {
+        // Attach parent clip OR object to this clip
+        AttachToObject(parentObjectId);
+    }
+    return parentClipObject;
+}
+
+// Return the associated Parent Tracked Object (if any)
+std::shared_ptr<openshot::TrackedObjectBase> Clip::GetParentTrackedObject() {
+    if (!parentObjectId.empty() && (!parentClipObject && !parentTrackedObject)) {
+        // Attach parent clip OR object to this clip
+        AttachToObject(parentObjectId);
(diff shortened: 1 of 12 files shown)


jonoomph:
[Tracker Effect has serious regressions (different `types` crash, attributes don't work, poor performance)](https://trello.com/c/zxNARDjF/27-tracker-effect-has-serious-regressions-different-types-crash-attributes-dont-work-poor-performance)

PR #954, opened 2024-05-08
Add compatibility with FFMPEG 7.0

eclipseo:
channel_layout has been replaced with ch_layout

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -9,7 +9,7 @@
  * @ref License
  */
 
-// Copyright (c) 2008-2019 OpenShot Studios, LLC, Fabrice Bellard
+// Copyright (c) 2008-2024 OpenShot Studios, LLC, Fabrice Bellard
 //
 // SPDX-License-Identifier: LGPL-3.0-or-later
 
@@ -671,8 +671,13 @@ bool FFmpegReader::HasAlbumArt() {
 
 void FFmpegReader::UpdateAudioInfo() {
 	// Set default audio channel layout (if needed)
+#if HAVE_CH_LAYOUT
+	if (!av_channel_layout_check(&(AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->ch_layout)))
+		AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->ch_layout = (AVChannelLayout) AV_CHANNEL_LAYOUT_STEREO;
+#else
 	if (AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->channel_layout == 0)
 		AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->channel_layout = av_get_default_channel_layout(AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->channels);
+#endif
 
 	if (info.sample_rate > 0) {
 		// Skip init - if info struct already populated
@@ -683,8 +688,13 @@ void FFmpegReader::UpdateAudioInfo() {
 	info.has_audio = true;
 	info.file_size = pFormatCtx->pb ? avio_size(pFormatCtx->pb) : -1;
 	info.acodec = aCodecCtx->codec->name;
+#if HAVE_CH_LAYOUT
+	info.channels = AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->ch_layout.nb_channels;
+	info.channel_layout = (ChannelLayout) AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->ch_layout.u.mask;
+#else
 	info.channels = AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->channels;
 	info.channel_layout = (ChannelLayout) AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->channel_layout;
+#endif
 	info.sample_rate = AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->sample_rate;
 	info.audio_bit_rate = AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->bit_rate;
 	if (info.audio_bit_rate <= 0) {
@@ -1593,11 +1603,16 @@ void FFmpegReader::ProcessAudioPacket(int64_t requested_frame) {
 
 		// determine how many samples were decoded
 		int plane_size = -1;
-		data_size = av_samples_get_buffer_size(&plane_size, AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->channels,
+#if HAVE_CH_LAYOUT
+		int nb_channels = AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->ch_layout.nb_channels;
+#else
+		int nb_channels = AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->channels;
+#endif
+		data_size = av_samples_get_buffer_size(&plane_size, nb_channels,
 											   audio_frame->nb_samples, (AVSampleFormat) (AV_GET_SAMPLE_FORMAT(aStream, aCodecCtx)), 1);
 
 		// Calculate total number of samples
-		packet_samples = audio_frame->nb_samples * AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->channels;
+		packet_samples = audio_frame->nb_samples * nb_channels;
 	} else {
 		if (audio_frame) {
 			// Free audio frame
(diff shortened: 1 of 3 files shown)


src/FFmpegReader.cpp
```diff
@@ -671,8 +671,13 @@ bool FFmpegReader::HasAlbumArt() {
 
 void FFmpegReader::UpdateAudioInfo() {
 	// Set default audio channel layout (if needed)
+#if HAVE_CH_LAYOUT
+	if (!AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->ch_layout.nb_channels)
```
jonoomph: I'm confused by this `if (!nb_channels)` logic... hard coding channel_layout to Stereo if 0 channels are listed? The former logic sets the channel layout if it's not already set, based on the # of channels found. Let's make sure this logic really makes sense, and also, let's try and keep both code paths similar, if possible.

src/FFmpegReader.cpp
```diff
@@ -1593,11 +1603,19 @@ void FFmpegReader::ProcessAudioPacket(int64_t requested_frame) {
 
 		// determine how many samples were decoded
 		int plane_size = -1;
+#if HAVE_CH_LAYOUT
+		data_size = av_samples_get_buffer_size(&plane_size, AV_GET_CODEC_ATTRIBUTES(aStream, aCodecCtx)->ch_layout.nb_channels,
```
jonoomph: Since we do this calculation on both code paths, it might be nice to set the `data_size` and `nb_samples` in the conditional #if, and then do the `packet_samples` calculation 1 time, below the conditional ifs.

src/FFmpegWriter.cpp
```diff
@@ -1042,7 +1042,9 @@ AVStream *FFmpegWriter::add_audio_stream() {
 
 	// Set the sample parameters
 	c->bit_rate = info.audio_bit_rate;
+#if !HAVE_CH_LAYOUT
```
jonoomph, line 1045: So, we no longer need to set `c->channels` manually? I'm assuming this gets set automatically when channel_layout is assigned below? (in the newer code path)

src/FFmpegWriter.cpp
```diff
@@ -1100,6 +1121,16 @@ AVStream *FFmpegWriter::add_audio_stream() {
 
 	AV_COPY_PARAMS_FROM_CONTEXT(st, c);
 
+#if HAVE_CH_LAYOUT
+	ZmqLogger::Instance()->AppendDebugMethod(
```
jonoomph: I would prefer we set a variable or two, instead of repeating the entire debug statement, just to keep things tidy.

src/FFmpegWriter.cpp
```diff
@@ -1803,9 +1851,13 @@ void FFmpegWriter::write_audio_packets(bool is_final, std::shared_ptr<openshot::
 
 			// Create output frame (and allocate arrays)
 			frame_final->nb_samples = audio_input_frame_size;
+#if HAVE_CH_LAYOUT
+			av_channel_layout_from_mask(&frame_final->ch_layout, info.channel_layout);
```
jonoomph, line 1858: Same question as before... I'm assuming we don't need to set channels manually anymore?

src/FFmpegWriter.cpp
```diff
@@ -1854,9 +1906,15 @@ void FFmpegWriter::write_audio_packets(bool is_final, std::shared_ptr<openshot::
 			frame_final->nb_samples = audio_input_frame_size;
 
 			// Fill the final_frame AVFrame with audio (non planar)
+#if HAVE_CH_LAYOUT
+			avcodec_fill_audio_frame(frame_final, audio_codec_ctx->ch_layout.nb_channels,
```
jonoomph: If possible, let's set a variable for # of channels, so we don't need to repeat the same exact code twice.

jonoomph:
@eclipseo Thanks a bunch for taking a stab at this! I've left a few comments for you (sorry for the nitpicks). Mostly looks good, just a few questions and a few ideas to keep the changes tidy. :+1:

jonoomph:
[PR: FFmpeg 7 Support in libopenshot-audio](https://trello.com/c/RH4dTtOM/100-pr-ffmpeg-7-support-in-libopenshot-audio)

src/FFmpegReader.cpp
eclipseo: Using av_channel_layout_check() instead: https://ffmpeg.org/doxygen/trunk/group__lavu__audio__channels.html#gad15a6bf80ee8551ee4a4789d970ccbea

src/FFmpegWriter.cpp
eclipseo: Yes ch_layout:

https://ffmpeg.org/doxygen/trunk/structAVChannelLayout.html

```
/**
  * An AVChannelLayout holds information about the channel layout of audio data.
  *
  * A channel layout here is defined as a set of channels ordered in a specific
  * way (unless the channel order is AV_CHANNEL_ORDER_UNSPEC, in which case an
  * AVChannelLayout carries only the channel count).
  * All orders may be treated as if they were AV_CHANNEL_ORDER_UNSPEC by
  * ignoring everything but the channel count, as long as av_channel_layout_check()
  * considers they are valid.
  *
  * Unlike most structures in FFmpeg, sizeof(AVChannelLayout) is a part of the
  * public ABI and may be used by the caller. E.g. it may be allocated on stack
  * or embedded in caller-defined structs.
  *
  * AVChannelLayout can be initialized as follows:
  * - default initialization with {0}, followed by setting all used fields
  *   correctly;
  * - by assigning one of the predefined AV_CHANNEL_LAYOUT_* initializers;
  * - with a constructor function, such as av_channel_layout_default(),
  *   av_channel_layout_from_mask() or av_channel_layout_from_string().
  *
  * The channel layout must be unitialized with av_channel_layout_uninit()
  *
  * Copying an AVChannelLayout via assigning is forbidden,
  * av_channel_layout_copy() must be used instead (and its return value should
  * be checked)
  *
  * No new fields may be added to it without a major version bump, except for
  * new elements of the union fitting in sizeof(uint64_t).
  */
 typedef struct [AVChannelLayout](https://ffmpeg.org/doxygen/trunk/structAVChannelLayout.html) {
     /**
      * Channel order used in this layout.
      * This is a mandatory field.
      */
     enum [AVChannelOrder](https://ffmpeg.org/doxygen/trunk/group__lavu__audio__channels.html#ga252e1528ae55fbf873266ca4c7e3b69c) [order](https://ffmpeg.org/doxygen/trunk/structAVChannelLayout.html#a12d3f7dda8395c51bd5e90dbd948064d);
  
     /**
      * Number of channels in this layout. Mandatory field.
      */
     int [nb_channels](https://ffmpeg.org/doxygen/trunk/structAVChannelLayout.html#adfd3f460a8ea1575baa32852d9248d3c);
  
     /**
      * Details about which channels are present in this layout.
      * For AV_CHANNEL_ORDER_UNSPEC, this field is undefined and must not be
      * used.
      */
     union {
         /**
          * This member must be used for AV_CHANNEL_ORDER_NATIVE, and may be used
          * for AV_CHANNEL_ORDER_AMBISONIC to signal non-diegetic channels.
          * It is a bitmask, where the position of each set bit means that the
          * AVChannel with the corresponding value is present.
          *
          * I.e. when (mask & (1 << AV_CHAN_FOO)) is non-zero, then AV_CHAN_FOO
          * is present in the layout. Otherwise it is not present.
          *
          * @note when a channel layout using a bitmask is constructed or
          * modified manually (i.e.  not using any of the av_channel_layout_*
          * functions), the code doing it must ensure that the number of set bits
          * is equal to nb_channels.
          */
         uint64_t [mask](https://ffmpeg.org/doxygen/trunk/structAVChannelLayout.html#a78b6087054bf8c308355384d050b5d51);
         /**
          * This member must be used when the channel order is
          * AV_CHANNEL_ORDER_CUSTOM. It is a nb_channels-sized array, with each
          * element signalling the presence of the AVChannel with the
          * corresponding value in map[i].id.
          *
          * I.e. when map[i].id is equal to AV_CHAN_FOO, then AV_CH_FOO is the
          * i-th channel in the audio data.
          *
          * When map[i].id is in the range between AV_CHAN_AMBISONIC_BASE and
          * AV_CHAN_AMBISONIC_END (inclusive), the channel contains an ambisonic
          * component with ACN index (as defined above)
          * n = map[i].id - AV_CHAN_AMBISONIC_BASE.
          *
          * map[i].name may be filled with a 0-terminated string, in which case
          * it will be used for the purpose of identifying the channel with the
          * convenience functions below. Otherise it must be zeroed.
          */
         [AVChannelCustom](https://ffmpeg.org/doxygen/trunk/structAVChannelCustom.html) *[map](https://ffmpeg.org/doxygen/trunk/structAVChannelLayout.html#af0e00bccc9a5d934b48c603a9a49c708);
     } [u](https://ffmpeg.org/doxygen/trunk/structAVChannelLayout.html#a4a05a3aac57d58ce285c8b536e45d13d);
  
     /**
      * For some private data of the user.
      */
     void *[opaque](https://ffmpeg.org/doxygen/trunk/structAVChannelLayout.html#ae4a7b1c772395f34bba287d035ae9653);
 } [AVChannelLayout](https://ffmpeg.org/doxygen/trunk/structAVChannelLayout.html);
```

src/FFmpegWriter.cpp
eclipseo: Indeed. Here we get the mask from info.channel_layout, and create all ch_layout values from it.

src/FFmpegWriter.cpp
eclipseo: ok

src/FFmpegWriter.cpp
eclipseo: ok

eclipseo:
> So, we no longer need to set c->channels manually? I'm assuming this gets set automatically when channel_layout is assigned below? (in the newer code path)

Yes ch_layour is a structure that contains both the number of channels and the layout at the same time.

I updated various thing according to your notes, could check it?

jonoomph:
@eclipseo Thanks! Your changes look great, much easier to follow. I'll do some additional testing on this today, just to verify no issues arise from our build servers, and then hopefully get it merged.

jonoomph:
@eclipseo Build error:
```
/home/runner/work/libopenshot/libopenshot/src/FFmpegWriter.cpp: In member function ‘AVStream* openshot::FFmpegWriter::add_audio_stream()’:
/home/runner/work/libopenshot/libopenshot/src/FFmpegWriter.cpp:1125:10: error: conflicting declaration ‘uint64_t channel_layout’
 1125 | uint64_t channel_layout;
      |          ^~~~~~~~~~~~~~
/home/runner/work/libopenshot/libopenshot/src/FFmpegWriter.cpp:1084:24: note: previous declaration as ‘const uint64_t channel_layout’
 1084 |         const uint64_t channel_layout = info.channel_layout;
      |                        ^~~~~~~~~~~~~~
/home/runner/work/libopenshot/libopenshot/src/FFmpegWriter.cpp:1136:20: error: assignment of read-only variable ‘channel_layout’
 1136 |     channel_layout = c->channel_layout;
      |     ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
gmake[2]: *** [src/CMakeFiles/openshot.dir/build.make:398: src/CMakeFiles/openshot.dir/FFmpegWriter.cpp.o] Error 1
gmake[2]: Leaving directory '/home/runner/work/libopenshot/libopenshot/build'
gmake[1]: *** [CMakeFiles/Makefile2:1810: src/CMakeFiles/openshot.dir/all] Error 2
gmake[1]: Leaving directory '/home/runner/work/libopenshot/libopenshot/build'
gmake: *** [Makefile:166: all] Error 2
```

eclipseo:
Ha my bad I had reused a variable name.

PR #966, opened 2024-06-08
Java Swig Bindings for libopenshot

jonoomph:
Add support for Java bindings for libopenshot. This generates the *.class and *.java files, and a JNI library (libopenshot-java.so) to invoke the actual C++ openshot library, and a openshotJNI.jar which contains all of these files.

This is disabled by default. To enable it during compilation, use `cmake -DENABLE_JAVA=1 ../`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
       - uses: haya14busa/action-cond@v1
         id: coverage
         with:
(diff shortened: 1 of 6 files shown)


jonoomph:
[Java Bindings for libopenshot](https://trello.com/c/iEisRd9Q/117-java-bindings-for-libopenshot)

PR #973, opened 2024-09-24
Do not add +1 to GetMaxFrame() calculation

jonoomph:
Do not add +1 to GetMaxFrame() calculation. This is incorrect math for calculating the max frame of a timeline, and not correct. For example, 1 second long timeline at 30 FPS, should have exactly 30 frames (not 31).


jonoomph:
[Dragging playhead, clips, or transitions past the start of the timeline, should stick to 0.0 (not jump back)](https://trello.com/c/uphcXd8s/14-dragging-playhead-clips-or-transitions-past-the-start-of-the-timeline-should-stick-to-00-not-jump-back)

PR #974, opened 2024-09-30
Adding GetMinTime & GetMinFrame methods to Timeline

jonoomph:
These are used to get the start time and position of the first element on the timeline. This will be a new UI function in the next version of OpenShot, so an export can quickly trim the start or end to the clips/effects on the timeline.


jonoomph:
[Dragging playhead, clips, or transitions past the start of the timeline, should stick to 0.0 (not jump back)](https://trello.com/c/uphcXd8s/14-dragging-playhead-clips-or-transitions-past-the-start-of-the-timeline-should-stick-to-00-not-jump-back)

PR #983, opened 2024-12-13
Fix for Windows 11 Audio Lag

jonoomph:
Setting default **audio buffer size** on Audio Device Manager to **512**, to resolve issue on Windows 11 (and some users Windows 10). Newer versions of Windows seem to request 2500 ish audio samples per callback, instead of the 512 requested by Linux. This increase in buffer size causes libopenshot to decode many more frames than normal, causing a 1 to 2 second lag before hearing any audio playback on the OS.

This is also building against our new version of [libopenshot-audio,](https://github.com/OpenShot/libopenshot-audio/pull/161) which as been upgraded to JUCE 7.0.10, and now uses C++ standard 17, and targets Mac OS 10.12 (changed from 10.9).


jonoomph:
[Windows 11 24H2 causes OpenShot audio issues:  Audio Delay and Audio Preference settings not working.](https://trello.com/c/7ex9sFS4/39-windows-11-24h2-causes-openshot-audio-issues-audio-delay-and-audio-preference-settings-not-working)

jonoomph:
This PR requires changes from libopenshot-audio to build correctly. I'm going to go ahead and merge, as it should be fine. Famous last words, lol.

PR #988, opened 2025-03-01
Import GIF formatted images using FFmpeg

jonoomph:
Import GIF formatted images using FFmpeg, to allow for animation and alpha channels.


jonoomph:
[Animated GIF files are treated as a single image file.](https://trello.com/c/DWVLPy41/45-animated-gif-files-are-treated-as-a-single-image-file)

PR #989, opened 2025-03-02
Fix GIF video_length and duration calculations, so we currently handl…

jonoomph:
Fix the **GIF** `video_length` and `duration` calculations, so we can correctly handle **animated** and **static** gifs.


jonoomph:
[Animated GIF files are treated as a single image file.](https://trello.com/c/DWVLPy41/45-animated-gif-files-are-treated-as-a-single-image-file)

PR #992, opened 2025-03-10
Set libx265 hvc1 tag (for Apple playback compatibility)

### LINKED ISSUE #990: H265 video could not be played correctly in Apple devices
As mentioned in [`ffmpeg` document](https://trac.ffmpeg.org/wiki/Encode/H.265#FinalCutandApplestuffcompatibility), H265 video could not be played on Apple devices due to lacking of `hvc1` tag on the video. With `ffmpeg`, we could easily tag it with `-tag:v hvc1`, but in `libopenshot`, there are no viable option in `FFMpegWriter::SetOption` method to add `tag:v` to the video. So the result would look like this, a no-video video:

Image

Currently, every time I finished rendering the video, I have to manually run this command to tag the video: `ffmpeg -y -i input.mp4 -c copy -tag:v hvc1 output.mp4` . This command takes no time to run, but the "seems to blank" output video could make all mac users (even people who are developer)  really confuse. We should add this tag for all H265 video so non-developer user could watch H265 right after finish rendering without extra step. This process will not affect any other platform like linux/windows.

jonoomph:
Set libx265 hvc1 tag (for Apple playback compatibility). Apparently, we can't just set this on the mux_dict, but rather set the tag on the video codec (libx265). This emulates the behavior of `-tag:v hvc1` using the **ffmpeg** command line.
Fixes https://github.com/OpenShot/libopenshot/issues/990

### DIFF
--- a/src/FFmpegWriter.cpp
+++ b/src/FFmpegWriter.cpp
@@ -647,11 +647,6 @@ void FFmpegWriter::WriteHeader() {
 		av_dict_set(&oc->metadata, iter->first.c_str(), iter->second.c_str(), 0);
 	}
 
-	// For HEVC (H265) to playback on Apple devices: https://github.com/OpenShot/libopenshot/issues/990
-	if (info.has_video && video_codec_ctx && video_codec_ctx->codec_id == AV_CODEC_ID_HEVC) {
-		av_dict_set(&mux_dict, "tag:v", "hvc1", 0);
-	}
-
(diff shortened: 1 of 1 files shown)


jonoomph:
[h.265 videos exported in OpenShot can not be played back on Apple devices due to hvc1 tag missing](https://trello.com/c/pr0OcOHM/46-h265-videos-exported-in-openshot-can-not-be-played-back-on-apple-devices-due-to-hvc1-tag-missing)

PR #993, opened 2025-03-10
Adding alpha to JSON properties of wave_color

jonoomph:
Adding alpha to JSON properties of wave_color, for some reason we were not returning this to the openshot-qt gui. This allows for a clip's audio wave color to be translucent.


jonoomph:
[Color Picker screen is missing the "Alpha Channel:" field.](https://trello.com/c/WjymjHRs/39-color-picker-screen-is-missing-the-alpha-channel-field)

PR #995, opened 2025-03-12
Improve "rotation" metadata to auto-rotate and auto-scale vertical videos

jonoomph:
Read `side data` rotation metadata from mobile videos, so that vertical videos can be auto-rotated and auto-scaled to fit the project size.

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -111,24 +111,39 @@ void Clip::init_reader_settings() {
 	}
 }
 
-// Init reader's rotation (if any)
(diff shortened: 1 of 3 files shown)


jonoomph:
[Vertical video is being cropped when imported.  Additionally, the Transform tool (blue handles) do not work when trying to adjust Scale X & Y](https://trello.com/c/0gwIonbl/6-vertical-video-is-being-cropped-when-imported-additionally-the-transform-tool-blue-handles-do-not-work-when-trying-to-adjust-sca)

PR #999, opened 2025-03-21
Add shadow effect to `libopenshot`

aperture147:
Currently there are no drop shadow effect implemented in `libopenshot`. To achieve that, people usually do this trick:

1. Duplicate the clip that requires shadow dropping
2. Use Brightness effect to make the duplicated clip pitch black
3. Use Blur effect to make the black clip under looked like a spread out shadow
4. Place the Clip below and a bit offset of the original clip

This approach has two disadvantages:
1. Cannot change the color of the shadow: User might want to change the shadow color to a shift the color and no Hue could shift the the color of black
2. Inconvenient to use: User has to duplicate the clip, rearrange and modify effects on it. This is not an easy task to do if user has more than one Clip to drop shadow (even when managing them programmatically)

My implementation of Shadow effect could solves those disadvantages by two features:
1. No extra Clip required.
2. Shadow color could be keyframed. User could change the color freely

Please check this out! All contributions are appreciated!


aperture147:
This effect uses code from #998, please also check it out @jonoomph!

aperture147:
Hi, this PR is related to #998. Please check it out first. I also included test cases for that pull request!

aperture147:
Hi guys... is there any update on this...

PR #1003, opened 2025-05-10
Fix to Caption Rendering (off by 1 frame, and handle 2-3 digit milliseconds)

jonoomph:
Fixing off by 1 error on caption rendering, and forcing milliseconds into 3 digits (for accurate time / frame calculations). The following SRT example should appear now at the exact same time, instead of 2 different times.

```
00:00:01:05 --> 00:10:00:000
This caption is at 05

00:00:01:050 --> 00:10:00:000
This caption is at 050
```


jonoomph:
[Captions Effect Timestamp (3 digits vs 2 digits millisecond)](https://trello.com/c/mPUSCFdJ/33-captions-effect-timestamp-3-digits-vs-2-digits-millisecond)

PR #1008, opened 2025-05-28
Lens Flare effect

jonoomph:
New **lens flare** effect, based on the Gimp FlareFX filter. This is faster than Gimp's version though, allows for user adjustable size and brightness, initial light color, and supports transparent regions.

Frame-00001
Frame-00002


jonoomph:
[Lens Flare effect (based on Gimp FlareFX filter)](https://trello.com/c/bOzOlr24/110-lens-flare-effect-based-on-gimp-flarefx-filter)

PR #1012, opened 2025-06-02
Fix Deinterlace effect crash (plus CPU improvements for Mask effect)

jonoomph:
- Fix Deinterlace effect to compute the correct number of rows per update (prevents crash)
  - Also, made isOdd property writable in the UI
- In Mask effect, precompute brightness/contrast adjustments outside the per-pixel loop
  - Cache image dimensions and raw pointers once instead of calling methods each iteration
  - Add #pragma omp parallel for to the pixel loop for multicore speedup
  - Clamp (A – gray) to [0, 255] so alpha_percent never exceeds 1.0 (resolves orange artifacts)

### DIFF
--- a/src/effects/Deinterlace.cpp
+++ b/src/effects/Deinterlace.cpp
@@ -12,6 +12,7 @@
 
 #include "Deinterlace.h"
 #include "Exceptions.h"
+#include <omp.h>
 
 using namespace openshot;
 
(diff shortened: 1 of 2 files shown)


jonoomph:
[Deinterlace effect Issues:  Unable to change Is Odd Frame from Yes to No & Crash when changing Is Odd Frame to No when working.](https://trello.com/c/tPtLc0lW/49-deinterlace-effect-issues-unable-to-change-is-odd-frame-from-yes-to-no-crash-when-changing-is-odd-frame-to-no-when-working)

PR #1013, opened 2025-06-03
Caching Improvements (Forward & Backwards, Play & Pause)

jonoomph:
Improvements to caching, more reliable, edge cases fixed, better sync with video playback, and ultimately smoother playback experience.

**Caching Direction:**
- Cache frames forward (e.g., 1, 2, 3) when playback speed is positive.
- Cache frames backward (e.g., 3, 2, 1) when playback speed is negative (rewind).
- Use cache->Touch(frame) to prioritize recently accessed frames, moving them to the front of the queue, with oldest frames at the end for eviction.

**Paused State:**
- Continue caching frames ahead the playhead until the cache is full, even when paused.
- Allow users to seek within the cached frame range without clearing the cache.
- Clear and rebuild the cache if the user seeks outside the cached range while paused.

**Playing State:**
- Cache frames from the current playhead position up to the max_frames_ahead setting in the direction of playback.
- Do not clear the cache when seeking during playback to avoid interruptions, especially for slow or complex projects.

### DIFF
--- a/src/CacheBase.h
+++ b/src/CacheBase.h
@@ -93,6 +93,10 @@ namespace openshot {
 		/// @param end_frame_number The ending frame number of the cached frame
 		virtual void Remove(int64_t start_frame_number, int64_t end_frame_number) = 0;
 
+		/// @brief Move frame to front of queue (so it lasts longer)
+		/// @param frame_number The frame number of the cached frame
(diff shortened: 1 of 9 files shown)


jonoomph:
[Improved Caching / Less Stuttering](https://trello.com/c/cviv4fAP/114-improved-caching-less-stuttering)

PR #1015, opened 2025-06-07
Fix regression with displaying changes to clips (frozen preview or ghosting)

jonoomph:
Regression caused by change to VideoCacheThread in 6d5e55175bcf5dae0650bb4066777eb0ee1c9d43


jonoomph:
[Regression with updating clips (no preview or ghosting appearing)](https://trello.com/c/SIgISHs0/116-regression-with-updating-clips-no-preview-or-ghosting-appearing)

PR #1018, opened 2025-09-05
Fix build with ffmpeg 8

marillat:


src/FFmpegWriter.cpp
```diff
@@ -1499,7 +1499,7 @@ void FFmpegWriter::open_video(AVFormatContext *oc, AVStream *st) {
 		switch (video_codec_ctx->codec_id) {
 			case AV_CODEC_ID_H264:
 				video_codec_ctx->max_b_frames = 0;  // At least this GPU doesn't support b-frames
-				video_codec_ctx->profile = FF_PROFILE_H264_BASELINE | FF_PROFILE_H264_CONSTRAINED;
+				video_codec_ctx->profile = AV_PROFILE_H264_BASELINE | AV_PROFILE_H264_BASELINE;
```
rathann: You're changing `*_H264_CONSTRAINED` to `*_H264_BASELINE` here, is that intentional?

rathann:
FYI, these definitions were deprecated back in FFmpeg 6.1 (https://github.com/FFmpeg/FFmpeg/commit/8238bc0b5e3dba271217b1223a901b3f9713dc6e) two years ago.

src/FFmpegWriter.cpp
marillat: Of course should be `AV_PROFILE_H264_CONSTRAINED`
Fixed in [737c9c8](https://github.com/OpenShot/libopenshot/pull/1018/commits/737c9c8a3c9ca8046d6500bad513dafc2d6f2b6c)

rathann:
It looks like this one is no longer required as this was fixed differently in #1045 .

PR #1027, opened 2025-12-18
Qt6 Support: USE_QT6=AUTO|ON|OFF

jonoomph:
Adding Qt6 detection and compatibility, without breaking Qt5. Also, a few version specific changes for API changes on Qt6, but it appears to work well.

  ### libopenshot build-time

  - libopenshot uses the CMake cache variable `-DUSE_QT6`.
  - Values:
      - `AUTO`: prefer Qt6 if found, otherwise fall back to Qt5
      - `ON`: force Qt6
      - `OFF`: force Qt5
  - After that, CMake sets QT_VERSION_MAJOR to 6 or 5 and links Qt${QT_VERSION_MAJOR}.

  ### openshot-qt runtime

  - openshot-qt uses the env var `OPENSHOT_QT_API`.
  - Values:
      - `auto`: try bindings in order and use the first one that imports
      - `pyqt6`
      - `pyside6`
      - `pyqt5`
  - In auto mode, it tries: `pyqt6`, then `pyside6`, then `pyqt5`.


jonoomph:
[Support Qt6 (and Qt5) for libopenshot](https://trello.com/c/ElwGih2n/202-support-qt6-and-qt5-for-libopenshot)

PR #1032, opened 2026-02-05
VideoCacheThread: Add Directional Pre-roll

jonoomph:
The primary idea is to cache "some" small amount of frames (30) prior to the playhead or seek position, to allow cached access to "before playhead" frames without busting cache. This seems to be a common issue on a video editor, where seeing to a spot, and then needing to go forwards and backwards quickly around this position without busting and clearing the entire cache. 

- start cache rebuilds with a preroll offset (settings-based)
- tighten isReady to require frames ahead/behind playhead add unit tests for preroll clamping and readiness
- refactor of memory trimming, to be simpler, and debounce every 30 seconds (to let the OS reclaim memory)

### DIFF
--- a/examples/Example.cpp
+++ b/examples/Example.cpp
@@ -13,83 +13,116 @@
 #include <chrono>
 #include <iostream>
 #include <memory>
+#include <string>
+#include "Clip.h"
 #include "Frame.h"
 #include "FFmpegReader.h"
-#include "FFmpegWriter.h"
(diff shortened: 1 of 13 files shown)


jonoomph:
[VideoCacheThread: Add Directional Pre-roll](https://trello.com/c/58deULRh/251-videocachethread-add-directional-pre-roll)

PR #1034, opened 2026-02-08
Fix sparse-VFR stall and stabilize missing-frame image fallback

jonoomph:
FFmpegReader: fix sparse-VFR stall and stabilize missing-frame image fallback

  - prevent ReadStream() infinite loop when packet counters stall at EOF with no packet/hold state by detecting no-progress and forcing EOF completion
  - improve missing-frame handling on sparse timelines by preferring chronological image sources:
      - previous finalized frame (f->number - 1)
      - last finalized video frame (last_final_video_frame) - last decoded video frame only if not from the future - black frame only as last resort
  - use the same finalized/decoded fallback when largest_frame_processed is no longer cached
  - track/reset last_final_video_frame across finalize/seek/close lifecycle

  Fixes:

  - hard stall around gap frames
  - black flashing from cache churn on sparse streams
  - out-of-order jitter caused by reusing future-decoded frames

### DIFF
--- a/src/FFmpegReader.cpp
+++ b/src/FFmpegReader.cpp
@@ -727,6 +727,7 @@ void FFmpegReader::Close() {
 		seek_video_frame_found = 0;
 		current_video_frame = 0;
 		last_video_frame.reset();
+		last_final_video_frame.reset();
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jonoomph:
[File Causes OpenShot to Crash! (sparse   video packets - lots of gaps, variable frame rate)](https://trello.com/c/sB1VBLZv/225-file-causes-openshot-to-crash-sparse-video-packets-lots-of-gaps-variable-frame-rate)

PR #1035, opened 2026-02-13
Harden playback/cache path for malformed media and concurrent timeline updates

jonoomph:
Harden playback/cache path for malformed media and concurrent timeline updates

  - Invalidate timeline cache on ApplyJsonDiff() clip insert (remove affected frame range).
  - Add lock in Timeline::ClearAllCache() for safe concurrent access.
  - Make VideoCacheThread cross-thread state safe (atomics + seek-state mutex).
  - Lock CacheMemory::Contains() to avoid races.
  - Handle malformed audio streams in FFmpegReader by disabling invalid audio and continuing video-only.
  - Add FPS/timebase safety fallbacks in FFmpeg frame/PTS math.
  - Guard Frame::GetSamplesPerFrame() against invalid inputs.
  - Add/adjust regression tests for cache invalidation and invalid rate handling.

### DIFF
--- a/src/AudioReaderSource.cpp
+++ b/src/AudioReaderSource.cpp
@@ -42,6 +42,8 @@ void AudioReaderSource::getNextAudioBlock(const juce::AudioSourceChannelInfo& in
 		}
 
         while (remaining_samples > 0) {
+            const int previous_remaining = remaining_samples;
+            frame.reset();
             try {
                 // Get current frame object
                 if (reader) {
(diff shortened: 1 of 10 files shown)


jonoomph:
[OpenShot crashes when the attached file is imported and then dragged onto a Track.](https://trello.com/c/d7p3E0JJ/121-openshot-crashes-when-the-attached-file-is-imported-and-then-dragged-onto-a-track)

PR #1036, opened 2026-02-13
Fix VideoCacheThread::isReady() when near the max frame of a timeline

jonoomph:
Fix VideoCacheThread::isReady() so preroll readiness is clamped by frames actually available to the timeline boundary (end/start), preventing playback from stalling in PLAY near the end.

Also, add regression coverage in tests/VideoCacheThread.cpp (isReady: clamps preroll requirement at timeline boundaries) for forward-near-end and backward-near-start cases.

### DIFF
--- a/src/Qt/VideoCacheThread.cpp
+++ b/src/Qt/VideoCacheThread.cpp
@@ -58,10 +58,32 @@ namespace openshot
         }
 
         int dir = computeDirection();
-        if (dir > 0) {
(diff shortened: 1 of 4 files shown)


jonoomph:
[OpenShot does not recognize that it has reached at the end of the timeline and stays in "Play"/"Preview" mode.](https://trello.com/c/RIeQxaoM/260-openshot-does-not-recognize-that-it-has-reached-at-the-end-of-the-timeline-and-stays-in-play-preview-mode)

PR #1037, opened 2026-02-15
Fixes to PTS offsets when audio and video are not aligned - to prevent black initial frames

jonoomph:
- Fix FFmpegReader first-frame black regression by anchoring PTS offset to video start (and correcting video offset to use packet PTS), so AAC preroll/skip-samples no longer shifts frame 1.
- Add examples/test_aac_preroll.mp4 fixture (small, deterministic AAC-preroll sample) for reproducing and testing the regression.


jonoomph:
[Fixes to PTS offsets when audio and video are not aligned - to prevent black initial frames](https://trello.com/c/5E2WOLKI/267-fixes-to-pts-offsets-when-audio-and-video-are-not-aligned-to-prevent-black-initial-frames)

PR #1040, opened 2026-03-13
Audio Transitions (Equal-Power Cross-fade)

jonoomph:
- Timeline now fades audio based on transition placement (global Mask effect)
- Mask position on same layer as clip (left side of clip fades audio in, right side of clip fades audio out, overlapping 2 clips cross-fades both clips)
- Added unit tests


jonoomph:
[Audio Transitions (Equal-Power Cross-Fade)](https://trello.com/c/V6hLgfvv/293-audio-transitions-equal-power-cross-fade)

PR #1041, opened 2026-03-14
Fix Crash in Tracker Effect (when frame size changes)

jonoomph:
Prevent Tracker crashes by resetting optical-flow state when the video frame size changes between tracked frames.


jonoomph:
[Vertical videos + tracker and object detection](https://trello.com/c/gYorQLTl/181-vertical-videos-tracker-and-object-detection)

PR #1042, opened 2026-03-16
Fix a regression in the FrameMapper audio mapping code, which acciden…

jonoomph:
Fix a regression in the FrameMapper audio mapping code, which accidentally silenced certain frames instead of padding them with silence.


jonoomph:
[Fixing Audio Mapping regression (caught by openshot-replay)](https://trello.com/c/UMTe3qtf/301-fixing-audio-mapping-regression-caught-by-openshot-replay)

PR #1043, opened 2026-03-16
Fixing small regression with FrameMapper

jonoomph:
Fixing small regression with FrameMapper (when trying to fix another regression... oops) which was breaking unit tests


jonoomph:
[Fixing Audio Mapping regression (caught by openshot-replay)](https://trello.com/c/UMTe3qtf/301-fixing-audio-mapping-regression-caught-by-openshot-replay)

PR #1046, opened 2026-03-22
Protect core clip keyframes from being empty when Json is set

jonoomph:
Protect core clip keyframes from being empty after loading from JSON (scale_x/y, location_x/y, origin_x/y, shear_x/y, rotation)


jonoomph:
[Deleting all keyframes makes clip disappear (instead of restoring defaults)](https://trello.com/c/1CHFCkiO/16-deleting-all-keyframes-makes-clip-disappear-instead-of-restoring-defaults)

PR #1048, opened 2026-03-24
Fixing Clip Memory Issues (Speed & Corruption)

jonoomph:
Disabling clip cache after lots of testing and benchmarking - cost of caching is greater than the benefit.


jonoomph:
[Slower Export speeds with v3.5.0 vs v3.4.0](https://trello.com/c/IHGxicnW/23-slower-export-speeds-with-v350-vs-v340)

PR #1049, opened 2026-03-25
Vulkan benchmark matrix (comparing different pipelines performance)

jonoomph:
This PR adds some vulkan build support to libopenshot, and a vulkan-benchmark executable to collect some good data on possible future directions for the libopenshot video editing pipeline. Vulkan is very interesting because it opens up a cross-device/cross-platform method for a full GPU pipeline.

Here are some of my results from this benchmark, graphed for clarity. It's mostly exploratory at this point, and just wanted to collect some stats on the various ways we can decode, transform, and composite frames in libopenshot, including Vulkan.

time_720p
time_4k
gpu_compare_aligned_4K
gpu_compare_aligned_720p


jonoomph:
[Vulkan Benchmark Matrix (experimental work, to collect some performance metrics around Vulkan pipelines)](https://trello.com/c/bsJIp3Eq/25-vulkan-benchmark-matrix-experimental-work-to-collect-some-performance-metrics-around-vulkan-pipelines)

PR #1050, opened 2026-03-28
Adding a new override (scale_mode) to Thumbnail function, so we can c…

jonoomph:
Adding a new override (scale_mode) to Thumbnail function, so we can crop thumbnails with no black bars for a more consistent look.


jonoomph:
[Smooth Zooming](https://trello.com/c/xwz5HH8u/30-smooth-zooming)

PR #1051, opened 2026-03-28
Prevent crash with concurrent change_mapping (FrameMapper audio resample context)

jonoomph:
Prevent crash with concurrent change_mapping. We need a mutex for protection when applying mappers to clips.


jonoomph:
[FrameMapper crash (audio resample context)](https://trello.com/c/1Z4cAgx8/35-framemapper-crash-audio-resample-context)

PR #1052, opened 2026-03-29
Improved OMP/FFmpeg Thread Control (Settings)

jonoomph:
This allows more fine-grained control over how many threads OMP and FFmpeg are using, and can be used to tune performance on a particular system.


jonoomph:
[Performance issues between v3.4.0 vs. v3.5.0](https://trello.com/c/Rk43lIkX/34-performance-issues-between-v340-vs-v350)

PR #1057, opened 2026-04-09
Add flac extension to common extension list so FFmpegReader is the first try

jonoomph:
…rst reader we attempt to open it.


jonoomph:
[Fix Flac Extension Support](https://trello.com/c/1Y2SEXxu/221-fix-flac-importing)

PR #1062, opened 2026-04-25
Color Grade effect (Color Correction, Scopes, Grading)

jonoomph:
Related to an entire color UI workflow: https://github.com/OpenShot/openshot-qt/pull/6016


jonoomph:
[First Class Color Grading and Correction!](https://trello.com/c/gQ2l7Vqe/240-first-class-color-grading-and-correction)

PR #1063, opened 2026-04-25
Film Grain effect

jonoomph:
The Film Grain effect adds a gentle moving texture to your video, similar to the tiny speckles you see in real film photography. This can make very clean digital footage feel warmer, more natural, or more cinematic. It can also help blend mixed footage together, especially when one clip looks too sharp or too smooth compared to the rest of your project.


jonoomph:
[Film Grain Effect](https://trello.com/c/F7X1jHRV/241-film-grain-effect)

PR #1067, opened 2026-05-03
Cap Mask contrast factor at maximum contrast

jonoomph:
-Raise the Mask contrast denominator floor from 0.00001 to 0.5, capping the contrast factor at 40. This prevents contrast values near 20 from turning the 8-bit gray lookup into a one-value step function, where tiny mask image differences can create hard transparent edge artifacts during gradient wipes. -Add a regression test covering the maximum contrast transition band.


jonoomph:
[Ruler Time: Editor](https://trello.com/c/Rc2JDwM3/255-ruler-time-editor)

PR #1068, opened 2026-05-05
Fixing FPS on FFmpegWriter exports

jonoomph:
Some exports could report framerates slightly incorrect, for example 30.14 instead of 30 FPS. This was due to an off by one issue on older versions of FFmpeg.


jonoomph:
[30 FPS Export == 30.14 FPS](https://trello.com/c/eMZ4UOLc/258-30-fps-export-3014-fps)

PR #1070, opened 2026-05-08
Fixing GetMaxFrame() to fix tiny float overshoot

jonoomph:
Fixing GetMaxFrame() to fix tiny float overshoot, causing occasional black frames at the end of an export.  For example, it might have concluded that we had 505.00001 frames, and this would then be converted later to 506 with ceil().


jonoomph:
[Exported project ends with a blank frame instead of the last frame of the imported video file.](https://trello.com/c/UYku50Wq/212-exported-project-ends-with-a-blank-frame-instead-of-the-last-frame-of-the-imported-video-file)

PR #1071, opened 2026-05-08
Fix SVG thumbnail rasterization with max decode size (fix fuzzy thumbnails for small SVG files)

jonoomph:
Fix SVG thumbnail rasterization with max decode size (SVG upscaling for tiny sizes)

-Treat SetMaxDecodeSize as the rasterization target for parentless SVG readers, instead of only as a downscale ceiling. This lets thumbnail generation render tiny SVG files at the requested decode size, avoiding blurry thumbnails caused by first rasterizing at the SVG document size. -Raster images and SVGs with a parent clip/timeline keep their existing sizing behavior.


jonoomph:
[SVG thumbnails are blurry in v3.5.1 compared to v3.4.0 or v3.3.0.](https://trello.com/c/FR2LjnfZ/227-svg-thumbnails-are-blurry-in-v351-compared-to-v340-or-v330)

PR #1072, opened 2026-05-08
Fix FFmpeg version fallback parsing

### LINKED ISSUE #947: parse_lib_version is broken on new versions of FFmpeg
Newer FFmpeg releases have their version information headers split into version_major.h and version.h. This causes the parse_lib_version macro from cmake/Modules/FindFFmpeg.cmake to parse them incorrectly. For example FFMPEG_AVFORMAT_VERSION is parsed as 20.100 instead of 60.20.100. This leads to various errors when trying to build libopenshot.

jonoomph:
Update FindFFmpeg.cmake to read component version defines from both version_major.h and version.h. Newer FFmpeg releases moved some version macros into version_major.h, which broke fallback detection when pkg-config did not provide component versions.

Fixes https://github.com/OpenShot/libopenshot/issues/947

### DIFF
--- a/cmake/Modules/FindFFmpeg.cmake
+++ b/cmake/Modules/FindFFmpeg.cmake
@@ -111,20 +111,35 @@ endmacro()
 #
 ### Macro: parse_lib_version
 #
-# Reads the file '${_pkgconfig}/version.h' in the component's _INCLUDE_DIR,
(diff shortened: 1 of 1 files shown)


jonoomph:
[FFMpeg Integration - parse_lib_version issue](https://trello.com/c/5UmBfIC5/157-ffmpeg-integration-parselibversion-issue)

PR #1073, opened 2026-05-09
Apply reader-side orientation metadata (i.e. rotate pixels before Clip)

jonoomph:
Apply reader-side orientation metadata and stabilize tracker stroke rendering

- Move phone/video rotation metadata handling into readers with an internal clip compatibility mode for legacy projects. FFmpegReader now reports oriented dimensions, applies orientation to decoded frames, and preserves old JSON behavior through reader_orientation_mode.

- Also adjust tracker/object-detection stroke width for preview raster scaling so tracked boxes remain visible during clip scale animations.


jonoomph:
[Vertical videos + tracker and object detection](https://trello.com/c/gYorQLTl/78-vertical-videos-tracker-and-object-detection)

PR #1074, opened 2026-05-09
Use TrackedObjectMask as "Mask: Source" for any Effect

jonoomph:
Adding the ability to use TrackedObjectMask as "Mask: Source" for any effect (i.e. blur/pixilate only the tracked area)

- Add tracked effect IDs as mask sources
- Generate bbox masks from Tracker/Object Detection
- Preserve original frame when no bbox is visible
- Add focused mask-source tests

### DIFF
--- a/src/EffectBase.cpp
+++ b/src/EffectBase.cpp
@@ -20,12 +20,17 @@
 #include "Exceptions.h"
 #include "Clip.h"
 #include "Timeline.h"
+#include "TrackedObjectBBox.h"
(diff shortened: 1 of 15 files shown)


jonoomph:
[Tracked Mask: Source (Blur effect - unable to adjust to apply to a specific area only)](https://trello.com/c/lTtoQ1GZ/26-tracked-mask-source-blur-effect-unable-to-adjust-to-apply-to-a-specific-area-only)

PR #1075, opened 2026-05-11
Object detection migration to Yolo v5 (ONNX)

jonoomph:
Newer versions of Yolo require updated OpenCV on ONNX layers, that our build servers do not yet have


jonoomph:
[Object Detector effect attributes not manageable once the effect is applied.](https://trello.com/c/jFdpu2LX/37-object-detector-effect-attributes-not-manageable-once-the-effect-is-applied)

PR #1077, opened 2026-05-21
Avoid Black Frames (Adding a new clip - and it does not appear)

jonoomph:


jonoomph:
["Video Preview" window turns black.](https://trello.com/c/aQGIdwoi/140-video-preview-window-turns-black)

PR #1079, opened 2026-05-25
Clean-up and modernization of our cmake build system

jonoomph:
- Modernized SWIG policy handling for Python, Ruby, and Java without raising the project-wide CMake policy version.
- Switched Python discovery to Python3 on CMake 3.12+, with legacy PythonInterp/PythonLibs fallback for older CMake.
- Removed dead pre-CMake-3.8 SWIG branches.
- Quieted optional babl pkg-config probing.
- Changed FFmpeg probing so avresample is only checked if swresample is unavailable.
- Cleaned Java binding debug configure messages.
- Addressed the SWIG build warnings:
  - Added SWIG-only juce::Thread declaration for Python/Ruby/Java.
  - Added Python QWidget typecheck precedence.
  - Filtered known Ruby binding limitations for multiple inheritance and unexposed Profile operators.
  - Suppressed Java-only SWIG categories for unsupported operators/proxy limitations via Java SWIG flags.

### DIFF
--- a/bindings/java/CMakeLists.txt
+++ b/bindings/java/CMakeLists.txt
@@ -12,12 +12,12 @@
 find_package(SWIG 4.0 REQUIRED)
 include(${SWIG_USE_FILE})
 
-### Enable some legacy SWIG behaviors in newer CMake versions
+### Use modern SWIG target behavior when available
 if (POLICY CMP0078)
-    cmake_policy(SET CMP0078 OLD)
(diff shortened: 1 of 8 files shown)


jonoomph:
[Clean-up of cmake and swig warnings / ugly output](https://trello.com/c/L0wvqrgE/286-clean-up-of-cmake-and-swig-warnings-ugly-output)

PR #1093, opened 2026-09-04
Fix clip location regression while preserving Crop transform behavior

jonoomph:
### Summary

  This fixes a backward-compatibility regression in clip positioning introduced by the geometry-aware location
  calculation.

  The newer calculation correctly addressed positioning for clips using SCALE_CROP, but it was unintentionally
  applied to every scale mode. As a result, existing projects could render images, titles, GIFs, and emojis at
  different positions after upgrading, even though their location_x, location_y, scale_x, and scale_y values
  had not changed.

  ### Fix

  Apply the geometry-aware location calculation only to SCALE_CROP.

  For all other scale modes—Fit, Stretch, and None—restore the historical canvas-relative calculation:

  - Horizontal location is relative to the project width.
  - Vertical location is relative to the project height.

  This preserves compatibility with existing project files while retaining the intended Crop positioning fix.

  ### User impact
  This resolves projects where clips moved or overlapped after upgrading, including:

  - Projects created in OpenShot 3.5.1 and opened in 4.0.0+
  - Cloud API projects rendered with newer libopenshot builds
  - Images, titles, GIFs, emojis, and other media using non-Crop scale modes

  No project-property changes are required for projects created before the regression.

  ### Test coverage
  Added regression tests covering:

  - Crop positioning with each anchor/gravity endpoint
  - Fit, Stretch, and None using historical canvas-relative positioning
  - Exact geometry from the reported Cloud API project
  - Non-square media and canvas dimensions
  - Positive and negative location values

  All libopenshot CI builds and checks pass.

### DIFF
--- a/src/Clip.cpp
+++ b/src/Clip.cpp
@@ -1692,8 +1692,16 @@ QTransform Clip::get_transform(std::shared_ptr<Frame> frame, int width, int heig
 		}
 		return location * (canvas_size - anchored_position);
 	};
-	x += location_offset(location_x_value, x - layout_x, layout_width, scaled_source_width);
-	y += location_offset(location_y_value, y - layout_y, layout_height, scaled_source_height);
+	if (scale == SCALE_CROP) {
+		x += location_offset(location_x_value, x - layout_x, layout_width, scaled_source_width);
+		y += location_offset(location_y_value, y - layout_y, layout_height, scaled_source_height);
+	} else {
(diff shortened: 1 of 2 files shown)


jonoomph:
[Project created in v3.5.1-Release-15816 then opened in v4.0.0-RC is not consistent.](https://trello.com/c/VEbIfDsu/128-project-created-in-v351-release-15816-then-opened-in-v400-rc-is-not-consistent)

PR #1097, opened 2026-09-13
Fix use-after-free when deleting timeline-owned clips

jonoomph:
Deleting a timeline-owned clip frees it immediately. The JSON deletion handler then read its position and duration to invalidate cached frames, accessing freed memory.

Read the clip bounds before removal and retain the existing cache invalidation behavior.

This fixes a confirmed deletion crash path associated with OPENSHOT-BV93. Other stacks in that Sentry issue group remain outside this fix.

### DIFF
--- a/src/Timeline.cpp
+++ b/src/Timeline.cpp
@@ -1566,12 +1566,12 @@ void Timeline::apply_json_to_clips(Json::Value change) {
 
 		// Remove existing clip
 		if (existing_clip) {
-			// Remove clip from timeline
-			RemoveClip(existing_clip);
-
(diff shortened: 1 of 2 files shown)


jonoomph:
[Fix additional Sentry errors with OpenShot 4.0.0 release (top 10 targeted bugs, plus a few misc issues encountered during testing)](https://trello.com/c/anSgMhM8/241-fix-additional-sentry-errors-with-openshot-400-release-top-10-targeted-bugs-plus-a-few-misc-issues-encountered-during-testing)

PR #1098, opened 2026-09-14
Fix camera race crash during Open/Close (and audio crackling after recording)

jonoomph:


jonoomph:
[Fix camera race crash during Open/Close (and audio crackling after recording)- #1098](https://trello.com/c/bTeqiyv7/245-fix-camera-race-crash-during-open-close-and-audio-crackling-after-recording-1098)

PR #1099, opened 2026-09-15
Replace ZeroMQ logging with direct file and console output

jonoomph:
Replace socket-based logging with a standard C++ logger that writes directly to a file and stderr. Each destination has its own severity threshold, allowing detailed file logs without flooding the terminal.

## Changes

- Remove the ZeroMQ dependency and logging connection setup.
- Add `openshot::Logger` with DEBUG through CRITICAL levels, OFF, and independent file and console controls.
- Support environment configuration for standalone scripts and applications, with documented precedence and invalid-value fallback.
- Update engine logging calls and preserve raw crash output and recovery markers.
- Retain `ZmqLogger` as a deprecated source alias and preserve legacy logging controls.
- Support Unicode log paths and older macOS deployment targets.
- Fix Windows test DLL lookup, the x64 OpenCV runtime path, and ObjectMask temporary-file creation.


jonoomph:
[Replace ZeroMQ logging with direct file and console output](https://trello.com/c/3bGR0ez5/247-replace-zeromq-logging-with-direct-file-and-console-output)

PR #1101, opened 2026-09-19
Fix crashes when opening or closing readers and exporting raw video

jonoomph:
This fixes two libopenshot stability issues:

- Overlapping reader Open() and Close() calls now check whether the reader is open while holding the existing mutex, preventing duplicate initialization or cleanup.
- Raw-video export now gives each packet its own buffer, preventing a double-free. It also writes all image planes and correctly scales packet timestamps.

Regression tests cover overlapping reader calls, reopening and reading afterward, and raw-video packet data, timestamps, and decoded colors. A dedicated Linux CI step checks these cases with Valgrind.


jonoomph:
[Fix crashes when opening or closing readers and exporting raw video](https://trello.com/c/DGMbqXme/252-fix-crashes-when-opening-or-closing-readers-and-exporting-raw-video)

